PR #21, opened 2021-09-27
Initial implementation of more robust tests for LogicValue and LogicV…

kimmeljo:
…alues. Still some issues to resolve - see comments in file.

## Backwards-compatibility

## Documentation

### DIFF
--- a/lib/src/values/logic_values.dart
+++ b/lib/src/values/logic_values.dart
@@ -557,6 +557,10 @@ abstract class LogicValues {
 
   const LogicValues._(this.length) : assert(length >= 0);
 
+  /// Converts `bool` [value] to a valid [LogicValues] with 1 bits either one or zero.
+  static LogicValues fromBool(bool value) =>
+      _SmallLogicValues(value ? 1 : 0, 0, 1);
+
   /// Converts `int` [value] to a valid [LogicValues] with [length] number of bits.
   ///
   /// [length] must be greater than or equal to 0.
--- a/test/logic_values_test.dart
+++ b/test/logic_values_test.dart
@@ -11,8 +11,6 @@
 import 'package:rohd/rohd.dart';
 import 'package:test/test.dart';
 
-// TODO: add some tests where length = 64 (border between Small and Big)
-
 void main() {
   
   group('LogicValue', () {
@@ -449,7 +447,7 @@ void main() {
       );
       expect( // test fromBigInt
         () => LogicValues.fromBigInt(BigInt.one, 32) | LogicValues.fromBigInt(BigInt.zero, 32),
-        throwsA(isA<Exception>())
+        throwsA(isA<AssertionError>())
       );
     });
 
@@ -640,7 +638,6 @@ void main() {
     });
 
   });
-
   group('arithmetic operations', () {
 
     test('addsub', () {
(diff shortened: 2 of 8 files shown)


kimmeljo:
Committing early so @mkorbel1 can review some of the outstanding issues (listed as TODOs at the top of the file).

test/logic_values_test.dart
```diff
@@ -11,23 +11,794 @@
...
+      expect( // + normal
+        LogicValues.fromString('0001') + LogicValues.fromString('0011'),
+        equals(LogicValues.fromString('0100')) // 1 + 3 = 4
+      );
+      expect( // - normal
+        LogicValues.fromString('0001') - LogicValues.fromString('0001'),
+        equals(LogicValues.fromString('0000')) // 1 - 1 = 0
+      );
+      /*
+      expect( // + overflow
```
mkorbel1, line 686: these tests failed due to a bug in _SmallLogicValues, nice job!
they appear to be the same at first inspection since the toString only shows 4 bits, but the internal representation was, for example `10000`, and the equality did not mask.
same bug exists in _BigLogicValues, but these tests won't catch it.  add similar tests with >64 bits as well?

mkorbel1: filed #22

test/logic_values_test.dart
```diff
@@ -11,23 +11,794 @@
...
+          LogicValues.fromString('0100').and(),
+          equals(LogicValue.zero)
+        );
+        expect( // and - valid (1's)
+          LogicValues.fromString('1111').and(),
+          equals(LogicValue.one)
+        );
+        /*
+        expect( // and - invalid
+          LogicValues.fromString('010x').and(),
```
mkorbel1, line 758: looks like a bug, good catch!

mkorbel1: filed #23

test/logic_values_test.dart
```diff
@@ -11,23 +11,794 @@
...
+      expect( // toInt - valid
+        LogicValues.fromString('111').toInt(),
+        equals(7)
+      );
+      /*
+      expect( // toInt - invalid
+        LogicValues.filled(65, LogicValue.one).toInt(),
+        throwsA(isA<Exception>())
+      );
+      expect( // toBigInt - invalid
```
mkorbel1, line 504: i think this shouldn't throw an exception?

mkorbel1:
In general, note that there's three separate implementations for `LogicValues`: `_SmallLogicValues`, `_BigLogicValues`, `FilledLogicValues`.  These tests mostly cover the `_SmallLogicValues` I think, but it is possible for bugs to exclusively exist in just one of the three implementations.

lib/src/values/logic_values.dart
```diff
@@ -27,8 +27,8 @@ import 'package:rohd/rohd.dart';
 class _SmallLogicValues extends LogicValues {
   // Each 0/1 bit in value is 0/1 if !invalid, else is x/z
 
-  final int _value;
```
mkorbel1, line 30: these bug fixes have been pushed into main, you can pull from main so these don't show up as a diff in this PR

test/logic_values_test.dart
```diff
@@ -11,23 +11,960 @@
 import 'package:rohd/rohd.dart';
 import 'package:test/test.dart';
 
-//TODO: reversed looks broken
-//TODO: add a test with x & 0 == 0
+// TODO: add some tests where length = 64 (border between Small and Big)
```
mkorbel1, line 14: good idea, there might be interesting signed bugs around this

mkorbel1 (commented):
looking good!

PR #31, opened 2021-10-08
Fix link for Dahlia and add link to Calyx

rachitnigam:
## Description & Motivation

Update README link for the Dahlia programming language to the user landing page instead of the documentation webpage.


mkorbel1:
Thank you for the fix and addition!

PR #73, opened 2021-12-23
Improve logicValues testing, fixing todos

eric-norige:
## Description & Motivation

Wanted to clean up logic_values_test.dart that I changed in previous pull request.  I noticed two TODOs and wanted to follow through on them.

## Related Issue(s)

AFAIK, none

## Testing

These are changes in tests, so not exactly applicable.

## Backwards-compatibility

Not applicable; changed only tests

## Documentation

Not applicable; changed only tests


mkorbel1 (approved):
looks good to me, thank you!

PR #81, opened 2022-02-02
Added a FIR filter example

wswongat:
## Description & Motivation
I would like to add an example to demonstrate the ROHD can also act like generator.
This example is based on the Chisel example
Don't forget to **rebase and merge** my commits.

## Related Issue(s)
NA

## Testing
This example only do the build check and added in example_test.dart

## Backwards-compatibility

NO

## Documentation

No

### DIFF
--- a/example/fir_filter.dart
+++ b/example/fir_filter.dart
@@ -6,6 +6,7 @@
 /// Author: wswongat
 ///
 import 'package:rohd/rohd.dart';
+import 'dart:io';
 
 class FirFilter extends Module {
   // For convenience, map interesting outputs to short variable names for consumers of this module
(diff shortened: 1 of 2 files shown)


test/example_test.dart
```diff
@@ -12,6 +12,7 @@ import 'package:test/test.dart';
 
 import '../example/example.dart' as counter;
 import '../example/tree.dart' as tree;
+import '../example/fir_filter.dart' as fir_filter;
 
 void main() {
```
mkorbel1, line 17: Add a teardown like this:
```
  tearDown(() {
    Simulator.reset();
  });
```

mkorbel1 (changes requested):
Awesome example addition, thank you for contributing it!  Added a couple minor comments.

wswongat:
@mkorbel1 Thank you for your comments. I just made a commit.

example/fir_filter.dart
```diff
@@ -0,0 +1,123 @@
...
+    // Register inputs and outputs of the module in the constructor
+    // Module logic must consume registered inputs and output to registered outputs
+    if (depth < 1) {
+      // Depth Check
+      throw Exception('Depth parameter should more than 1');
+    }
+    // Add input/output port
+    en = addInput('en', en);
+    inputVal = addInput('inputVal', inputVal, width: bitWidth);
+    resetB = addInput('reset', resetB);
```
mkorbel1, line 31: Minor, but the name of the reset pin should be updated to 'resetB' as well

mkorbel1 (changes requested):
Updates look good except one minor thing, thank you!

Sorry for the delay in re-reviewing, I was out of town.

example/fir_filter.dart
wswongat: Thank you for reminding me.
All fixed.

mkorbel1 (approved):

PR #87, opened 2022-03-16
Fix typo in README

eric-norige:
I just noticed this grammar error when skimming the readme and fixed it quickly in the web editor.  Wow, that was convenient.


mkorbel1:
Thank you for fixing it!

Issue #56, opened 2021-11-17
Make `bin` handle `_` for binary number separation


mkorbel1:
## Motivation
In SystemVerilog, you can make a binary number like `6'b11_00_11` to separate portions of the binary.  It would be nice to do that with ROHD.

## Desired solution
The `bin` function automatically removes `_` characters from the string before parsing it as binary.

## Alternatives considered  
Current implementation requires only numbers 0 and 1.

PR #88, opened 2022-03-20
Enhanced bin() to ignore '_'. Ticket - https://github.com/intel/rohd/…

shubskmr:
…issues/56

## Description & Motivation

need to handle '_'s in the binary
https://github.com/intel/rohd/issues/56

## Related Issue(s)

https://github.com/intel/rohd/issues/56

## Testing

added tests in [test/logic_values_test.dart]

## Backwards-compatibility

No

## Documentation

yes documentation updated needed and are included as part of the comments

### DIFF
--- a/lib/src/values/logic_value.dart
+++ b/lib/src/values/logic_value.dart
@@ -134,10 +134,10 @@ class LogicValue {
   }
 }
 
-/// Converts a binary [String] representation to a binary [int].
+/// Converts a binary [String] representation to a binary [int]. Ignores all '_' in the provided binary.
 ///
 /// Exactly equivalent to `int.parse(s, radix:2)`, but shorter to type.
-int bin(String s) => int.parse(s, radix: 2);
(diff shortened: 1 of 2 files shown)


lib/src/values/logic_value.dart
```diff
@@ -134,10 +134,10 @@ class LogicValue {
   }
 }
 
-/// Converts a binary [String] representation to a binary [int].
+/// Converts a binary [String] representation to a binary [int]. Ignores all '_' in the provided binary.
 ///
 /// Exactly equivalent to `int.parse(s, radix:2)`, but shorter to type.
```
mkorbel1, line 139: We should remove part from the doc comment since it's no longer exactly equivalent.  Move the "Ignores all ..." part to where this "int.parse..." line is, so that the first line remains shorter and the additional detail of ignoring _ is separated by a blank line.

mkorbel1 (changes requested):

mkorbel1 (approved):

PR #131, opened 2022-06-14
Fix for issue #126 Simulator reset does not wait for simulation to complete and test to reproduce the issue

### LINKED ISSUE #126: Simulator.reset does not wait for simulation to complete creating race condition
## Describe the bug
The `_simulationEndRequested` after reset is not set for run loop causing simulation to keep running even after it was asked to be ended.

## To Reproduce
Step 1) `simulator.endSimulation()`
Step 2) `simulator.reset()` before `simulationHasEnded`

## Expected behavior
After we call reset the simulation is not running anymore.

## Actual behavior
Previous simulation tries to run even after reset because it never ended.

madhuriakella:
## Description & Motivation

Simulator Bug found causing failures 

## Related Issue(s)

Fix #126 

## Testing

Added a new test to reproduce the issue

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/simulator.dart
+++ b/lib/src/simulator.dart
@@ -92,6 +92,8 @@ class Simulator {
   ///
   /// Note: values deposited on [Module]s from the previous simulation remain.
   static Future<void> reset() async {
+    if (_simulationEndRequested) await simulationEnded;
+
     _currentTimestamp = 0;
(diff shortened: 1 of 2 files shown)


mkorbel1 (approved):
Thank you, looks great!

Issue #16, opened 2021-09-23
Implement modulus operator %


mkorbel1:
Across applicable areas (gates, `Logic`, `LogicValues`, etc.)

PR #136, opened 2022-07-06
add modulo support

chaparalas:
## Description & Motivation

Add modulo support

## Related Issue(s)

[](https://github.com/intel/rohd/issues/16)

## Backwards-compatibility

NO

## Documentation

Yes

### DIFF
--- a/lib/src/logic.dart
+++ b/lib/src/logic.dart
@@ -287,6 +287,9 @@ class Logic {
   /// WARNING: Signed math is not fully tested.
   Logic operator /(dynamic other) => Divide(this, other).y;
 
+  /// Modulo operation.
+  Logic operator %(Logic other) => Modulo(this, other).y;
+
   /// Arithmetic right-shift.
   Logic operator >>(Logic other) => ARShift(this, other).y;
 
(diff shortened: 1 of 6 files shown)


test/logic_value_test.dart
```diff
@@ -493,29 +493,30 @@ void main() {
           () => LogicValue.ofString('0000') - LogicValue.ofString('000100'),
           throwsA(isA<Exception>()));
     });
-    test('muldiv', () {
```
mkorbel1, line 496: Add modulo in addition, instead of replacing multiplication and division tests.

lib/src/logic.dart
```diff
@@ -287,6 +287,9 @@ class Logic {
   /// WARNING: Signed math is not fully tested.
   Logic operator /(dynamic other) => Divide(this, other).y;
 
+  /// Modulo operation.
+  Logic operator %(Logic other) => Modulo(this, other).y;
```
mkorbel1, line 291: Add an example line to the README as well

mkorbel1 (changes requested):

test/logic_value_test.dart
chaparalas: Moved into arithmetic operations group

mkorbel1 (approved):
Looks great!  Thank you for your contribution!

PR #156, opened 2022-08-26
Improve performance of Combinational execute

### LINKED ISSUE #106: Optimize `Combinational._execute` X-driving
## Motivation
The `_execute` function in `Combinational` currently drives all outputs to `x`, then executes the conditionals.  This is less efficient than it could be, since `Conditional.execute` returns a list of signals driven.

## Desired solution
Only drive signals to `x` if they are not driven by any other sub-`Conditional` as determined by the `Conditional.execute` function.

mkorbel1:
## Description & Motivation

Improve the performance of `Combinational`'s `_execute` function so that it doesn't glitch extra `x`'s throughout the design unnecessarily.  For some designs this can provide a very meaningful performance boost to simulation speed.

## Related Issue(s)

Fix #106 

## Testing

Existing tests cover this.

Added new benchmarking code that covers a little bit of combinational logic via a pipeline example.

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -123,9 +123,11 @@ class Combinational extends _Always {
     }
 
     // combinational must always drive all outputs or else you get X!
-    for (var receiverOutputPair in _assignedReceiverToOutputMap.entries) {
-      if (!drivenLogics.contains(receiverOutputPair.key)) {
-        receiverOutputPair.value.put(LogicValue.x, fill: true);
+    if (_assignedReceiverToOutputMap.length != drivenLogics.length) {
+      for (var receiverOutputPair in _assignedReceiverToOutputMap.entries) {
+        if (!drivenLogics.contains(receiverOutputPair.key)) {
+          receiverOutputPair.value.put(LogicValue.x, fill: true);
+        }
       }
     }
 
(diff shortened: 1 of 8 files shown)


lib/src/modules/conditional.dart
```diff
@@ -117,13 +117,16 @@ class Combinational extends _Always {
 
     _isExecuting = true;
 
-    // combinational must always drive all outputs or else you get X!
-    for (var element in _assignedReceiverToOutputMap.values) {
-      element.put(LogicValue.x, fill: true);
+    var drivenLogics = <Logic>{};
+    for (var element in conditionals) {
+      drivenLogics.addAll(element.execute());
```
mkorbel1, line 122: Perhaps use some sort of counter/flag instead of constructing and combining sets would be better here.

mkorbel1: this is tricky to implement in a way that actually yields good performance boosts since each signal could be driven multiple times per `execute` and each `execute` could be called multiple times per timestamp.  the common case is that all items will be driven every cycle (in a "good" design), so the cost of iterating through all the flags would still be recurring.

this can be revisited if it turns out that the creation of the `Set`s is expensive enough to warrant a change

PR #171, opened 2022-09-28
resolves #153

### LINKED ISSUE #153: Access bits of a `Logic` by `Logic` index
## Motivation
Sometimes it's desirable to access a bit of a `Logic` signal by another `Logic`.  This is achievable with `>> n` and `& Const(1)`, but it would be nice to just use `[n]` directly to generate that code for ease of use and readability.

## Desired solution
Modify the overloaded `operator[]` on `Logic` so that it can accept either an `int` or another `Logic`, and appropriately generates hardware if necessary (if it's a `Logic`).

akshay-wankhede:
## Description & Motivation

Modified the overloaded operator[] on Logic so that it can accept either an int or another Logic, and appropriately generates hardware if necessary (if it's a Logic).

## Related Issue(s)

resolves #153 

## Testing

Added unit test: 'simcompare Index by Logic test'

## Backwards-compatibility

No. Fully backward compatible.

## Documentation

No.


lib/src/logic.dart
```diff
@@ -424,8 +424,12 @@ class Logic {
   }
 
   /// Accesses the [index]th bit of this signal.
```
mkorbel1, line 426: Add some clarification here that `index` could be either an int or a `Logic`.

lib/src/modules/gates.dart
```diff
@@ -513,3 +513,64 @@ class Mux extends Module with InlineSystemVerilog {
     return '$control ? $d1 : $d0';
   }
 }
+
+/// A generic two-input bit index gate [Module].
```
mkorbel1, line 517: This isn't really a "generic" gate anymore, it's a gate specifically for returning a single bit based on an index.  Let's clarify this doc comment so that just from reading it we can tell what this module does.

lib/src/modules/gates.dart
```diff
@@ -513,3 +513,64 @@ class Mux extends Module with InlineSystemVerilog {
...
+  late final String _a, _b, _y;
+
+  /// The primary input to this gate.
+  Logic get a => input(_a);
+
+  /// The bit index for this gate.
+  Logic get b => input(_b);
+
+  /// The output of this gate.
+  Logic get y => output(_y);
```
mkorbel1, line 530: Let's name these inputs and output signals better.  I know there's some poor naming on some other generic gates, but this one has very specific meaning.  Maybe `original`, `index`, and `selection` or something instead of a, b, y?

lib/src/modules/gates.dart
```diff
@@ -513,3 +513,64 @@ class Mux extends Module with InlineSystemVerilog {
...
   }
 }
+
+/// A generic two-input bit index gate [Module].
+///
+/// It always takes two inputs and has one output of width 1.
+class IndexGate extends Module with InlineSystemVerilog {
+  late final String _a, _b, _y;
+
+  /// The primary input to this gate.
```
mkorbel1, line 523: Let's not expose `a` and `b` inputs outside of this gate since they shouldn't really be accessed outside of the module.  I know some other modules do this, we should remove those too eventually.

lib/src/modules/gates.dart
```diff
@@ -513,3 +513,64 @@ class Mux extends Module with InlineSystemVerilog {
...
+  Logic get b => input(_b);
+
+  /// The output of this gate.
+  Logic get y => output(_y);
+
+  /// Constructs a two-input bit index gate for an abitrary custom functional implementation.
+  ///
+  /// The bit [_b] will be indexed as an output.
+  /// [Module] is in-lined as SystemVerilog, it will use _a[index], where index is a Const(_b)
+  IndexGate(Logic a, Logic b): super() {
```
mkorbel1, line 536: Naming in here for a and b should be fixed too.

lib/src/modules/gates.dart
```diff
@@ -513,3 +513,64 @@ class Mux extends Module with InlineSystemVerilog {
...
+
+  /// The bit index for this gate.
+  Logic get b => input(_b);
+
+  /// The output of this gate.
+  Logic get y => output(_y);
+
+  /// Constructs a two-input bit index gate for an abitrary custom functional implementation.
+  ///
+  /// The bit [_b] will be indexed as an output.
```
mkorbel1, line 534: The variable `_b` is private (since it starts with `_`) so shouldn't be referenced in the doc comment for a public constructor.  This sentence is a little confusing also, could be clarified like "[selected] is a one-bit [Logic] equal to the [index]th bit of [original]".  The first sentence about "arbitrary" doesn't make sense for this module.

lib/src/modules/gates.dart
```diff
@@ -513,3 +513,64 @@ class Mux extends Module with InlineSystemVerilog {
...
+  /// [Module] is in-lined as SystemVerilog, it will use _a[index], where index is a Const(_b)
+  IndexGate(Logic a, Logic b): super() {
+    _a = 'a_${a.name}';
+    _b = 'b_${b.name}';
```
mkorbel1, line 538: the index/b should be unpreferred since that can be fed in with an in-lined expression, but keep original/a as non-unpreferred.

lib/src/modules/gates.dart
```diff
@@ -513,3 +513,64 @@ class Mux extends Module with InlineSystemVerilog {
...
+    b.glitch.listen((args) {
+      _execute();
+    });
+  }
+
+  /// Executes the functional behavior of this gate.
+  void _execute() {
+      if (b.hasValidValue()) {
+          var indexVal = b.value.toInt();
+          y.put(a.getRange(indexVal, indexVal + 1).value);
```
mkorbel1, line 563: here you should first get the value of `a`, then get the subset of it.  when you call `getRange` on a `Logic`, it builds hardware, which means each glitch new hardware will get built, which is inefficient.  Instead, something like this should work which only operates on immutable values without generating new hardware:
```dart
y.put(a.value.getRange(indexVal, indexVal + 1));
```

test/gate_test.dart
```diff
@@ -421,5 +432,19 @@ void main() {
           gtm.generateSynth(), gtm.runtimeType.toString(), vectors);
       expect(simResult, equals(true));
     });
+
+    test('Index by Logic test', () async {
+      var gtm = IndexGateTestModule(Logic(width: 8), Logic(width: 8));
+      await gtm.build();
+      var vectors = [
+        Vector({'a': 14, 'b': 0}, {'a_indexedby_b': 0}),
+        Vector({'a': 14, 'b': 2}, {'a_indexedby_b': 1})
```
mkorbel1, line 441: add some vectors to test invalid a and/or invalid b produce an invalid output (some other tests have examples of this)

mkorbel1 (changes requested):

lib/src/logic.dart
```diff
@@ -424,8 +424,12 @@ class Logic {
...
-  Logic operator [](int index) {
-    return slice(index, index);
+  Logic operator [](dynamic index) {
+      if (index is Logic) {
+          return IndexGate(this, index).y;
+      }
+
```
mkorbel1, line 431: It would be a good idea to explicitly check if `index` is neither an `int` nor a `Logic` and throw and `Exception` in that case explaining that it only accepts int or Logic.

lib/src/logic.dart
```diff
@@ -423,9 +423,15 @@ class Logic {
...
-    return slice(index, index);
+  /// Accesses the [index]th bit of this signal, where index can be [int] or [Logic].
+  /// Throws Exception when index is not in range or index is neither [int] or [Logic].
+  Logic operator [](dynamic index) {
+    if (index is Logic) {
+      return IndexGate(this, index).selection;
+    } else if (index is int) {
+      return this.getRange(index, index + 1);
+    }
+    throw Exception('Expected [int] or [Logic]');
```
mkorbel1, line 434: I believe the bracket notation is used in doc comments, but not in Strings in things like Exceptions.  You could just put ticks around it instead:
```dart
    throw Exception('Expected `int` or `Logic`');
```

lib/src/logic.dart
```diff
@@ -423,9 +423,15 @@ class Logic {
     }
   }
 
-  /// Accesses the [index]th bit of this signal.
-  Logic operator [](int index) {
-    return slice(index, index);
+  /// Accesses the [index]th bit of this signal, where index can be [int] or [Logic].
+  /// Throws Exception when index is not in range or index is neither [int] or [Logic].
```
mkorbel1, line 427: Add a new-line between the summary and details of a documentation comment.  See: https://dart.dev/guides/language/effective-dart/documentation#do-separate-the-first-sentence-of-a-doc-comment-into-its-own-paragraph

mkorbel1 (changes requested):
Very close, just a couple little changes, looking good!

mkorbel1:
Make sure you run `dart format .` and `dart analyze` to ensure you pass the automated checks

mkorbel1:
PR #174 is adding a bunch more lint checks, which will probably flag some new issues for you if you get in after that (when you pull the changes in).  I'll wait a little bit to give you an opportunity to beat the PR in, but in the next day or two it will go in and then you'll have a little clean-up to do.

mkorbel1 (approved):
Looks great!

mkorbel1 (approved):
Looks good! Nice work!

PR #173, opened 2022-10-05
Rework script "run_coverage"

chykon:
## Description & Motivation

The old script called the "format_coverage" command, which accessed the ".packages" file, which will be [completely deprecated](https://dart.dev/tools/pub/cmd/pub-get#--legacy-packages-file) in Dart 2.19. Dart 2.18 requires either explicitly requesting the generation of ".packages" with a temporary flag, or removing the reference to the ".packages" file from the script.

The script has been rewritten in a more strict style that stops execution when an error occurs (the old script was executed on error to the end).

The location directory has also been changed to follow the [packaging convention](https://dart.dev/tools/pub/package-layout#internal-tools-and-scripts).

## Related Issue(s)

No.

## Testing

Run the script, wait for execution, open index.html

## Backwards-compatibility

No. I did not find any references to the old script in the code.

## Documentation

No.


mkorbel1 (approved):

I tried your branch out locally and everything seems to work.  I hadn't noticed the .packages change.

PR #178, opened 2022-10-12
Remove accidental `dontDeleteTmpFiles` for a bus test

mkorbel1:
## Description & Motivation

During development/debug, a `dontDeleteTmpFiles` was included on one of the tests which makes tests keep some temporary files around even when they pass.  This shouldn't be there.

## Related Issue(s)

N/A

## Testing

Existing tests pass.

## Backwards-compatibility

No

## Documentation

No


mkorbel1:
Might be good to add a check on https://github.com/intel/rohd/pull/177 that tmp_test directory is empty in the CI flow.

PR #181, opened 2022-10-14
Increase Dart SDK version

chykon:
## Description & Motivation

ROHD currently uses Dart version [2.14](https://github.com/intel/rohd/blob/6ec7ec56744c94e42d7011952fa3391bde031180/pubspec.yaml#L10). It is proposed to upgrade the version to the latest actual - [2.18](https://dart.dev/guides/language/evolution#dart-218). Among the changes, I would especially like to highlight the `super parameters` added in [2.17](https://dart.dev/guides/language/evolution#dart-217), which allow writing more compact constructors.

It also becomes possible to use `enhanced enums`, the idea of using which was mentioned here - #124.

After the version upgrade, it was necessary to use a `dart fix` to apply the recommendations related to the new language features.

## Related Issue(s)

No.

## Testing

After the changes, testing was launched, which was successfully completed.

## Backwards-compatibility

No.

## Documentation

No.

### DIFF
--- a/lib/src/modules/pipeline.dart
+++ b/lib/src/modules/pipeline.dart
@@ -101,7 +101,7 @@ class Pipeline {
       List<Logic> signals = const [],
       Map<Logic, Const> resetValues = const {},
       this.reset}) {
-    _stages = stages.map((stage) => _PipeStage(stage)).toList();
+    _stages = stages.map(_PipeStage.new).toList();
     _stages.add(_PipeStage((p) => [])); // output stage
 
     if (_numStages == 0) {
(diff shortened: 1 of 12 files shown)


lib/src/modules/pipeline.dart
```diff
@@ -101,7 +101,7 @@ class Pipeline {
       List<Logic> signals = const [],
       Map<Logic, Const> resetValues = const {},
       this.reset}) {
-    _stages = stages.map((stage) => _PipeStage(stage)).toList();
+    _stages = stages.map(_PipeStage.new).toList();
```
chykon, line 104: [About constructor tear-offs](https://medium.com/dartlang/dart-2-15-7e7a598e508a).

chykon: This is just an explanation. I wanted to add a comment to the commit, but it turned out like a review.

lib/src/values/small_logic_value.dart
```diff
@@ -41,12 +41,12 @@ class _SmallLogicValue extends LogicValue {
...
   }
 
-  const _SmallLogicValue(int value, int invalid, int width)
+  const _SmallLogicValue(int value, int invalid, super.width)
       : assert(width <= LogicValue._INT_BITS,
             '_SmallLogicValue should have low number of bits'),
         _value = ((1 << width) - 1) & value,
         _invalid = ((1 << width) - 1) & invalid,
-        super._(width);
+        super._();
```
chykon, line 49: I will duplicate the explanation, because the commit comment is not so easy to find.

Here there is an implicit transfer of [super parameters](https://github.com/dart-lang/language/blob/master/accepted/2.17/super-parameters/feature-specification.md#examples).

mkorbel1 (approved):
Thank you for your contribution! 

I believe this sort-of does break backwards compatibility since it requires users to upgrade their version of Dart installed to at least 2.18.0, but I think it's time to do so rather than sit at an old Dart version forever.  The next release will upgrade minor version (major still 0), indicating potentially backwards compatibility issues, so that's ok.

Issue #138, opened 2022-07-26
Custom definition names of `Module`s are not sanitized


mkorbel1:
## Describe the bug
It is possible to specify a `definitionName` that is not sanitary (as sanitized by `Sanitizer`), which means you can generate invalid SystemVerilog.

## To Reproduce
Define a module with an unsanitary `definitionName` passed to `Module`'s constructor.

## Expected behavior
ROHD sanitizes the definition name so that it generates valid SystemVerilog.

## Actual behavior
The unsanitary definition name is included directly in the generated SystemVerilog.

**Additional context**  
Sanitization may as well occur directly within the constructor.

quekyj:
Hi @mkorbel1 , I will have a look into this.

quekyj:
Issue resolved.

PR #190, opened 2022-10-24
resolves issue #138

quekyj:
## Description & Motivation
Fix bug of the ``definition name`` is not sanitized on ``Module`` class extend. 

## Related Issue(s)
Issue #138 

## Testing

- ``dart test`` passed on local. 
- Add a new unit test to test for the sanitization on the unclean variable name

## Backwards-compatibility

No

## Documentation

No


test/definition_name_test.dart
```diff
@@ -0,0 +1,121 @@
...
+
+class RunTimeNameModule extends ValidDefNameModule {
+  String mockRunTimeString;
+
+  RunTimeNameModule(super.a, super.defName, this.mockRunTimeString);
+  @override
+  String get definitionName {
+    const String? mockDefinitionName = null;
+
+    return Sanitizer.sanitizeSV(mockDefinitionName ?? mockRunTimeString);
```
mkorbel1, line 36: What do you test of your changes in `Module` if you override `definitionName`?

test/logic_name_test.dart
```diff
@@ -0,0 +1,38 @@
...
+    final mod = LogicTestModule(bus);
+    await mod.build();
+    final sv = mod.generateSynth();
+    expect(sv, contains('input logic validName'));
+  });
+
+  test('GIVEN logic name is invalid THEN expected to see sanitized name',
+      () async {
+    final bus = Logic(name: '&*-FinvalidN11Me');
+    final mod = LogicTestModule(bus);
```
mkorbel1, line 24: You should be able to immediately check that `bus.name` has the sanitized value.  There are scenarios where users will be interested in the name of a `Logic` even before they generate SV.

mkorbel1 (changes requested):

lib/src/module.dart
```diff
@@ -134,7 +135,17 @@ abstract class Module {
   ///
   /// This could become uniquified by a [Synthesizer] unless
   /// [reserveDefinitionName] is set.
-  String get definitionName => _definitionName ?? runtimeType.toString();
+  String get definitionName {
+    if (reserveDefinitionName && _definitionName == null) {
```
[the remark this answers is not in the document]
quekyj: create ``static`` function instead

lib/src/exception/name_exception.dart
```diff
@@ -0,0 +1,39 @@
+import 'package:rohd/rohd.dart';
```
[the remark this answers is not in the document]
quekyj: refer to ``modules.dart`` for export

test/definition_name_test.dart
quekyj:
```
class byte
```


create a reservedName class to test for runtime type

test/logic_name_test.dart
quekyj: No need to generate the system verilog code

lib/src/exception/name_exception.dart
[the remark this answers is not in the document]
quekyj: Add ``exceptions.dart`` but would prefer to only import specific exception in test to follow minimal to use rule of thumb. This method is still open to discussion.

test/logic_name_test.dart
quekyj: Indeed a better approach, the test should be able to run much faster than need to generate system verilog code.

test/definition_name_test.dart
quekyj: Added reserved name class and tested.

lib/src/exception/name_exception.dart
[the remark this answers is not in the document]
mkorbel1: That's a fair point.  The question is whether we would want to expose all the Exceptions whenever someone imports ROHD's package, and I think usually they probably don't since they won't need to throw their own versions of most of the exceptions.  What you have here should be good I think.

lib/src/exceptions/name_exceptions.dart
```diff
@@ -0,0 +1,48 @@
+/// Copyright (C) 2021 Intel Corporation
```
mkorbel1, line 1: Minor, but update copyright to 2022

lib/src/exceptions/exceptions.dart
```diff
@@ -0,0 +1,4 @@
+/// Copyright (C) 2021 Intel Corporation
```
mkorbel1, line 1: Update copyright to 2022

test/definition_name_test.dart
```diff
@@ -0,0 +1,102 @@
+import 'package:rohd/rohd.dart';
```
mkorbel1, line 1: Include header in this file as well

test/logic_name_test.dart
```diff
@@ -0,0 +1,29 @@
+import 'package:rohd/rohd.dart';
```
mkorbel1, line 1: include header

lib/src/module.dart
```diff
@@ -122,6 +123,18 @@ abstract class Module {
           '  Call build() before accessing this.');
   String _uniqueInstanceName;
 
+  static String? _nameValidation(
```
mkorbel1, line 126: Add a doc string describing what this function is for, even for private methods, it's helpful

lib/src/module.dart
```diff
@@ -122,6 +123,18 @@ abstract class Module {
...
+      String? definitionName, bool reserveDefinitionName) {
+    if (reserveDefinitionName && definitionName == null) {
+      throw NullReservedNameException();
+    } else if (reserveDefinitionName &&
+        Sanitizer.isSanitary(definitionName!) != true) {
```
mkorbel1, line 131: Could just be `!Sanitizer.isSanitary(definitionName!)`

test/definition_name_test.dart
```diff
@@ -0,0 +1,102 @@
...
+  group('GIVEN that Reserved Definition Name is TRUE,', () {
+    test('WHEN definition name is valid, THEN expect to compile successfully.',
+        () async {
+      final defName =
+          DefinitionName(name: DefinitionName.getValidName(), isReserved: true);
+      final mod = ValidDefNameModule(Logic(), defName);
+      await mod.build();
+      final sv = mod.generateSynth();
+      // Then it should return compile successfully
+      expect(sv, contains('module specialName('));
```
mkorbel1, line 44: Just like we did for the `Logic` name tests, we can just check `mod.definitionName` immediately rather than generating the SV and parsing it

mkorbel1 (changes requested):
Functionally this all looks good!  Just a few more clean-up things.

lib/src/module.dart
quekyj: Any template or script is available for doc string documentation? Maybe can create a template on documentation style to make sure it is similar across all the doc.

lib/src/module.dart
```diff
@@ -123,12 +123,17 @@ abstract class Module {
           '  Call build() before accessing this.');
   String _uniqueInstanceName;
 
+  /// Return string type definition name if validation passed
+  /// else throw exception.
+  ///
+  /// This validation method ensure that definition name is valid if
```
mkorbel1, line 129: You could rewrite this as
```dart
  /// This validation method ensure that [definitionName] is valid if
  /// [reserveDefinitionName] set to `true`.
```

mkorbel1 (approved):

PR #192, opened 2022-11-01
Fix bug related to clock divider simulation, fix #191

### LINKED ISSUE #191: `Sequential` outputs as clock inputs to other `Sequential`s trigger late in the simulator
### Describe the bug

If the output of one `Sequential` is used as a clock trigger to another `Sequential`, the phased simulator can sometimes detect the edge and drive the outputs of the downstream `Sequential` one tick late.

This breaks clock dividers.

### To Reproduce

Create a clock divider and connect a flop to the output of the divided clock.

### Expected behavior

Divided clock properly triggers edge detection on `Sequential`s.

### Actual behavior

Divided clock doesn't trigger edges properly, causing incorrect `Sequential` behavior.

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

_No response_

mkorbel1:
## Description & Motivation

If the output of one Sequential is used as a clock trigger to another Sequential, the phased simulator can sometimes detect the edge and drive the outputs of the downstream Sequential one tick late.

This is because there was a gate in the `Sequential` clock sampling logic that ignored clock toggles during the `clkStable` phase.  Instead, we should immediately execute the flop if we're already in `clkStable`.

## Related Issue(s)

Fix #191 

## Testing

Added a new test where two counters are driven by an original clock and that clock divided, respectively.

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -336,14 +336,22 @@ class Sequential extends _Always {
       clk.glitch.listen((event) async {
         // we want the first previousValue from the first glitch of this tick
         _preTickClkValues[i] ??= event.previousValue;
-        if (!_pendingExecute) {
-          unawaited(Simulator.clkStable.first.then((value) {
-            // once the clocks are stable, execute the contents of the FF
-            _execute();
-            _pendingExecute = false;
-          }));
+
+        if (Simulator.phase == SimulatorPhase.clkStable) {
+          // this could be an output of a flop driving the clock of this
+          // flop, so execute immediately (e.g. clock divider)
+          _execute();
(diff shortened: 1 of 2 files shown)


chaparalas (approved):
Consider adding a async-reset clk-div testcase.

mkorbel1:
> Consider adding a async-reset clk-div testcase.

Added in latest commit, good idea, thanks

mkorbel1:
Need to add more testing around chains of clock dividers, which are broken with current implementation in this PR.  Likely solution is to use topological sort to pick the order of `Sequential` execution.

PR #193, opened 2022-11-03
Add table of content to Readme.md for easier navigation

quekyj:
## Description & Motivation

The documentation in the Readme.md is hard to navigate around. Adding Table of content allow users to have a glance of the entire documentation and quickly navigate to the desire topic or doc.

## Related Issue(s)

None

## Testing

None. But Require @mkorbel1 to review. 

## Backwards-compatibility

No

## Documentation

Ya. Table of content added to the Readme.md.


mkorbel1:
Do you think this is needed on top of the built-in table of contents from GitHub?
https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/

Are there limitations to the generated one that you specifically think this improves upon?

The two downsides I see to adding this manually are:
1. The first thing people see is a table of contents instead of a quick description and feature list.
2. There's an additional maintenance burden each time README is updated to ensure the table of contents is accurate

quekyj:
I see. Don't know that GitHub already have built in table of content. If that the case, we can stick with GitHub built-in. But do you think that we should add some instruction on how to access table of content somewhere? Just worry others that are not familiar with this GitHub feature might unable to find the table of content.

PR #195, opened 2022-11-08
[FIX-ISSUE-7] Creating isIn function to check if an Logic is in array

RPG-coder-intc:
The Array can be a mix of Logic, LogicValue, int, bool & BigInt.
This function belongs to Logic Class

## Description & Motivation

Inclusion of IsIn() function 

[TODO]
- [x] Is the function results synthesized on hardware?
- [ ] Test coverage logic_test

## Related Issue(s)
For more details please refer to, ISSUE #7 

## Testing
- [ ] Yet to decide...

## Documentation
- [ ] Write Documentation

### DIFF
--- a/lib/src/logic.dart
+++ b/lib/src/logic.dart
@@ -592,4 +592,25 @@ class Logic {
       getRange(0, startIndex),
     ].swizzle();
   }
+
+  /// Returns true if the [Logic] calling this function is in [List] else false.
+  ///
+  /// The list can be [Logic] or [int] or [bool] or [BigInt] or mix of all
+  /// ([dynamic]) supplied
+  ///
+  bool isIn(List<dynamic> list) {
+    /* 
+      Iterate through list of Logic, LogicVal, int or bool. 
+      Homogeneous (all elements same) or Hetrogeneous (mix of types) array 
+    */
+    for (final logicVar in list) {
+      if ((logicVar is Logic && logicVar.value == value) ||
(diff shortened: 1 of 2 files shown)


lib/src/logic.dart
```diff
@@ -592,4 +592,25 @@ class Logic {
       getRange(0, startIndex),
     ].swizzle();
   }
+
+  /// Returns true if the [Logic] calling this function is in [List] else false.
+  ///
+  /// The list can be [Logic] or [int] or [bool] or [BigInt] or mix of all
+  /// ([dynamic]) supplied
+  ///
+  bool isIn(List<dynamic> list) {
```
mkorbel1, line 601: Since this operates on a `Logic`, we want this function to return a 1-bit `Logic` rather than a `bool`.

lib/src/logic.dart
```diff
@@ -592,4 +592,25 @@ class Logic {
...
+    /* 
+      Iterate through list of Logic, LogicVal, int or bool. 
+      Homogeneous (all elements same) or Hetrogeneous (mix of types) array 
+    */
+    for (final logicVar in list) {
+      if ((logicVar is Logic && logicVar.value == value) ||
```
mkorbel1, line 607: I think you can leverage the `Logic.eq` function which already handles either `Logic` or other constant values and wrapping in a `Const`

mkorbel1 (changes requested):

Issue #141, opened 2022-07-27
Add support for `++` and `--` types of operations in conditional logic


mkorbel1:
## Motivation
There are cases in procedural execution in `_Always` blocks where `++` and `--` are useful.  Dart does not allow overloading these operators, however.

## Desired solution
Some sort of function on `Logic` which acts like `++` and `--`, both postfix and prefix (i.e. `++x` vs. `x++`).  Some thought has to go into the API and function names to make this intuitive.

To effectively use these requires some careful thought.  A prefix is simple, i.e. `++x` is the same thing as `x + 1`.  A postfix requires that the original signal is returned, but then a modified conditional operation is performed afterwards.  This means that it needs to return some special type of object that other `Conditional`s can handle, but that also might need to play nicely with other `Logic` manipulation within conditional procedural logic.  There's a number of combinations to consider, and this could involve some modifications to the way things like `ConditionalAssignment`s work in general.

## Alternatives considered  
Just omit the feature.

## Additional details
There are real designs where users had wished for a feature like this.

mkorbel1:
Adding something like += and -= would also be good!

RPG-coder-intc:
Hey Max, I would like to give a try on this issue! :)

mkorbel1:
@RPG-coder-intc great!  Let me know once you have a strategy mapped out if you'd like to review the approach.  I expect there could be some tricky surprises in implementing this.

RPG-coder-intc:
A function will be made for postIncrement and preIncrement taking the Logic object value as a reference. The result computation will be worked on this reference object taken as a parameter.

RPG-coder-intc:
Similar pattern will be followed on `+=` (`addAssign(&ref)`) , -= (`diffAssign(&ref)`) , `*=` (`mulAssign(&ref)`), `/=` (`divAssign(&ref)`).

RPG-coder-intc:
`logicVar++` ===  `postIncr(&ref)`
`++logicVar` ===  `preIncr(&ref)`

chykon:
Perhaps it's better to use `subAssign` instead of `diffAssign`?

RPG-coder-intc:
> Perhaps it's better to use `subAssign` instead of `diffAssign`?

It can be done. What's your say @mkorbel1

mkorbel1:
[...]

PR #196, opened 2022-11-08
[Fix-issue-141] Adding Logic functionality for incr(), decr(), mulAssign() & divAssign() operation

RPG-coder-intc:
## Description & Motivation
Adding functionality to Logic class
- `incr(Logic? a=Const(1))` - Pre Increment. That is shorthand for addition (default by 1) and assigning  back to same Logic
- `decr(Logic? a=Const(1))` - Pre Devrement. That is shorthand for subtraction (default by 1) and assigning  back to same Logic
- `mulAssing(Logic a)` - Shorthand for multiply and assign back to same Logic
- `divAssing(Logic a)` - Shorthand for division and assign back to same Logic

All function above returns ConditionalAssignment obj

## Related Issue(s)
For more details refer to, Issue #141 

## Testing
- [x] part of Shorthand Assignment Module

## Documentation
- [x] Written Docs for incr, decr, mulAssign, divAssign


lib/src/logic.dart
```diff
@@ -575,6 +575,31 @@ class Logic {
   /// Greater-than-or-equal-to.
   Logic operator >=(dynamic other) => GreaterThanOrEqual(this, other).out;
 
+  /// Pre-Increment (++var) or var+=incrVal
+  // ignore: prefer_expression_function_bodies
```
mkorbel1, line 579: Shouldn't ignore these lints

lib/src/logic.dart
```diff
@@ -575,6 +575,31 @@ class Logic {
...
+
+  /// Shorthand for multiplication assign (++var)
+  // ignore: prefer_expression_function_bodies
+  ConditionalAssign mulAssign(Logic mulVal) {
+    return this < this * mulVal;
+  }
+
+  /// Pre-Increment (++var)
+  // ignore: prefer_expression_function_bodies
+  ConditionalAssign divAssign(Logic? divVal) {
```
mkorbel1, line 598: this shouldn't be a nullable for `divVal`?

mkorbel1 (changes requested):

lib/src/logic.dart
```diff
@@ -575,30 +575,88 @@ class Logic {
   /// Greater-than-or-equal-to.
   Logic operator >=(dynamic other) => GreaterThanOrEqual(this, other).out;
 
-  /// Pre-Increment (++var) or var+=incrVal
-  // ignore: prefer_expression_function_bodies
-  ConditionalAssign preIncr({Logic? incrVal}) {
-    return this < ((incrVal != null) ? this + incrVal : this + Const(1));
-  }
+  /// Shorthand Increment.
```
mkorbel1, line 578: Let's elaborate a bit more in the doc comments.  Something like "Shorthand for a [Conditional] which increments `this` by [incrVal]".  Make sure the first line of the doc comment has a blank line afterwards.
```dart
/// Short descriptions
///
/// Further details down here
void myFunction...
```

In general, for these doc strings, try to make them full english sentences that clearly explain the purpose of the API.

lib/src/logic.dart
```diff
@@ -575,6 +575,89 @@ class Logic {
   /// Greater-than-or-equal-to.
   Logic operator >=(dynamic other) => GreaterThanOrEqual(this, other).out;
 
+  /// Shorthand Increment.
+  /// By default for a [Logic] var, if no [incrVal] is provided result is ++var
+  /// else result is var+=[incrVal]
+  ///
+  /// ```dart
+  ///
+  /// Logic a = Logic(width: 8);
```
mkorbel1, line 584: In the example, you don't need to add inputs and outputs, the only part that's relevant is that they are `Logic` type

lib/src/logic.dart
```diff
@@ -575,6 +575,89 @@ class Logic {
...
+  /// final piOut = addOutput('piOut', width: 8);
+  ///
+  /// Combinational([
+  ///   piOut < a,
+  ///   piOut.incr(b),
+  /// ]);
+  ///
+  /// ```
+  ///
+  ConditionalAssign incr([Logic? incrVal]) =>
```
mkorbel1, line 597: It would be nice to support `dynamic` here, since we could want to add by an `int`, `LogicValue`, or `BigInt` as well.  I think all of the math operations (`+-/*`) will already support `dynamic` and properly convert as needed.

mkorbel1 (changes requested):

test/conditionals_test.dart
```diff
@@ -11,6 +11,34 @@ import 'package:rohd/rohd.dart';
...
+    final pdOut = addOutput('pdOut', width: 8);
+    final maOut = addOutput('maOut', width: 8);
+    final daOut = addOutput('daOut', width: 8);
+
+    Combinational([
+      piOut < preIncr,
+      pdOut < preDecr,
+      maOut < mulAssign,
+      daOut < divAssign,
+      piOut.incr(b),
```
mkorbel1, line 34: Need to add tests with `incr()` and `decr()` with no value passed

lib/src/logic.dart
```diff
@@ -575,6 +575,74 @@ class Logic {
...
+  /// // Given a and b Logic input and piOut as output
+  /// Combinational([
+  ///   piOut < a,
+  ///   piOut.incr(b),
+  /// ]);
+  ///
+  /// ```
+  ///
+  ConditionalAssign incr([dynamic incrVal]) =>
+      this < ((incrVal != null) ? this + incrVal : this + Const(1));
```
mkorbel1, line 594: Could use this to make it shorter and more readable
`this < (this + incrVal ?? 1)`

lib/src/logic.dart
```diff
@@ -575,6 +575,74 @@ class Logic {
   /// Greater-than-or-equal-to.
   Logic operator >=(dynamic other) => GreaterThanOrEqual(this, other).out;
 
+  /// Shorthand for a [Conditional] which increments this by [incrVal]
+  ///
+  /// By default for a [Logic] var, if no [incrVal] is provided result is ++var
```
mkorbel1, line 580: Can't name a variable `var` in Dart

Would also be nice to add `'s around the "var" so that it's clear its a variable

mkorbel1 (changes requested):

mkorbel1 (approved):
Looks great!

PR #197, opened 2022-11-10
[Issue-114] Add test to sequential driver on same signal to return Exception

### LINKED ISSUE #114: Add a test against multiple sequential drivers on the same signal
## Motivation
It doesn't make sense for a signal to be driven by two drivers in a `Sequential` block.  The code already prevents this, but testing is incomplete.

## Desired solution
Add a test ensuring that an exception is thrown when there are multiple drivers for a flop.

quekyj:
## Description & Motivation

It doesn't make sense for a signal to be driven by two drivers in a Sequential block. The code already prevents this, but testing is incomplete.

Also fixed bug on the previous exception won't run.

## Related Issue(s)

Fix #114 

## Testing
``dart test`` complete successfully. New test added to the ``conditionals_test.dart``.

```
test(
      'should return SignalRedrivenException when there are multiple drivers '
      'for a flop.', () async {
    final mod = SignalRedrivenSequentialModule(
        Logic(), Logic(), Logic(width: 8), Logic(width: 8));
    await mod.build();
    final vectors = [
      Vector({'a': 1, 'd': 1}, {}),
      Vector({'a': 0, 'b': 0, 'd': 2}, {'q': 1}),
    ];

    try {
      fail('Exception not thrown!');
      // ignore: avoid_catches_without_on_clauses
    } catch (e) {
      expect(e.runtimeType, equals(SignalRedrivenException));
    }
  });
```

## Backwards-compatibility

Nope.

## Documentation

Nope.


lib/src/modules/conditional.dart
```diff
@@ -524,10 +520,13 @@ class ConditionalAssign extends Conditional {
   @override
   List<Conditional> getConditionals() => [];
 
+  /// Assign [receiver] the value of the [driver].
+  ///
+  /// The [drivenSignals] is a collection of [receiver] in a Set.
```
mkorbel1, line 525: Since this is overridden via `@override`, it will inherit the doc string from the super class, so you dont need to add here.

lib/src/modules/conditional.dart
```diff
@@ -617,15 +616,12 @@ class Case extends Conditional {
...
-  Set<Logic> execute() {
-    final drivenLogics = <Logic>{};
-
+  void execute(Set<Logic> drivenSignals) {
     if (!expression.value.isValid) {
       // if expression has X or Z, then propogate X's!
       for (final receiver in getReceivers()) {
         receiverOutput(receiver).put(LogicValue.x);
       }
-      return {};
```
mkorbel1, line 628: You still want to return here, but just nothing (`return;`) because you dont want to continue down

test/conditionals_test.dart
```diff
@@ -159,6 +161,43 @@ class SequentialModule extends Module {
   }
 }
 
+class SignalRedrivenSequentialModule extends Module {
```
mkorbel1, line 164: This module could be a lot simpler for just testing redriven things, don't need so many irrelevant signals.

test/conditionals_test.dart
```diff
@@ -253,4 +292,24 @@ void main() {
...
+    await mod.build();
+    final vectors = [
+      Vector({'a': 1, 'd': 1}, {}),
+      Vector({'a': 0, 'b': 0, 'd': 2}, {'q': 1}),
+    ];
+
+    try {
+      await SimCompare.checkFunctionalVector(mod, vectors);
+      fail('Exception not thrown!');
+      // ignore: avoid_catches_without_on_clauses
```
mkorbel1, line 310: Why ignore this lint?

mkorbel1 (changes requested):

lib/src/types/redriven_monitor_set.dart
```diff
@@ -0,0 +1,62 @@
+/// Copyright (C) 2022 Intel Corporation
```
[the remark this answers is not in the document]
quekyj: On second thought, I think we can move  ``traversable_collection.dart`` and ``synchronous_propagator.dart`` into type as new type might be added again in the future. Besides, by doing this we can make sure that ``utilities`` responsibility is scope to only provide functional support while ``type`` is the custom collection created.

lib/src/types/redriven_monitor_set.dart
```diff
@@ -0,0 +1,62 @@
...
+    _set.addAll(elements);
+  }
+
+  /// The duplicate members in the collection
+  ///
+  /// Returns an [UnmodifiableSetView] if the collection contains duplicates
+  UnmodifiableSetView<T> get getDuplicates => UnmodifiableSetView(_duplicates);
+
+  /// Returns `true` if collection contains duplicates
+  bool get isDuplicates => _duplicates.isNotEmpty;
```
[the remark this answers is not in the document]
quekyj: Ya, hasDuplicates look better.

test/conditionals_test.dart
quekyj: Replace with: 
```
on Exception catch (e):
```

test/conditionals_test.dart
quekyj: Oh ya, thanks for the reminder! 👍

lib/src/simulator.dart
```diff
@@ -55,6 +55,9 @@ class Simulator {
   /// Tracks whether an end to the active simulation has been requested.
   static bool _simulationEndRequested = false;
 
+  /// Track if exception is thrown during the execution
+  static Exception? exception;
```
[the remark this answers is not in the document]
quekyj: Tried, but look like the exception is unhandled. In this case, the exception must be thrown in the ``Simulator.run()`` in order to catch.

```
00:03 +62 -1: test/conditionals_test.dart: should return SignalRedrivenException when there are multiple drivers for a flop. [E]                                
  Sequential drove the same signal(s) multiple times: {Logic(8): q}
  package:rohd/src/modules/conditional.dart 384:9   Sequential._execute
  package:rohd/src/modules/conditional.dart 344:13  Sequential._setup.<fn>.<fn>
```

Also tested by changing ``unwaited`` to ``await`` in conditionals, but its doesn't fix but instead making other test fail as well.

lib/src/modules/conditional.dart
```diff
@@ -341,6 +344,10 @@ class Sequential extends _Always {
             // once the clocks are stable, execute the contents of the FF
             _execute();
             _pendingExecute = false;
+          }).catchError((dynamic err) {
```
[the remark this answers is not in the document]
quekyj:
```
if (!o.value.isValid) {
  // invalid value causes exception without helpful message,
  // so throw it
  throw Exception(errorReason);
expect(o.value.toInt(), equals(value), reason: errorReason);
```

@mkorbel1 Any scenario of a value is ``int`` but ``invalid``? The code is throwing the same error message. Maybe can use other message for this.

quekyj:
- Create a function to stop the simulation and collect the stack trace. 
- Test for stack trace:
  - Instead of printing it, catch the logger message or make it through ``SimulatorException`` that received stack trace.
  - Look like the StackTraceClass https://api.dart.dev/stable/2.18.4/dart-core/StackTrace-class.html

lib/src/simulator.dart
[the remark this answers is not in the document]
quekyj: Added the ``throwException()`` function.

lib/src/modules/conditional.dart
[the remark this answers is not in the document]
quekyj: Fixed by added test.

lib/src/exceptions/conditionals/signal_redriven_exception.dart
```diff
@@ -0,0 +1,28 @@
...
+///
+/// 2022 November 9
+/// Author: Yao Jing Quek <yao.jing.quek@intel.com>
+///
+
+import 'package:rohd/rohd.dart';
+
+/// Throws [SignalRedrivenException] whenever a [Logic] signal
+/// is redriven for mutiple times in [Conditional].
+///
```
mkorbel1, line 16: remove extra blank line

lib/src/exceptions/sim_compare/mismatch_output_value_exception.dart
```diff
@@ -0,0 +1,35 @@
...
+///
+
+import 'package:rohd/rohd.dart';
+
+import 'package:rohd/src/utilities/simcompare.dart';
+
+/// Throws [MismatchOutputValueException] whenever the vectors
+/// expected from simulator comparison is difference from
+/// the output from [Module] simulated.
+///
```
mkorbel1, line 19: delete extra blank line

lib/src/exceptions/conditionals/signal_redriven_exception.dart
```diff
@@ -0,0 +1,28 @@
...
+class SignalRedrivenException implements Exception {
+  late final String _message;
+
+  /// constructor for SignalRedrivenException,
```
mkorbel1, line 20: Start sentences with a capital letter, end with a period, in general try to be careful to have good grammar since this is documentation.

https://dart.dev/guides/language/effective-dart/documentation

lib/src/exceptions/name/invalid_reserved_name_exception.dart
```diff
@@ -36,7 +18,7 @@ class InvalidReservedNameException implements Exception {
   late final String _message;
 
   /// constructor for InvalidReservedNameException,
```
mkorbel1, line 20: Documentation for functions and constructors should describe what it *does*, not what it *is*.  

E.g.
```dart
/// Constructs an exception indicating that a reserved name is not valid.
///
/// The [message] explains the reasoning behind this exception being thrown.
```

lib/src/exceptions/name/null_reserved_name_exception.dart
```diff
@@ -0,0 +1,28 @@
...
+///
+/// null_reserved_name_exception.dart
+/// An exception that is thrown when reserved name is null
+///
+/// 2022 November 15
+/// Author: Yao Jing Quek <yao.jing.quek@intel.com>
+///
+import 'package:rohd/rohd.dart';
+
+/// This Exception show that reserved name eg. definitionName is NULL but
```
mkorbel1, line 12: show -> shows
NULL -> `null`

Put tick's (`) around variable names and values that aren't in scope.  Use brackets ([]) around in-scope variables.

lib/src/logic.dart
```diff
@@ -12,9 +12,10 @@ import 'dart:async';
 
 import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
+
 import 'package:rohd/rohd.dart';
+import 'package:rohd/src/types/synchronous_propagator.dart';
```
mkorbel1, line 17: I'm rethinking calling this folder "types", sorry.  I think it may confuse users since it sounds like these might be types of other things such as hardware constructs.  Maybe let's keep synchronous_propagator in utilities/ and we can rename this folder to something like "collections"?

lib/src/modules/conditional.dart
```diff
@@ -316,14 +318,24 @@ class Sequential extends _Always {
...
+                  // once the tick has completed,
+                  // we can update the override maps
+                  for (final driverInput in _driverInputsPendingPostUpdate) {
+                    _inputToPreTickInputValuesMap[driverInput] =
+                        driverInput.value;
+                  }
+                  _driverInputsPendingPostUpdate.clear();
+                  _pendingPostUpdate = false;
+                },
+              ).catchError((dynamic err, dynamic stackTrace) {
```
mkorbel1, line 333: From the docs:
```
  // The `Function` below stands for one of two types:
  // - (dynamic) -> FutureOr<T>
  // - (dynamic, StackTrace) -> FutureOr<T>
```

It also looks like you can use the `test` argument to check whether it is an exception so you don't need the if statement. 
`test: (Object err) => err is Exception`

That means I think the signature should be:
`(Exception err, StackTrace stackTrace)`

lib/src/simulator.dart
```diff
@@ -265,6 +271,12 @@ class Simulator {
     _simulationEndRequested = true;
   }
 
+  /// End Simulation and return the exception
+  static void throwException(Exception err, StackTrace stacktrace) {
+    _exception = err;
```
mkorbel1, line 276: As implemented, this only records the last `throwException` before the simulator finally raises the exception.  You probably either want to save the *first* exeption, or all of them.  If we capture just the first, then we should think what to do if there's more.

lib/src/simulator.dart
```diff
@@ -55,6 +55,11 @@ class Simulator {
   /// Tracks whether an end to the active simulation has been requested.
   static bool _simulationEndRequested = false;
 
+  /// Track if exception is thrown during the execution
+  static Exception? _exception;
+
+  static StackTrace? _stacktrace;
```
mkorbel1, line 61: Add a doc comment

lib/src/simulator.dart
```diff
@@ -125,6 +130,7 @@ class Simulator {
 
     _currentTimestamp = 0;
     _simulationEndRequested = false;
+    _exception = null;
```
mkorbel1, line 133: clear the stacktrace too

lib/src/types/duplicate_detection_set.dart
```diff
@@ -0,0 +1,67 @@
...
+import 'dart:core';
+
+import 'package:collection/collection.dart';
+
+/// A Set collection that monitor for duplication
+///
+/// [DuplicateDetectionSet] can be used if duplicate element are needed to be
+/// catch for certain usage.
+class DuplicateDetectionSet<T> extends SetBase<T> {
+  final Set<T> _set = <T>{};
```
mkorbel1, line 20: Add some doc comments, even though they are private

lib/src/types/duplicate_detection_set.dart
```diff
@@ -0,0 +1,67 @@
...
+  Iterator<T> get iterator => _set.iterator;
+
+  @override
+  int get length => _set.length;
+
+  @override
+  T? lookup(Object? element) => _set.lookup(element);
+
+  @override
+  bool remove(Object? value) {
```
mkorbel1, line 58: Add a special doc comment explaining that you can only remove if it never had any duplicates, and that it would otherwise throw an exception.

lib/src/utilities/simcompare.dart
```diff
@@ -107,14 +108,15 @@ abstract class SimCompare {
...
               final errorReason =
                   'For vector #${vectors.indexOf(vector)} $vector,'
                   ' expected $o to be $value, but it was ${o.value}.';
               if (value is int) {
                 if (!o.value.isValid) {
                   // invalid value causes exception without helpful message,
                   // so throw it
-                  throw Exception(errorReason);
+                  throw MismatchOutputValueException(vectors, vector, o, value);
                 }
```
mkorbel1, line 120: We could do `expect(o.value.isValid, isTrue)` instead of throwing an exception here

mkorbel1 (changes requested):

lib/src/exceptions/conditionals/signal_redriven_exception.dart
quekyj: Alright, noted.

quekyj: ### References:
Below are references for documentation guidelines. 

```
[1] https://github.com/flutter/flutter/blob/74cfc3db67dc89d617ebccd2763301777106c7fd/packages/flutter_driver/lib/src/common/wait.dart
[2] https://github.com/flutter/flutter/blob/19eab105a57e519753af9556b045eb1b0ccae9fd/packages/flutter_tools/lib/src/build_system/exceptions.dart#L11
[3] https://dart.dev/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs
```

lib/src/exceptions/name/invalid_reserved_name_exception.dart
quekyj: Alright, updated the doc.

lib/src/logic.dart
quekyj: Yup, sounds good.

lib/src/simulator.dart
quekyj: I changed to List type. Not sure it will help.

lib/src/simulator.dart
```diff
@@ -271,10 +274,12 @@ class Simulator {
...
 
-  /// End Simulation and return the exception
-  static void throwException(Exception err, StackTrace stacktrace) {
-    _exception = err;
-    _stacktrace = stacktrace;
+  /// Collect [exception] and [stacktrace] and add to [_simExceptions] `List`.
+  ///
+  /// Exceptions collected will be logged and thrown during the [run] process
+  /// on simulation.
+  static void throwException(Exception exception, StackTrace stacktrace) {
```
mkorbel1, line 281: Should be `stackTrace`

lib/src/simulator.dart
```diff
@@ -271,10 +274,12 @@ class Simulator {
     _simulationEndRequested = true;
   }
 
-  /// End Simulation and return the exception
-  static void throwException(Exception err, StackTrace stacktrace) {
-    _exception = err;
-    _stacktrace = stacktrace;
+  /// Collect [exception] and [stacktrace] and add to [_simExceptions] `List`.
```
mkorbel1, line 277: The `_simExceptions` list is private, and shouldn't be referenced in the public API docs.  Rather, describe the behavior of this function from the user's perspective.  Something like:
"Collects an [exception] and associated [stackTrace] triggered asynchronously during simulation to be thrown synchronously by [run].  Calling this function will end the simulation after this [tick] completes".

lib/src/simulator.dart
```diff
@@ -286,15 +291,17 @@ class Simulator {
...
+        _simExceptions.isEmpty &&
         !_simulationEndRequested &&
         (_maxSimTime < 0 || _currentTimestamp < _maxSimTime)) {
       await tick();
     }
 
-    if (_exception != null) {
-      logger.severe(_exception.toString(), _exception, _stacktrace);
-      throw _exception!;
+    if (_simExceptions.isNotEmpty) {
```
mkorbel1, line 300: No need to check if it's not empty, since you're already looping through a list which may be empty

lib/src/simulator.dart
```diff
@@ -310,3 +317,23 @@ class Simulator {
     await simulationEnded;
   }
 }
+
+/// A simulator exception that produces object of exception and stack trace.
+class _SimulatorException {
+  /// Tracks for [Exception] thrown during [Simulator] `run()`.
+  final Exception _exceptions;
```
mkorbel1, line 324: should be `_exception` and `_stackTrace` since these are singular

lib/src/simulator.dart
```diff
@@ -310,3 +317,23 @@ class Simulator {
...
+
+  /// Tracks for [StackTrace] thrown during [Simulator] `run()`.
+  final StackTrace _stackTraces;
+
+  /// Constructs a simulator exception, using [exception] and [stackTrace].
+  _SimulatorException(Exception exception, StackTrace stackTrace)
```
mkorbel1, line 330: You can just write `_SimulatorException(this.exception, this.stackTrace)`

mkorbel1 (changes requested):

mkorbel1 (changes requested):

lib/src/simulator.dart
```diff
@@ -1,3 +1,4 @@
+// ignore_for_file: public_member_api_docs, sort_constructors_first
```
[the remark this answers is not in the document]
quekyj: weird, not sure when did i add this.

lib/src/exceptions/sim_compare/mismatch_output_value_exception.dart
```diff
@@ -0,0 +1,36 @@
+/// Copyright (C) 2022 Intel Corporation
```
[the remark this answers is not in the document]
quekyj: Ya, we can do that.

lib/src/modules/conditional.dart
```diff
@@ -956,23 +961,21 @@ class If extends Conditional {
...
       }
     } else if (driverValue(condition)[0] == LogicValue.zero) {
       for (final conditional in orElse) {
-        drivenLogics.addAll(conditional.execute());
+        conditional.execute(drivenSignals);
       }
     } else {
       // x and z propagation
       for (final receiver in getReceivers()) {
         receiverOutput(receiver).put(driverValue(condition)[0]);
```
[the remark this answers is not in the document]
quekyj:
```
00:04 +124 -1: test/translations_test.dart: simcompare translation [E]                                                                                                     
  Sequential drove the same signal(s) multiple times: {Logic(16): storageBank_0, Logic(16): storageBank_1, Logic(16): rdData0, Logic(16): rdData1, Logic(16): storageBank_2, Logic(16): storageBank_3, Logic(16): storageBank_4, Logic(16): storageBank_5, Logic(16): storageBank_6, Logic(16): storageBank_7}
  package:rohd/src/simulator.dart 301:7             Simulator.run
  ===== asynchronous gap ===========================
  package:rohd/src/utilities/simcompare.dart 145:5  SimCompare.checkFunctionalVector
  ===== asynchronous gap ===========================
  test/translations_test.dart 138:7                 main.<fn>.<fn>
```

Error after added the check.

test/conditionals_test.dart
```diff
@@ -396,4 +411,21 @@ void main() {
...
     }
   });
+
+  test(
+      'should return SignalRedrivenException when driven with '
+      'x signals and valid signals.', () async {
+    final mod = SignalRedrivenSequentialModuleWithX(Logic(), Logic(), Logic());
+    await mod.build();
+    final vectors = [
+      Vector({'a': 1, 'd': 1, 'c': 1}, {'b': LogicValue.z}),
```
mkorbel1, line 421: This would only be `z` the first cycle, since this is a `Sequential` in the test.

I think this case should hit the re-driven exception if `a` and `d` are both 1 for more than 1 cycle and `c` is valid (not `x` or `z`)

test/conditionals_test.dart
```diff
@@ -189,6 +189,36 @@ class SequentialModule extends Module {
   }
 }
 
+class SingleIfModule extends Module {
```
mkorbel1, line 192: Neither of these two classes are used?  Did you mean to add a test that uses them?

mkorbel1 (changes requested):

test/conditionals_test.dart
mkorbel1: Actually, it looks like the tests were there but got removed somehow in a merge?

mkorbel1 (approved):
I think this looks good now!

Issue #194, opened 2022-11-08
Long combinational chains of logic can reach stack size limit during simulation


mkorbel1:
### Describe the bug

The long chain of `Logic.put` for a very long combinational logic chain can reach the stack size limit leading to a stack overflow.

### To Reproduce

This example hit the issue:
https://github.com/stevenmburns/rohd_sklansky/pull/6

### Expected behavior

Simulation gracefully handles very large designs

### Actual behavior

Stack Overflow

### Additional: pubspec.yaml

```yaml
Reproduced on v0.4.0, Dart v2.18.4
```

PR #199, opened 2022-11-15
`_Wire`s underneath `Logic`s

mkorbel1:
## Description & Motivation

Refactored `Logic` to hold its value on a `_Wire` which can be shared across many `Logic`s that are connected to each other.  This has two big benefits:
- Improved simulation performance by reducing the number of `glitch`es that need to propagate during combinational logic
- Decreases the stack requirements to help mitigate this issue: #194

## Related Issue(s)

Helps mitigate #194

## Testing

- Added new tests where corner cases could have been introduced around combinational logic propagation and edge listening.
- Ran the (limited) existing benchmark suite and observed slight performance improvement.  The existing pipeline benchmark is not sufficient probably to show the true benefit for larger designs, though.

## Backwards-compatibility

Sort of, `posedge` and `negedge` will now throw an `Exception` if the width is not 1.

## Documentation

No

### DIFF
--- a/lib/src/logic.dart
+++ b/lib/src/logic.dart
@@ -121,7 +121,15 @@ class _Wire {
       unawaited(_preTickSubscription.cancel());
       unawaited(_postTickSubscription.cancel());
       newChanged.listen(_changedController.add);
+      _changedBeingWatched = false;
     }
+  }
+
+  /// Tells this [_Wire] to adopt all the behavior of [other] so that
+  /// it can replace [other].
+  void _adopt(_Wire other) {
+    _glitchController.emitter.adopt(other._glitchController.emitter);
+    other._migrateChangedTriggers(changed);
   }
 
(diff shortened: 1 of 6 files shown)


eric-norige:
This refactor seems good to me.  Is there any measurement on sim performance speed improvement?  Do we need a benchmark suite of designs and stimulus to simulate to identify progress/regressions on simulation performance?

mkorbel1:
> Is there any measurement on sim performance speed improvement?

The one benchmark that's currently in the repo showed some modest relative improvement in simulation performance, but it's not a very meaningful measurement or dramatic difference.

> Do we need a benchmark suite of designs and stimulus to simulate to identify progress/regressions on simulation performance?

Yes!  I don't have them available and haven't had time to develop them, though.  If you (or anyone else) is interested in developing some or contributing existing things that could be reused as benchmarks it would be greatly valuable!

Issue #112, opened 2022-03-30
Automatically extract/update the version of ROHD that generates things


mkorbel1:
## Motivation
It would be nice if the generated outputs (e.g. SystemVerilog, waveform dumps) had a guaranteed accurate version number of ROHD that generated them.  A version number (from pubspec.yaml) would be good, but even better would be to also include a git hash, so it can be determined exactly what generated the code.

## Desired solution
Find a reliable way to figure out git hash, release information, version number, etc. and inject it into generated code.

## Alternatives considered  
Currently information is either omitted or manually updated (error prone).

quekyj:
Hi @mkorbel1, let me try to work on this.

PR #200, opened 2022-11-16
[Issue-112] Automatically extract rohd version from pubspec to add into generated sv

quekyj:
## Description & Motivation

It would be nice if the generated outputs (e.g. SystemVerilog, waveform dumps) had a guaranteed accurate version number of ROHD that generated them. A version number (from pubspec.yaml) would be good, but even better would be to also include a git hash, so it can be determined exactly what generated the code.

## Related Issue(s)

issue #112 

## Testing
Add test to expect the rohd version and git commit are available in the generated system verilog code generation.

## Backwards-compatibility

no.

## Documentation

Ya, added to each of the file created.


lib/src/utilities/configuration.dart
```diff
@@ -0,0 +1,38 @@
...
+/// A utility to extract the ROHD configuration from pubspec.yaml
+/// configuration document.
+abstract class Configuration {
+  /// A getter to return the current configuration of ROHD
+  static Map<String, dynamic> get getConfig {
+    const config = './pubspec.yaml';
+    final f = File(config);
+    final yamlText = f.readAsStringSync();
+    final yaml = loadYaml(yamlText) as Map;
+
```
quekyj, line 23: Change to String and return only version. How do you get the path to ROHD when you not working on rohd? They have their own ``pubspec.yaml`` instead? 

Can use dart tool to define the path? Need to query for correct location where the version is. 

Check out this:
``https://pub.dev/packages/package_config``

Easy but not elegant
Make another dart class with the update of the version. Add a CI check if the version in there not match with ``pubspec.yaml`` then fail. 

Or:
look from ``package.json`` but its a hacky method.


## Git Hash
- Add https://dart.dev/tools/pub/publishing 
- Make a script wrapper around publish instead run it manually

test/version_hash_dumper_test.dart
```diff
@@ -0,0 +1,42 @@
...
+    Combinational([
+      If(a, then: [c < a], orElse: [c < b])
+    ]);
+  }
+}
+
+void main() async {
+  test('should contains version number and git hash when sv is generated',
+      () async {
+    final version = Configuration.getConfig['version'] as String;
```
quekyj, line 31: Use 

```
Configuration.getGitHash
Configuration.getVersion
```

lib/src/utilities/configuration.dart
quekyj: https://suragch.medium.com/reading-configuration-files-and-settings-in-flutter-and-dart-cc14bbd93698

pubspec.yaml
```diff
@@ -1,6 +1,7 @@
 name: rohd
 description: The Rapid Open Hardware Development (ROHD) framework, a framework for describing and verifying hardware
 version: 0.4.0
+git_hash: 77a6b14ee70adbbe7968ca08d39eacd979e3e890
```
mkorbel1, line 4: Remove git_hash from the pubspec.yaml

mkorbel1:
If we create a (pre-)release script, a couple things it can do:
- Update some file with the current git hash
- Confirm that the version in pubspec.yaml matches the version in `Config`

quekyj:
Write a unit test to check for `pubspec.yaml` and the one in `config`.

quekyj:
Omit the hash part, too complicated to implement.

pubspec.yaml
```diff
@@ -14,6 +14,7 @@ dependencies:
   logging: ^1.0.1
   meta: ^1.3.0
   test: ^1.17.3
+  yaml: ^3.1.1
```
mkorbel1, line 17: This can be in `dev_dependencies` since it's only used for testing

lib/src/module.dart
```diff
@@ -498,6 +500,7 @@ abstract class Module {
 /**
  * Generated by ROHD - www.github.com/intel/rohd
  * Generation time: ${DateTime.now()}
+ * ROHD Version: ${Config.version}
```
mkorbel1, line 503: Nice, let's do the same thing in `wave_dumper.dart`.  Something like:
```
\$version
  ROHD v${Config.version}
\$end
```
In the `_writeHeader` function

mkorbel1 (changes requested):

pubspec.yaml
quekyj: sure!

lib/src/wave_dumper.dart
```diff
@@ -127,7 +128,7 @@ class WaveDumper {
   $dateString
 \$end
 \$version
-  ROHD
+  ROHD v${Config.version}
```
mkorbel1, line 131: We could add a test to confirm the version shows up in generated VCDs?

lib/src/utilities/config.dart
```diff
@@ -0,0 +1,15 @@
+/// Copyright (C) 2021-2022 Intel Corporation
```
mkorbel1, line 1: Make the copyright 2022 and creation date accurate (wasn't 2021 May 7)

test/config_test.dart
```diff
@@ -0,0 +1,50 @@
+/// Copyright (C) 2021 Intel Corporation
```
mkorbel1, line 1: Fix copyright and creation dates

mkorbel1 (changes requested):

lib/src/wave_dumper.dart
quekyj: Seems like dart do not support optional parameters with named parameter.

https://github.com/dart-lang/language/issues/1076

test/config_test.dart
```diff
@@ -27,6 +27,24 @@ class SimpleModule extends Module {
   }
 }
 
+const tempDumpDir = 'tmp_test';
```
mkorbel1, line 30: You should be able to `import` the `wave_dumper_test.dart` file to get access to these same functions and variables rather than copy-pasting.

mkorbel1 (changes requested):

mkorbel1 (approved):
Looks good!

PR #203, opened 2022-11-21
Fixes: 157

akshay-wankhede:
## Description & Motivation

At present, signExtend uses Mux which generates a SystemVerilog which is not as elegant as the replication operator. This fix removes the use of Mux and introduces another component that generates a replication operator in the output SystemVerilog. Also, when the input signal has a width of 1, the output does not need to be swizzled.

## Related Issue(s)

Fixes: 157

## Testing

In addition to the existing signExtend tests, added:
- test invalid bit replication
- test single bit replication (when the bit value is 1)
- test single bit replication (when the bit value is 0)

## Backwards-compatibility

No, this is not a breaking change.

## Documentation

No.


lib/src/modules/gates.dart
```diff
@@ -724,3 +724,57 @@ class IndexGate extends Module with InlineSystemVerilog {
     return '$target[$idx]';
   }
 }
+
+/// A Replication Operator [Module].
+///
+/// It takes two inputs (bit and width) and outputs a [Logic] representing
+/// the input bit repeated over the input width
+class ReplicationOp extends Module with InlineSystemVerilog {
+  late final String _inputName;
```
mkorbel1, line 733: It's nice to add some doc comments to these even though they are private

lib/src/modules/gates.dart
```diff
@@ -724,3 +724,57 @@ class IndexGate extends Module with InlineSystemVerilog {
...
+  Logic get _input => input(_inputName);
+
+  /// The output of this gate.
+  Logic get replicated => output(_outputName);
+
+  /// Constructs a ReplicationOp
+  ///
+  /// The bit [bit] will be repeated over the [_width] as an output.
+  /// [Module] is in-lined as SystemVerilog, it will use {width{bit}}
+  ReplicationOp(Logic bit, this._width) : super() {
```
mkorbel1, line 748: You don't need `super()`, it's implicit

You don't need to make `_width` `late`, since it's set in the constructor.

I know other code in here follows this same style, but you don't need to make `_inputName` and `_outputName` `late` either, you could set them as part of the constructor:
```dart
ReplicationOp(Logic bit, this._width) : _inputName = '...', _outputName = '...';
```

mkorbel1 (changes requested):
Looks great!!

Just a couple little nits, but other than that this is really nice!

mkorbel1 (approved):
Looks great!

PR #207, opened 2022-11-23
Fix bugs related to module naming and instantiation

### LINKED ISSUE #205: Module and signal names can collide in illegal ways in generated SystemVerilog
### Describe the bug

It is possible to name a signal and a module instance the same thing, but in generated SystemVerilog they cannot be the same.  They should be uniquified.

### To Reproduce

Create a signal name and module instance name that are the same and generate SystemVerilog

### Expected behavior

Names are properly uniquified so that generated SystemVerilog is valid

### Actual behavior

Invalid SystemVerilog is generated, leading to a compile time error

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

In a fix, it's important to consider every combination of:
- input name
- output name
- internal signal name
- module definition name
- module instance name

mkorbel1:
## Description & Motivation

- Fixed a bug where signal names and module instance names could conflict.
- Fixed a bug where in some cases modules might not be properly detected as sub-modules.

## Related Issue(s)

Fix #205 

## Testing

- Added a new test to cover the sub-module bug
- Added a new suite of tests covering combinations of names for input ports, output ports, internal signals, internal module instance names, internal module definition names, containing module definition names, and containing module instance names

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/module.dart
+++ b/lib/src/module.dart
@@ -407,6 +407,9 @@ abstract class Module {
     } else {
       if (!dontAddSignal && !isOutput(signal) && subModule == null) {
         _addInternalSignal(signal);
+        for (final dstConnection in signal.dstConnections) {
+          await _traceInputForModuleContents(dstConnection);
+        }
       }
       if (signal.srcConnection != null) {
         await _traceOutputForModuleContents(signal.srcConnection!);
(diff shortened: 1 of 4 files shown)


chaparalas (commented):
Does the dart static check flag an instance name = signal name ?

mkorbel1:
> Does the dart static check flag an instance name = signal name ?

Variable names in Dart are statically analyzed.  Signal name and instance names are uniquified automatically, where allowed, when generating SystemVerilog.  If names cannot be successfully uniquified (e.g. port names are guaranteed to be the maintained; it is possible to "reserve" an instance name of a module), then an error will be flagged during generation.

PR #208, opened 2022-11-23
Allow constants for modulo and shift operations on `Logic`

### LINKED ISSUE #170: Shift gates fail when shift value is a literal
The line below, will result in a runtime error when the value for b (dynamic variable) is passed as a literal.
https://github.com/intel/rohd/blob/c9ba07d3e978abb969e519252ca4dd6db0f40302/lib/src/modules/gates.dart#L285

Possible fix:
Instead of 'b.name' use 'bLogic.name'

mkorbel1:
## Description & Motivation

Most math operations supported a `dynamic` type to be operated on a `Logic`, but modulo and shift operations (`%`, `>>`, `<<`, `>>>`) only accepted other `Logic`s, which was inconvenient because it requires creating of a `Const` to use them.

This change allows `dynamic` on all these operations as well.

This also closes a testing hole around shift operations.

## Related Issue(s)

Close #170 via testing the fix that has been in there since #174

## Testing

Added tests for all new functionality, as well as missing tests around shift operations

## Backwards-compatibility

No, this loosens the API but is completely backwards compatible

## Documentation

No

### DIFF
--- a/test/math_test.dart
+++ b/test/math_test.dart
@@ -12,21 +12,8 @@ import 'package:rohd/src/utilities/simcompare.dart';
 import 'package:test/test.dart';
 
 class MathTestModule extends Module {
-  Logic get a => input('a');
-  Logic get b => input('b');
-
-  Logic get aPlusB => output('a_plus_b');
-  Logic get aMinusB => output('a_minus_b');
-  Logic get aTimesB => output('a_times_b');
-  Logic get aDividedByB => output('a_dividedby_b');
-  Logic get aModuloB => output('a_modulo_b');
-
-  Logic get aPlusConst => output('a_plus_const');
-  Logic get aMinusConst => output('a_minus_const');
-  Logic get aTimesConst => output('a_times_const');
-  Logic get aDividedByConst => output('a_dividedby_const');
-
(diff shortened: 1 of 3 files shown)


test/math_test.dart
```diff
@@ -40,22 +27,40 @@ class MathTestModule extends Module {
     final aTimesB = addOutput('a_times_b', width: a.width);
     final aDividedByB = addOutput('a_dividedby_b', width: a.width);
     final aModuloB = addOutput('a_modulo_b', width: a.width);
+    final aModuleConst = addOutput('a_modulo_const', width: a.width);
```
eric-norige, line 30: typo: `Module` should be `Module`, no?

eric-norige (commented):
I'm missing where in the code the `dynamic` value is turned into a `Const` Logic, but maybe that's somewhere common to all of these.

test/math_test.dart
mkorbel1: Yes, but you made the same typo :)

Thanks for the catch!

mkorbel1: Created #209

Issue #12, opened 2021-09-23
Create additional simpler constructors to decrease verbosity of simple control blocks


mkorbel1:
The default implementation for `If`, `Iff`, etc. take lists of `Conditional`s as inputs, which can be excessively verbose in cases where there is only one `Conditional`.  Adding a special constructor (e.g. something like `Iff.s`) which takes a single `Conditional` could make simple code look cleaner.

quekyj:
```
If.s(condition, then, [orElse])
```

quekyj:
counter_test.dart
```
Sequential.multi([
      SimpleClockGenerator(10).clk,
      reset
    ], [
      If(reset, then: [
        val < 0
      ], orElse: [
        If(en, then: [val < nextVal])
      ])
    ]);
```

chykon:
Shouldn't this issue be closed after this (https://github.com/intel/rohd/pull/210) merge?

mkorbel1:
Yes, you're right, thank you @chykon!

PR #210, opened 2022-11-25
[Issue-12] simpler constructor of simple control blocks

quekyj:
## Description & Motivation

The default implementation for If, Iff, etc. take lists of Conditionals as inputs, which can be excessively verbose in cases where there is only one Conditional. Adding a special constructor (e.g. something like Iff.s) which takes a single Conditional could make simple code look cleaner.

## Related Issue(s)

Issue #12 

## Testing

Added a new unit test that should return ``true`` if single conditional constructor is used.

## Backwards-compatibility

Nope.

## Documentation

Yes.

### DIFF
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -917,7 +917,7 @@ ${padding}end ''');
 /// Represents a block of code to be conditionally executed, like `if`/`else`.
 class If extends Conditional {
   /// [Conditional]s to be executed if [condition] is true.
-  late final List<Conditional> then;
+  final List<Conditional> then;
 
   /// [Conditional]s to be executed if [condition] is not true.
   final List<Conditional> orElse;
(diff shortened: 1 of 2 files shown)


lib/src/modules/conditional.dart
```diff
@@ -928,6 +928,13 @@ class If extends Conditional {
   /// If [condition] is 1, then [then] executes, otherwise [orElse] is executed.
   If(this.condition, {this.then = const [], this.orElse = const []});
 
+  /// If [condition] is 1, then [sig] is excutes.
+  ///
+  /// Use this constructor when you only have a single condition.
+  If.s(this.condition, Conditional sig, [this.orElse = const []]) {
```
mkorbel1, line 934: Let's change it from `sig` to `then` so that it's more clear that it's the same thing.

The `orElse` should also be just one `Conditional` rather than a list

Rather than making `then` `late`, you can do `If.s(...) : this(condition, then: [then], orElse: [orElse] )`

mkorbel1 (changes requested):

quekyj:
If [condition] is 1, then [then] will be executed.

```
ElseIf.s(this.condition, this.then);
```

quekyj:
If [condition] is 1, then [then] will be executed.

```
ElseIf.s(this. Condition, this.then);
```

mkorbel1 (approved):
This looks great, nice work!

PR #211, opened 2022-11-30
[Issue-133] simcompare calculate signaltowidth instead of manual input

### LINKED ISSUE #133: In `SimCompare` calculate the `signalToWidthMap` instead of manually inputting in `iverilogVector`
## Motivation
It's annoying and confusing to need to set the `signalToWidth` map for `SimCompare` after already setting port widths properly.

## Desired solution
Remove the `signalToWidthMap` from `SimCompare.iverilogVector`, and instead calculate it based on the port widths of the module.

## Additional details
This change might be good to do along with the CIRCT branch merge.

quekyj:
## Description & Motivation

It's annoying and confusing to need to set the signalToWidth map for SimCompare after already setting port widths properly.

## Related Issue(s)

Fix #133 

## Testing

No new unit test been created. 

## Backwards-compatibility

Yes. ``signalToWidth`` parameter is no longer supported while ``Module`` must be provided to the ``simcompare.iverilogVector()``.

## Documentation

Nope.


lib/src/utilities/simcompare.dart
```diff
@@ -144,19 +144,19 @@ abstract class SimCompare {
...
     bool dumpWaves = false,
-    Map<String, int> signalToWidthMap = const {},
     List<String> iverilogExtraArgs = const [],
     bool allowWarnings = false,
   }) {
     String signalDeclaration(String signalName) {
-      if (signalToWidthMap.containsKey(signalName)) {
-        final width = signalToWidthMap[signalName]!;
-        return '[${width - 1}:0] $signalName';
+      final signals = module.signals.firstWhere((e) => e.name == signalName);
```
mkorbel1, line 157: variable name should be `signal` not `signals` since it is singular

lib/src/utilities/simcompare.dart
```diff
@@ -144,19 +144,19 @@ abstract class SimCompare {
 
   /// Executes [vectors] against the Icarus Verilog simulator.
   static bool iverilogVector(
+    Module module,
     String generatedVerilog,
```
mkorbel1, line 148: I think we can remove `generatedVerilog` and call `generateSynth` within this function instead

lib/src/utilities/simcompare.dart
```diff
@@ -144,19 +144,19 @@ abstract class SimCompare {
...
     String topModule,
```
mkorbel1, line 149: We can make `topModule` optional and just use `module.definitionName` by default.  I believe there's 1 test where we need to override it, but most of the time we can take the default.

mkorbel1 (changes requested):
Looks great! Just a few changes

lib/src/utilities/simcompare.dart
quekyj: Seems like dart do not support optional parameters with named parameter.

https://github.com/dart-lang/language/issues/1076

eric-norige (approved):

lib/src/utilities/simcompare.dart
mkorbel1: Better than using `''` as the "empty" check, you should make `moduleName` nullable: `String? moduleName`.  Then you can just change it to the `definitionName` if it's null: `topModule = moduleName ?? module.definitionName`.

mkorbel1 (changes requested):
Just the one comment that I reopened, other than that looks good!

eric-norige:
> Just the one comment that I reopened, other than that looks good!

Good catch; I learned that github reviews can show you old versions of code, and I think I reviewed an old version.  I agree that null is better than '' for "no parameter passed".  This probably means that '' should be checked for and an error reported.

lib/src/utilities/simcompare.dart
quekyj: Yup, I was thinking the same things.

mkorbel1 (approved):
Looks good!

mkorbel1:
I think since this isn't part of the public API, let's count this as *not* backwards-compatibility-breaking, especially since migrating for those who are using it is just a matter of deleting some information.

Can you get this branch back up through merge conflicts and then we can merge it in?

mkorbel1:
Looks like you pulled in all the website changes as well into this branch?  Then we should either revert that so it's just simcompare in here or else merge the website first

quekyj:
Aww, I will remove the website. Looks like there were some mistakes there during my commit.

quekyj:
@mkorbel1 I think it should be good to go now. I reset to previous unpolluted commit and remerge.

test/conditionals_test.dart
```diff
@@ -477,144 +465,7 @@ void main() {
       Vector({'a': 0}, {'x': 1}),
     ];
     await SimCompare.checkFunctionalVector(mod, vectors);
-    final simResult = SimCompare.iverilogVector(
-        mod.generateSynth(), mod.runtimeType.toString(), vectors);
-    expect(simResult, equals(true));
-  });
-
-  test(
```
mkorbel1, line 485: Why delete these tests?

mkorbel1 (changes requested):

mkorbel1 (approved):
looks great!

PR #213, opened 2022-12-01
Fix bugs related to 64-bit int conversion, fix #212

### LINKED ISSUE #212: 64-bit width `LogicValue`s can be incorrect if their integer value is larger than the maximum positive `int` value in Dart
### Describe the bug

Doing things like grabbing a 64-bit slice of a `_BigLogicValue` or constructing a `LogicValue` using `ofBigInt` with width 64 with a value that could be interpreted as a negative signed `int` can cause clamping behavior in `BigInt.toInt()`.  This causes wrong operation results in a number of cases.

### To Reproduce

One example that reproduces the issue (this fails):
```dart
      final extraWide = LogicValue.ofBigInt(
        BigInt.parse('f' * 16 + 'f0' * 8, radix: 16),
        128,
      );
      final smaller = extraWide.getRange(0, 64);
      expect(smaller.toInt(), equals(0xf0f0f0f0f0f0f0f0));
```

### Expected behavior

Large 64-bit int values are treated as unsigned during conversions

### Actual behavior

Sometimes int values are treated as signed and clamped

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

This type of behavior could occur in a number of places, so it's important to consider everywhere it might occur.

mkorbel1:
## Description & Motivation

- Fix issues described in #212
- Add extension with utility on BigInt for unsigned conversion

## Related Issue(s)

Fix #212

## Testing

Added new tests covering changes

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/values/big_logic_value.dart
+++ b/lib/src/values/big_logic_value.dart
@@ -10,6 +10,25 @@
 
 part of values;
 
+/// Extends [BigInt] with utility functions that are useful for dealing with
+/// large bit vectors and conversion between types.
+extension BigLogicValueBigIntUtilities on BigInt {
+  /// Returns this [BigInt] as an [int].
+  ///
+  /// Always interprets the number as unsigned, and thus never clamps to fit.
+  int toIntUnsigned(int width) {
+    if (width > LogicValue._INT_BITS) {
+      throw Exception('Cannot convert to BigInt when width $width'
+          ' is greater than ${LogicValue._INT_BITS}');
+    } else if (width == LogicValue._INT_BITS) {
+      // With `0x` in front of a hex literal, it will be interpreted as unsigned.
+      return int.parse('0x${toRadixString(16)}');
+    } else {
+      return toInt();
+    }
+  }
+}
+
 /// A [LogicValue] whose number of bits is greater than the size of an [int].
 ///
 /// The implementation is similar to [_SmallLogicValue], except it uses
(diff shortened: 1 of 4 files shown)


lib/src/values/big_logic_value.dart
```diff
@@ -10,6 +10,25 @@
...
+  /// Returns this [BigInt] as an [int].
+  ///
+  /// Always interprets the number as unsigned, and thus never clamps to fit.
+  int toIntUnsigned(int width) {
+    if (width > LogicValue._INT_BITS) {
+      throw Exception('Cannot convert to BigInt when width $width'
+          ' is greater than ${LogicValue._INT_BITS}');
+    } else if (width == LogicValue._INT_BITS) {
+      // With `0x` in front of a hex literal, will be interpreted as unsigned.
+      return int.parse('0x${toRadixString(16)}');
```
eric-norige, line 25: There's got to be a better way to get at this; maybe 32 bits at a time can be safely extracted from the bigint through bit shifting?

mkorbel1: Sorry, looks like I just barely merged before your comment!

You're right, this appears to work well also, and is probably better?  What do you think?
```dart
  int toIntUnsigned(int width) {
    if (width > LogicValue._INT_BITS) {
      throw Exception('Cannot convert to BigInt when width $width'
          ' is greater than ${LogicValue._INT_BITS}');
    } else if (width == LogicValue._INT_BITS) {
      const maskWidth = 32;
      final mask = _BigLogicValue._maskOfWidth(maskWidth);
      return (this & mask).toInt() |
          (((this >> maskWidth) & mask).toInt() << maskWidth);
    } else {
      return toInt();
    }
  }
```

eric-norige: Thanks for noticing my late comment.  I see no correctness issues in this code.  Unfortunately it's a bit more complex than I expected, but I don't see a nice way to clean that up.  Despite version control history recording why these changes were made, I still suggest adding a comment explaining that this intentionally complex code is done this way to avoid signed conversion.  
Finally, I worry a bit about what happens when this all is run in a javascript environment where _INT_BITS isn't 64.  If _INT_BITS isn't even, even adjusting maskWidth won't be able to fix this code.  But maybe that's a problem for the future.

mkorbel1: Good point, I'll add a comment.

There's already a lot of problems with javascript that I'm noticing.  For example, in this PR I disabled the lint checks for large constants already (like `0xffffffffffffffff`) that won't work with javascript.  I think there's going to be a number of surprises when compiling to js, unfortunately.

PR #214, opened 2022-12-01
Replace awkward map+reduce chain with smart literal

eric-norige:
## Description & Motivation

Saw some awkward code, wanted to cleanup

## Related Issue(s)

None

## Testing

Existing tests should valid correctness of this code.

## Backwards-compatibility

Completely backwards compatible

## Documentation

No need.


mkorbel1 (approved):
Looks good! Thank you for the contribution!

PR #215, opened 2022-12-01
Optimize LogicValue operations to avoid Strings where possible

### LINKED ISSUE #202: Make a shortcut to create a 0-width `LogicValue`
### Motivation

There are multiple ways to create a `LogicValue` with no width, but none of them feel that clean.

Something like `LogicValue.empty` which always returns a 0-width `LogicValue`, ideally the most efficiently implemented way, would be convenient.

### Desired solution

Add a new constant to `LogicValue` to create a zero-width value.

### Alternatives considered

Just use `LogicValue.of([])`, or something similar.

### Additional details

_No response_

mkorbel1:
## Description & Motivation

String manipulation is not ideal for performance but was used in a number of places for `LogicValue` operations.  This PR removes some of those places where possible.

See conversation that initiated some of these changes here: https://github.com/intel/rohd/pull/213#discussion_r1037443152

This PR specifically fixes:
- Conversion to unsigned `int` from `BigInt` to not use `String`
- `LogicValue.of` to not use `String` (which also optimizes all of the `swizzle` operations on `LogicValue`)

## Related Issue(s)

Fixed #202 (since it was convenient for implementation)

Improved upon fixes in #213

## Testing

Added new tests to cover potential corner cases

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/values/logic_value.dart
+++ b/lib/src/values/logic_value.dart
@@ -137,17 +137,21 @@ abstract class LogicValue {
     // shift small chunks in together before shifting BigInt's, since
     // shifting BigInt's is expensive
     for (final lv in it) {
-      if (lv.width + smallBuffer.width <= _INT_BITS) {
+      final lvPlusSmall = lv.width + smallBuffer.width;
+      if (lvPlusSmall <= _INT_BITS) {
         smallBuffer = lv._concatenate(smallBuffer);
       } else {
-        final upperBound = (_INT_BITS - smallBuffer.width) +
-            _INT_BITS * max(lv.width ~/ _INT_BITS - 1, 0) as int;
+        // only put 64-bit chunks onto `fullResult`, rest onto `smallBuffer`
+        final upperBound =
+            _INT_BITS * (lvPlusSmall ~/ _INT_BITS) - smallBuffer.width;
         fullResult = lv
             .getRange(0, upperBound)
             ._concatenate(smallBuffer)
             ._concatenate(fullResult);
         smallBuffer = lv.getRange(upperBound, lv.width);
       }
+      assert(smallBuffer.width <= _INT_BITS,
+          'Keep smallBuffer small to meet invariants and efficiency');
     }
 
     // grab what's left
--- a/lib/src/values/logic_value.dart
+++ b/lib/src/values/logic_value.dart
@@ -137,10 +137,16 @@ abstract class LogicValue {
     // shift small chunks in together before shifting BigInt's, since
     // shifting BigInt's is expensive
     for (final lv in it) {
-      smallBuffer = lv._concatenate(smallBuffer);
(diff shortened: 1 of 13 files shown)


lib/src/values/logic_value.dart
```diff
@@ -115,16 +118,56 @@ abstract class LogicValue {
...
+  /// var it = [LogicValue.zero, LogicValue.x, LogicValue.ofString('01xz')];
   /// var lv = LogicValue.of(it);
-  /// print(lv); // This prints `3b'1x0`
+  /// print(lv); // This prints `6'b01xzx0`
   /// ```
-  static LogicValue of(Iterable<LogicValue> it) => LogicValue.ofString(
-      it.map((e) => e.toString(includeWidth: false)).toList().reversed.join());
+  ///
+  static LogicValue of(Iterable<LogicValue> it) => it.fold(LogicValue.empty,
+      (previousValue, element) => element._concatenate(previousValue));
```
eric-norige, line 134: I like the idea of this much better than concatenating strings and parsing, but I'm sure we can implement this better.  Doing sequential concatenation will create n-2 temporary values and will potentially take O(n^2) time.  Consider concatenating 1000 bits; the first bit will be shifted left by 1 999 times.  
Instead, I recommend a 2-pass solution.  In the first pass, you're going to sum the widths and check if all inputs are FIlledLogicValues.  Then you can create the final output kind of LogicValue and shift each of the _values into the right place in that output.

lib/src/values/small_logic_value.dart
```diff
@@ -180,4 +180,18 @@ class _SmallLogicValue extends LogicValue {
...
           _invalid >> shamt,
           width);
+
+  @override
+  BigInt get _bigIntInvalid =>
+      BigInt.from(_invalid) & _BigLogicValue._maskOfWidth(width);
+
+  @override
+  BigInt get _bigIntValue =>
+      BigInt.from(_value) & _BigLogicValue._maskOfWidth(width);
```
eric-norige, line 190: Why is this masking needed?  No masking is needed when returning a plain `_intValue`, so `BigInt.from(_value)` should be already correct, no?

eric-norige (changes requested):

lib/src/values/small_logic_value.dart
mkorbel1: I originally didn't have the mask, but it creates a bug when the `_value` can be interpreted as a negative number since when it is bitwise-`|`ed against another bus it sign extends the 1s.

lib/src/values/logic_value.dart
mkorbel1: Thank you for the help and suggestions (here and outside of this thread)!  The implementation I just pushed based on your feedback is ~3x faster than the one this comment referred to originally.

eric-norige (approved):

PR #216, opened 2022-12-02
added the inline function for isValid and isFloating

priya-t12:
## Description & Motivation

Added two inline functions to Deprecate hasValidValue and isFloating in logic.dart file. 
These changes are done so that users can just access value themselves to check

## Backwards-compatibility

## Documentation


lib/src/logic.dart
```diff
@@ -100,9 +100,15 @@ class Logic {
   BigInt get valueBigInt => value.toBigInt();
 
   /// Returns `true` iff the value of this signal is valid (no `x` or `z`).
+  /// 
+  /// Throws an exception 
+  @Deprecated('Use currentValue.isValid instead.')
```
mkorbel1, line 105: Should say "Use value.isValid instead." since `_currentValue` is a private variable (starting with an underscore `_`)

mkorbel1 (changes requested):
Thank you for the contribution!  Just one adjustment.

There were some other changes in this file merged into main, so you'll have to resolve some merge conflicts by pulling those as well.

lib/src/logic.dart
mkorbel1: This still needs to be updated to "value.isValid" in the `Deprecated` note

lib/src/logic.dart
```diff
@@ -335,9 +335,13 @@ class Logic {
   BigInt get valueBigInt => value.toBigInt();
 
   /// Returns `true` iff the value of this signal is valid (no `x` or `z`).
+  @Deprecated('Use currentValue.isValid instead.')
   bool hasValidValue() => value.isValid;
 
   /// Returns `true` iff *all* bits of the current value are floating (`z`).
+  ///
+  /// Throws an exception
```
mkorbel1, line 343: Why did we add "Throws an exception" here? we should remove that

mkorbel1 (approved):
Looks good!

Issue #204, opened 2022-11-22
Generated SystemVerilog includes index access on 1-bit signals


mkorbel1:
### Describe the bug

If a signal is defined as 1 bit in SystemVerilog:
```verilog
logic x;
```
Then you should not access the 0'th bit of it like this:
```verilog
x[0]
```
However, ROHD can generate this.

### To Reproduce

Access the 0'th bit of a `Logic` and generate SystemVerilog for it.

### Expected behavior

When a signal is 1 bit, treat a 0-index into it as a nop and generate accesses as just the raw signal.  It's probably even better to just short-circuit this and return the signal itself immediately to save the extra work of a nop module.

### Actual behavior

Generated verilog has `x[0]`-like behavior for 1-bit signals

### Additional: Context

This bug likely applies to both `slice`/`BusSubset` via constant indexing and `IndexGate`.

akshay-wankhede:
I can work on this one

PR #220, opened 2022-12-07
Fixes: 204

akshay-wankhede:
## Description & Motivation
- Fixed: slice on a 1 bit signal should generate $x into SystemVerilog, instead of a $x[0], for single bit signals
- Fixed: test/config_test.dart: ROHD version in wavedumper test fails on Windows, since delete temporary directory is called before the simulator run returns.
- Fixed: IndexGate and BusSubset to generate $x into SystemVerilog, instead of $x[0] for single bit signals
- Updated: doc comments

## Related Issue(s)

Fixes: [204](https://github.com/intel/rohd/issues/204)
## Testing

- Added: additional tests on 1 bit signals using slice, IndexGate and BusSubset

## Backwards-compatibility

No.

## Documentation

Yes.


lib/src/modules/gates.dart
```diff
@@ -705,6 +707,11 @@ class IndexGate extends Module with InlineSystemVerilog {
 
   /// Executes the functional behavior of this gate.
   void _execute() {
+    if (_original.width == 1) {
```
mkorbel1, line 710: I don't think we need this if statement, since we probably want X-prop if the `_index` is invalid, and also want to flag an error if the `_index` is out of bounds, which code below should already cover

lib/src/modules/gates.dart
```diff
@@ -705,6 +707,11 @@ class IndexGate extends Module with InlineSystemVerilog {
...
+      selection.put(_original.value);
+      return;
+    }
+
     if (_index.hasValidValue()) {
       final indexVal = _index.value.toInt();
       selection.put(_original.value.getRange(indexVal, indexVal + 1));
```
mkorbel1, line 717: I think we need to drive `x` if `indexVal` is out of range, I guess a miss from before, but one that your new tests will cover (with above modification)

test/gate_test.dart
```diff
@@ -430,31 +430,72 @@ void main() {
...
         Vector({'original': LogicValue.x, 'index': 0},
-            {'index_output': LogicValue.x})
+            {'index_output': LogicValue.x}),
+        Vector({'original': LogicValue.x, 'index': 1},
+            {'index_output': LogicValue.x}),
+        Vector({'original': LogicValue.x, 'index': 2},
+            {'index_output': LogicValue.x}),
+        Vector({'original': LogicValue.one, 'index': 0},
+            {'index_output': LogicValue.one}),
+        Vector({'original': LogicValue.one, 'index': 1},
```
mkorbel1, line 463: these 1-bit accesses with an out of range index should be `x` after other changes

test/gate_test.dart
```diff
@@ -430,31 +430,72 @@ void main() {
...
+      final testLogicOne = Logic()..put(LogicValue.one);
+      final testLogicZero = Logic()..put(LogicValue.zero);
+      final testLogicInvalid = Logic()..put(LogicValue.x);
+
       expect(testLogic[0].value.toInt(), 0);
       expect(testLogic[2].value.toInt(), 1);
       expect(() => testLogic[10], throwsException);
+
+      expect(testLogicOne[0].value.toInt(), 1);
+      expect(testLogicOne[1].value.toInt(), 1);
```
mkorbel1, line 492: 1-bit accesses with out of range index should be exceptions, which should happen with other changes mentioned

mkorbel1 (changes requested):

lib/src/logic.dart
```diff
@@ -631,6 +631,8 @@ class Logic {
   /// value will be reversed relative to the original signal.
   /// Negative/Positive index values are allowed. (The negative indexing starts from where the array ends)
   ///
+  /// When, the signal width is 1, [endIndex] and [startIndex] will be ignored.
```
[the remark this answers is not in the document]
akshay-wankhede: Avoided BusSubset creation, in case the index is invalid (>=1), it throws an exception.

lib/src/logic.dart
```diff
@@ -614,7 +614,10 @@ class Logic {
   /// nextVal <= val[-9]; // Error!: allowed values [-8, 7]
   /// nextVal <= val[8]; // Error!: allowed values [-8, 7]
   /// ```
-  ///
+  /// Note: When, indexed by a Logic value, out-of-bounds will always return an
+  /// invalid (LogicValue.x) value. This behavior is differs in simulation as
+  /// compared to the generated SystemVerilog. In the generated SystemVerilog,
+  /// [index] will be ignored, and the logic $x is returned as-is.
```
mkorbel1, line 620: What's this `$x` in here? typo?

lib/src/modules/gates.dart
```diff
@@ -679,6 +679,8 @@ class IndexGate extends Module with InlineSystemVerilog {
   /// The bit [index] will be indexed as an output.
```
mkorbel1, line 679: Not from this PR, but mentioning "bit" is not quite correct, maybe call it a signal here since it might be more than 1 bit wide.  Also, it's probably better called a "two-input index gate" without the word "bit" in the line above.

mkorbel1 (changes requested):
A couple little doc changes, but other than that looks great!

lib/src/logic.dart
```diff
@@ -648,6 +652,10 @@ class Logic {
         (startIndex < 0) ? width + startIndex : startIndex;
     final modifiedEndIndex = (endIndex < 0) ? width + endIndex : endIndex;
 
+    if (width == 1 && endIndex == 0 && modifiedEndIndex == modifiedStartIndex) {
```
mkorbel1, line 655: Shouldn't this be `modifiedEndIndex == 0`?  What if `endIndex` is -1 on a 1-bit signal?

mkorbel1 (changes requested):

lib/src/logic.dart
akshay-wankhede: in that case BusSubset will throw an exception during initialization

mkorbel1 (approved):
Awesome work!

PR #227, opened 2022-12-16
[Issue-104] checks to prevent same conditional used multiple times

### LINKED ISSUE #104: Add checks to prevent the same `Conditional` from being used multiple times
## Motivation
If someone uses the same `Conditional` object multiple times in the same `_Always` or in different `_Always` blocks, it can cause unexpected behavior and simulation/synthesis mismatches.

## Desired solution
There should be an explicit check that `Conditional`s are not reused.  Perhaps a build-time check where each `Conditional` is assigned a parent `_Always`, and the `_Always` confirms there are no duplicates within itself?

quekyj:
## Description & Motivation

If someone uses the same Conditional object multiple times in the same _Always or in different _Always blocks, it can cause unexpected behavior and simulation/synthesis mismatches.

## Related Issue(s)

Fix #104 

## Testing

Make sure an exception is thrown when conditionals are reuse.

## Backwards-compatibility

no

## Documentation

no


test/conditionals_test.dart
```diff
@@ -373,6 +392,97 @@ void main() {
...
     });
+
+    test('should return exception if a conditional is used multiple times.',
+        () async {
+      expect(
+          () => MultipleConditionalModule(Logic(), Logic()), throwsException);
+    });
+  });
+
+  test(
```
mkorbel1, line 403: Why are all these tests added with this PR?  Were they missing last PR?

mkorbel1 (commented):
Looks good! Just one question

test/conditionals_test.dart
quekyj: It is because of the merge conflict. Some spacing added. So, it treats it like a newly added line since I add the conditionals test on the middle.

mkorbel1: It looks like the conflict resolution actually added duplicates to all these tests?  You should remove these duplicates

mkorbel1 (changes requested):

test/conditionals_test.dart
quekyj: Removed duplicated code

mkorbel1 (approved):
Looks good!

PR #229, opened 2022-12-19
[Issue-225] Add Else.s() constructor to for single then execution

### LINKED ISSUE #225: Add `Else.s` simple constructor that accepts only 1 `Conditional`
### Motivation

https://github.com/intel/rohd/pull/210 implements `If.s`, `Iff.s`, and `ElseIf.s`, but not `Else.s`.

### Desired solution

It would be nice to also add `Else.s` for completeness.

### Alternatives considered

_No response_

### Additional details

_No response_

quekyj:
## Description & Motivation

https://github.com/intel/rohd/pull/210 implements If.s, Iff.s, and ElseIf.s, but not Else.s.

## Related Issue(s)

Fix #225 

## Testing

Add new unit test for single conditionals else then.

## Backwards-compatibility

No.

## Documentation

No.

### DIFF
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -831,6 +831,12 @@ class Else extends Iff {
   /// If none of the proceding [Iff] or [ElseIf] are executed, then
   /// [then] will be executed.
   Else(List<Conditional> then) : super(Const(1), then);
+
+  /// If none of the proceding [Iff] or [ElseIf] are executed, then
+  /// [then] will be executed.
+  ///
+  /// Use this constructor when you only have a single [then] condition.
+  Else.s(Conditional then) : super(Const(1), [then]);
 }
 
 /// Represents a chain of blocks of code to be conditionally executed, like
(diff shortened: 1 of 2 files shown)


lib/src/modules/conditional.dart
```diff
@@ -831,6 +831,12 @@ class Else extends Iff {
   /// If none of the proceding [Iff] or [ElseIf] are executed, then
   /// [then] will be executed.
   Else(List<Conditional> then) : super(Const(1), then);
+
+  /// If none of the proceding [Iff] or [ElseIf] are executed, then
+  /// [then] will be executed.
+  ///
+  /// Use this constructor when you only have a single [then] condition.
+  Else.s(Conditional then) : super(Const(1), [then]);
```
mkorbel1, line 839: I think you can just do `... : this([then]);` instead of putting the `Const(1)` again

mkorbel1 (changes requested):
Looks good, one minor change

mkorbel1 (approved):
Looks good!

PR #231, opened 2022-12-20
[Issue-228] Make endIndex in getRange an optional positional argument, default to width

### LINKED ISSUE #228: Make `endIndex` in `getRange` an optional positional argument, default to width
### Motivation

It's relatively common to want to grab a bus or value from a certain bit all the way to the end, but right now that requires explicitly specifying the width as the `end` argument of `getRange`.  It would be nice if supplying only the `start` would default to grabbing all the way to the end.

### Desired solution

Modify `getRange` in both `Logic` and `LogicValue` so that `endIndex` is an optional positional argument:
```dart
... getRange(int startIndex, [int? endIndex]) {
```
If `endIndex` is null, then default the value of it to `width`.

### Alternatives considered

_No response_

### Additional details

_No response_

quekyj:
## Description & Motivation

It's relatively common to want to grab a bus or value from a certain bit all the way to the end, but right now that requires explicitly specifying the width as the end argument of getRange. It would be nice if supplying only the start would default to grabbing all the way to the end.

## Related Issue(s)

Fix #228 

## Backwards-compatibility

No

## Documentation

### DIFF
--- a/lib/src/logic.dart
+++ b/lib/src/logic.dart
@@ -678,7 +678,8 @@ class Logic {
   /// nextVal <= val.getRange(0, 6); // = val.slice(0, -2) & output: 0b001110, where the output.width=6
   /// ```
   ///
-  Logic getRange(int startIndex, int endIndex) {
+  Logic getRange(int startIndex, [int? endIndex]) {
+    endIndex ??= width;
     if (endIndex == startIndex) {
       return Const(0, width: 0);
     }
--- a/lib/src/logic.dart
+++ b/lib/src/logic.dart
@@ -668,6 +668,9 @@ class Logic {
   /// [endIndex] are equal, then a zero-width signal is returned.
   /// Negative/Positive index values are allowed. (The negative indexing starts from where the array ends)
   ///
+  /// If [endIndex] is not provided, [width] of the [Logic] will
+  /// be used as the default values which assign it to the last index.
+  ///
   /// ```dart
   /// Logic nextVal = addOutput('nextVal', width: width);
   /// // Example: val = 0xce, val.width = 8, bin(0xce) = "0b11001110"
@@ -676,6 +679,9 @@ class Logic {
   ///
   /// // Positive getRange
   /// nextVal <= val.getRange(0, 6); // = val.slice(0, -2) & output: 0b001110, where the output.width=6
+  ///
+  /// // Get range from startIndex
+  /// nextVal <= val.getRange(-3); // the endIndex will be auto assign to val.width
   /// ```
   ///
   Logic getRange(int startIndex, [int? endIndex]) {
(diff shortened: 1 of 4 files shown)


lib/src/logic.dart
```diff
@@ -678,7 +678,8 @@ class Logic {
   /// nextVal <= val.getRange(0, 6); // = val.slice(0, -2) & output: 0b001110, where the output.width=6
   /// ```
```
mkorbel1, line 679: Make sure to update the doc comments about behavior if `endIndex` is not provided

lib/src/values/logic_value.dart
```diff
@@ -476,7 +476,8 @@ abstract class LogicValue {
   /// LogicValue.ofString('0101').getRange(0, 7);   // Error - bad inputs end > length-1
   /// ```
   ///
-  LogicValue getRange(int startIndex, int endIndex) {
```
mkorbel1, line 479: Make sure to update the doc comments about behavior if `endIndex` is not provided

test/bus_test.dart
```diff
@@ -472,6 +484,10 @@ void main() {
         Vector({'a': 0}, {'a_neg_range3': 0}),
         Vector({'a': 0x80}, {'a_neg_range3': 1}),
         Vector({'a': bin('10000000')}, {'a_neg_range3': bin('1')}),
+        // Test set 4
+        Vector({'a': 0}, {'a_neg_range4': 0}),
+        Vector({'a': 0xaf}, {'a_neg_range4': 5}),
+        Vector({'a': bin('11000101')}, {'a_neg_range4': bin('110')}),
```
mkorbel1, line 490: Add tests for `LogicValue` as well, these tests only cover `Logic`

mkorbel1 (changes requested):

lib/src/values/logic_value.dart
```diff
@@ -467,16 +467,22 @@ abstract class LogicValue {
   /// and [endIndex] are equal, then a zero-width value is returned.
   /// Negative/Positive index values are allowed. (The negative indexing starts from the end=[width]-1)
   ///
+  /// If [endIndex] is not provided, [width] of the [Logic] will
```
mkorbel1, line 470: In this case it's `LogicValue` not `Logic`

lib/src/values/logic_value.dart
```diff
@@ -467,16 +467,22 @@ abstract class LogicValue {
...
+  /// be used as the default values which assign it to the last index.
+  ///
   /// ```dart [TODO]
   /// LogicValue.ofString('0101').getRange(0, 2);   // == LogicValue.ofString('01')
   /// LogicValue.ofString('0101').getRange(1, -2);  // == LogicValue.zero
   /// LogicValue.ofString('0101').getRange(-3, 4);  // == LogicValue.ofString('010')
+  /// LogicValue.ofString('0101').getRange(1); // == LogicValue.ofString('010')
```
mkorbel1, line 477: Can you align the comment with the ones above?

mkorbel1 (changes requested):
A couple minor comment issues, other than that looks great!

mkorbel1 (approved):
Looks great!

PR #240, opened 2022-12-30
Fix combinational sensitivity excessive pessimism, fix #233

### LINKED ISSUE #233: `Combinational` sensitivity detection is excessively pessimistic
### Describe the bug

The current implementation that searches for expanded sensitivity lists in `Combinational` will not look "inside" other `Module`s along the path from its outputs to inputs.  If a `Sequential` is inside of another `Module` along the path, it will skip over it assuming that any input may have a combinational path to any output.  This is excessively pessimistic, and can lead to detection of non-existent combinational loops in `Logic` reentrance checking.

### To Reproduce

Connections like shown below will reproduce this issue:
```
  ______________________
  | Module             |
--|--- [Sequential] ---|--
| |____________________| |
|                        |
-----[Combinational]------
```

If the combinational in the diagram is also responsible for driving some other input to the module, it can cause a false combinational loop detection.

### Expected behavior

No invalid detection of combinational loops.

### Actual behavior

X generation due to reentrance detection in `Logic.put`

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

_No response_

mkorbel1:
## Description & Motivation

Issue #233 finds that searches for `Combinational` sensitivity lists, it is excessively pessimistic in that sometimes there aren't really any combinational paths, but it assumes that there may be.  This can cause an issue where a false combinational loop is detected, causing incorrect X-generation.

This PR fixes the bug by calculating more detailed combinational paths for `Module`s at `build` time.  This information can be leveraged by `Combinational` to understand more accurately where combinational paths may exist.  Results are cached to avoid repeated calculations when multiple `Combinational`s search through similar paths of potential sensitivities.  All `Module`s with custom functionality must explicitly define any combinational paths from all inputs to applicable outputs or else `Combinational` will be unable to properly determine an accurate sensitivity list.  A new `FullyCombinational` mixin is included in this PR to automatically add all input/output combinations as potential combinational paths, since that's a common case for many small custom gates.

Additionally,
- Added `RohdException` base type for all ROHD exceptions, making it easier to handle ROHD-related exceptions
- Added new `ModuleNotBuiltException`, used in multiple places in `Module`

## Related Issue(s)

Fix #233 

## Testing

- Added a new test that had reproduced the original issue without this fix.
- Existing tests cover scenarios where non-trivial sensitivities exist.

## Backwards-compatibility

Other custom functionality modules defined outside of the ROHD framework which had previously benefited from assumption of the behavior of `FullyCombinational` will now be treated as the complete opposite (no combinational paths).  This could mean `Combinational`s would now be missing sensitivities that previously existed.  If those sensitivities were true, then existing simulations in ROHD might now have incorrect behavior.  This is likely rare, and since this is really a bug fix it doesn't quite feel like it's breaking backwards-compatibility.

## Documentation

Doc comments updated/included

### DIFF
--- a/lib/src/module.dart
+++ b/lib/src/module.dart
@@ -14,6 +14,8 @@ import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 
 import 'package:rohd/rohd.dart';
+import 'package:rohd/src/collections/traverseable_collection.dart';
+import 'package:rohd/src/exceptions/module/module_exceptions.dart';
 import 'package:rohd/src/exceptions/name/name_exceptions.dart';
 import 'package:rohd/src/utilities/config.dart';
 import 'package:rohd/src/utilities/sanitizer.dart';
@@ -121,7 +123,8 @@ abstract class Module {
   /// unique name within its scope.
   String get uniqueInstanceName => hasBuilt || reserveName
       ? _uniqueInstanceName
-      : throw Exception('Module must be built to access uniquified name.'
+      : throw ModuleNotBuiltException(
+          'Module must be built to access uniquified name.'
           '  Call build() before accessing this.');
   String _uniqueInstanceName;
 
@@ -188,7 +191,8 @@ abstract class Module {
   /// Only returns valid information after [build].
   Iterable<Module> hierarchy() {
     if (!hasBuilt) {
-      throw Exception('Module must be built before accessing hierarchy.'
+      throw ModuleNotBuiltException(
+          'Module must be built before accessing hierarchy.'
           '  Call build() before executing this.');
     }
     Module? pModule = this;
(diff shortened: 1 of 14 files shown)


lib/src/module.dart
```diff
@@ -252,6 +259,118 @@ abstract class Module {
...
+  ///
+  /// The default behavior of this function is to search through from all
+  /// inputs to all potential outputs.  If a [Module] implements custom behavior
+  /// internally (e.g. a custom gate or a cosimulated module), then it makes
+  /// sense to override this function to give an accurate picture.  If the
+  /// default behavior doesn't work (because no visible connectivity exists
+  /// inside the [Module]), then the return value will end up with all empty
+  /// [List]s in the values of the [Map].
+  ///
+  /// The result of this function is stored at [build] time.  The result is
```
mkorbel1, line 332: Update doc comment with changes about build-time

lib/src/module.dart
```diff
@@ -252,6 +259,118 @@ abstract class Module {
...
+  /// primarily used for calculating valid and complete sensitivity lists
+  /// for [Combinational] execution.
+  @protected
+  Map<Logic, List<Logic>> getCombinationalPaths() {
```
mkorbel1, line 336: Reorder functions so get and getreverse are near each other

Issue #224, opened 2022-12-15
Expose `ReplicationOp` as a function on `Logic`


mkorbel1:
### Motivation

The replication operator is useful (introduced in https://github.com/intel/rohd/pull/203), but currently only can be instantiated by constructing a new `ReplicationOp`.  It would be nice to expose it as a function on `Logic` itself.

### Desired solution

Add some function on `Logic`, like `Logic.replicate(int multiplier)` or something like that.

### Additional details

Probably good to consider multi-bit handling at the same time, referenced here: https://github.com/intel/rohd/issues/223

akshay-wankhede:
Please assign this issue to me

Issue #223, opened 2022-12-15
`ReplicationOp` should support multi-bit replication


mkorbel1:
### Summary

PR https://github.com/intel/rohd/pull/203 introduced `ReplicationOp`, but only 1-bit `Logic`s are supported, it does support multi-bit `Logic`s.  

### To Reproduce

Use `ReplicationOp` with a multi-bit value for the `bit` input.

### Expected behavior

It's probably nice to support multi-bit, but requires a bit more logic to handle that.

akshay-wankhede:
Please assign it to me

PR #247, opened 2023-01-18
Fixes: 223 and Fixes: 224

akshay-wankhede:
## Description & Motivation
Fixes: 223
PR https://github.com/intel/rohd/pull/203 introduced ReplicationOp, but only 1-bit Logics are supported, it does support multi-bit Logics.

To Reproduce
Use ReplicationOp with a multi-bit value for the bit input.

Expected behavior
It's probably nice to support multi-bit, but requires a bit more logic to handle that.

Fixes: 224
Add some function on Logic, like Logic.replicate(int multiplier) or something like that.

## Related Issue(s)

[223](https://github.com/intel/rohd/issues/223)
[224](https://github.com/intel/rohd/issues/224)

## Testing

- Added tests ([tests/replication_test.dart])
  - Test invalid multipliers (0, -1)
  - Test that multiply by 1 returns the same signal
  - Test other multipliers
 
## Backwards-compatibility

No.
## Documentation

Added/Modified the required doc comments.


lib/src/logic.dart
```diff
@@ -842,4 +842,10 @@ class Logic {
       getRange(0, startIndex),
     ].swizzle();
   }
+
+  /// Returns a replicated signal using [ReplicationOp] with new
+  /// width = this.width * [multiplier]
+  /// The input width cannot be negative or 0, an exception will be thrown in
```
mkorbel1, line 848: I think you mean "the [multiplier] cannot be negative or 0..."

lib/src/modules/gates.dart
```diff
@@ -772,13 +772,20 @@ class ReplicationOp extends Module
...
+  /// Input [_multiplier] cannot be negative or zero, an exception will be
+  /// thrown, otherwise.
   /// [Module] is in-lined as SystemVerilog, it will use {width{bit}}
-  ReplicationOp(Logic bit, this._width)
-      : _inputName = 'input_${bit.name}',
-        _outputName = Module.unpreferredName('output_${bit.name}') {
-    addInput(_inputName, bit);
-    addOutput(_outputName, width: _width);
+  ReplicationOp(Logic original, this._multiplier)
+      : _inputName = 'input_${original.name}',
```
mkorbel1, line 781: I'm unsure if SV allows chaining of replication operators.  Did you run into issues making the input an `unpreferredName` when simulating the generated SV?  If we don't have a test that stresses this a bit (chaining replication operators with other in-line SV operations), we should.  It's better to make it "unpreferred" if we can.

lib/src/modules/gates.dart
```diff
@@ -772,13 +772,20 @@ class ReplicationOp extends Module
...
+        _outputName = Module.unpreferredName('output_${original.name}') {
+    if (_multiplier < 1) {
+      throw Exception('Multiplier({_multiplier}) is not >= 1');
```
mkorbel1, line 784: We're trying to make exceptions more specific (see the exceptions/ directory).  Try to avoid using the base `Exception` and instead either use an existing one or make a new one.

lib/src/modules/gates.dart
```diff
@@ -792,7 +799,7 @@ class ReplicationOp extends Module
 
   /// Executes the functional behavior of this gate.
   void _execute() {
-    replicated.put(_input.value, fill: true);
+    replicated.put(List.filled(_multiplier, _input.value).swizzle());
```
mkorbel1, line 802: What if we add a `replicate` function to `LogicValue` so that we can just call it here, plus get the benefit of doing replication on `LogicValue`s too?

test/replication_test.dart
```diff
@@ -0,0 +1,71 @@
+/// Copyright (C) 2022 Intel Corporation
```
mkorbel1, line 1: Update copyright to 2023

lib/src/logic.dart
```diff
@@ -842,4 +842,10 @@ class Logic {
       getRange(0, startIndex),
     ].swizzle();
   }
+
```
mkorbel1, line 845: Update copyright at top to 2021-2023

lib/src/modules/gates.dart
```diff
@@ -762,7 +762,7 @@ class ReplicationOp extends Module
   // output component name
   final String _outputName;
   // Width of the output signal
-  final int _width;
+  final int _multiplier;
```
mkorbel1, line 765: Update copyright at top to 2021-2023

test/replication_test.dart
```diff
@@ -0,0 +1,71 @@
...
+/// SPDX-License-Identifier: BSD-3-Clause
+///
+/// extend_test.dart
```
mkorbel1, line 4: Update name of file and description of contents

test/replication_test.dart
```diff
@@ -0,0 +1,71 @@
...
+    final b = addOutput('b', width: newWidth <= 0 ? a.width : newWidth);
+
+    b <= a.replicate(multiplier);
+  }
+}
+
+void main() {
+  group('Logic', () {
+    tearDown(Simulator.reset);
+    group('extend', () {
```
mkorbel1, line 27: Why is this group called "extend"?

test/replication_test.dart
```diff
@@ -0,0 +1,71 @@
...
+      Future<void> replicateVectors(List<Vector> vectors, int multiplier,
+          {int originalWidth = 8}) async {
+        final newWidth = originalWidth * multiplier;
+        final mod = ReplicationOpModule(
+            Logic(width: originalWidth), multiplier, newWidth);
+        await mod.build();
+        await SimCompare.checkFunctionalVector(mod, vectors);
+        final simResult = SimCompare.iverilogVector(
+            mod.generateSynth(), mod.runtimeType.toString(), vectors,
+            signalToWidthMap: {'a': originalWidth, 'b': newWidth});
```
mkorbel1, line 37: Recent merged PR makes `signalToWidthMap` unnecessary

test/replication_test.dart
```diff
@@ -0,0 +1,71 @@
...
+/// 2023 Jan 18
+/// Author: Akshay Wankhede <akshay.wankhede@intel.com>
+///
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd/src/utilities/simcompare.dart';
+import 'package:test/test.dart';
+
+class ReplicationOpModule extends Module {
+  ReplicationOpModule(Logic a, int multiplier, int newWidth) {
```
mkorbel1, line 16: Why is `newWidth` necessary?  Why can't we just calculate it by multiplying `a.width*multiplier`?

mkorbel1 (changes requested):
Looks good, just some small changes

test/replication_test.dart
```diff
@@ -0,0 +1,73 @@
...
+    final newWidth = a.width * multiplier;
+    a = addInput('a', a, width: a.width);
+    final b = addOutput('b', width: newWidth < 1 ? a.width : newWidth);
+
+    b <= a.replicate(multiplier);
+  }
+}
+
+void main() {
+  group('Logic', () {
```
mkorbel1, line 26: Add some test(s) for `LogicValue.replicate` as well?

mkorbel1 (changes requested):

mkorbel1 (approved):
Looks great!

mkorbel1:
Looks like just some conflicts to resolve then we can merge it!

mkorbel1 (approved):

PR #221, opened 2022-12-13
ROHD landing page website structure

quekyj:
## Description & Motivation

It's useful to have a web landing that displays ROHD. This site is going to act as the learning center and the compilation of the bootcamp and tutorial.

[https://quekyj.github.io/rohd/](https://quekyj.github.io/rohd/) to preview ROHD website.

## Related Issue(s)

None

## Testing

It's a static web page, don't think it requires any testing. 

## Backwards-compatibility

No. No major changes have been made to the codebase. 

## Documentation

No.

PR #256, opened 2023-02-06
Improve GitHub Codespaces configuration

chykon:
## Description & Motivation

Thanks to [contribution](https://github.com/intel/rohd/pull/226) tried GitHub Codespaces along with ROHD. Pretty handy for certain situations. I would like to suggest some improvements:

* The initial container configuration uses the [Dart (Community)](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/dart) template, which is marked as [Deprecated (No longer maintained)](< https://containers.dev/templates#:~:text=Dart,No%20longer%20maintained)>). It is proposed to replace it with the [actual](https://containers.dev/templates#:~:text=Ubuntu,Container%20Spec%20Maintainers) template [Ubuntu](https://github.com/devcontainers/templates/tree/main/src/ubuntu) version 22.04. The default template ([Default Linux Universal](https://github.com/devcontainers/templates/tree/main/src/universal)) is not suitable due to being based on Ubuntu 20.04, which would require compiling `Icarus Verilog` version 11 and increase container deployment time.
* Because the [Ubuntu](https://github.com/devcontainers/templates/tree/main/src/ubuntu) template is basic and generic, it requires installation of the software needed to work on ROHD. This is done with a setup script autorun by the `updateContentCommand` when the container is created.
* The `public key` for the Dart repository is suggested to be stored locally, rather than downloaded each time the container is created.
* When launching the container for VSCode, the [Dart extension](https://marketplace.visualstudio.com/items?itemName=Dart-Code.dart-code) will be installed. It was in the original configuration as well.

Separately, I would like to ask for opinions: is it necessary to automatically open files when starting codespace? Now the files `README.md`, `CONTRIBUTING.md`, `CODE_OF_CONDUCT.md` are opened. This should help new contributors, but it can also be annoying.

## Related Issue(s)

No.

## Testing

You can check:

[Open in GitHub Codespaces](https://github.com/codespaces/new?hide_repo_select=true&ref=update-codespaces&repo=597408556)

## Backwards-compatibility

No.

## Documentation

No, but it will be possible to add the GitHub Codespaces badge to `README.md`. Instructions [here](https://docs.github.com/en/codespaces/setting-up-your-project-for-codespaces/setting-up-your-repository/adding-a-codespaces-badge).


.devcontainer/devcontainer.json
```diff
@@ -1,28 +1,23 @@
...
-	// Use 'postCreateCommand' to run commands after the container is created.
-	// "postCreateCommand": "uname -a",
+  "updateContentCommand": "tool/gh_codespaces/run_setup.sh",
 
-	// Set `remoteUser` to `root` to connect as root instead. More info: https://aka.ms/vscode-remote/containers/non-root.
-	"remoteUser": "vscode"
+  "customizations": {
+    "codespaces": {
+      "openFiles": [
+        "README.md",
```
mkorbel1, line 12: I think these are great to have, but another that might be nice is `example/example.dart` so that first-time openers would immediately see a file they can run.

mkorbel1 (commented):
Thank you so much @chykon for this PR, this looks great!  I was able to open your branch and it all looks good.

I think an "Open in GitHub Codespaces" badge would be a great addition to the README.  If you like, you can add it to this PR, otherwise I will include it in another documentation-related PR that I'm working on.

.devcontainer/devcontainer.json
chykon: I added a file with an example and put it first in the queue for opening.

From what I understand, `README.md` is being completely redesigned, so I suggested adding an badge to <https://github.com/intel/rohd/pull/221#discussion_r1098544124>

mkorbel1 (approved):

@chykon, we really appreciate all the thought, time, and effort you've been putting into making ROHD and its repository better!

Issue #152, opened 2022-08-19
`Const`s where `val` is a `LogicValue` have unsanitary names


mkorbel1:
## Describe the bug
If you create a `Const` where `val` is a `LogicValue` (or some other legal `val` whose `toString` produces an invalid variable name), then the generated `name` for the `Logic` can be invalid since it is not sanitized.  The name of the signal should be sanitized.

## To Reproduce
```dart
Const(LogicValue.ofString('10xz'))
```

## Expected behavior
Constant is created without exception.

## Actual behavior
Exception due to unsanitary name.

aherrera97intl:
Will start working on this

PR #258, opened 2023-02-07
Added test to check sanitized names

aherrera97intl:
## Description & Motivation

Added function to validate sanitizer

## Related Issue(s)

https://github.com/intel/rohd/issues/152

## Testing

Call sanitizer to check on signal given

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/test/logic_name_test.dart
+++ b/test/logic_name_test.dart
@@ -27,7 +27,7 @@ void main() {
   });
 
   test('Test signals for sanitized names', () async {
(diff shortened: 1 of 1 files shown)


mkorbel1 (approved):
Looks good!

chykon:
Maybe the `Z value` should also be added to the test string?

aherrera97intl:
Done, added z to signal check

PR #262, opened 2023-02-09
Edit some project files

chykon:
## Description & Motivation

The following changes have been made:

* Changed `LICENSE` text to [template](https://github.com/github/choosealicense.com/blob/gh-pages/_licenses/bsd-3-clause.txt) so the license is now automatically recognized GitHub. The dates have also been updated.
* The order of the fields in `pubspec.yaml` has been changed to match that specified in [documentation](https://dart.dev/tools/pub/pubspec).
* In the `.gitignore` file:
   * the `pubspec.lock` rule has been moved to the `...created by pub` group;
   * `tmp*` changed to `tmp_test/` because after <https://github.com/intel/rohd/pull/173> the `tmp_coverage/` directory is no longer created in the project root;
   * added explicit extension or directory pointers for `.vcd` and `.vscode/`.
* `security.md` renamed to `SECURITY.md` to match the standard convention (like `LICENSE`, `README.md`, etc.).
* Removal of extra spaces and other small changes.

## Related Issue(s)

No.

## Testing

Testing is not required.

## Backwards-compatibility

No.

## Documentation

No.


LICENSE
```diff
@@ -1,28 +1,28 @@
...
-BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY,
-OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT
-OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE, DATA, OR PROFITS;
-OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY THEORY OF LIABILITY,
-WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT (INCLUDING NEGLIGENCE
-OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE OF THIS SOFTWARE,
-EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
- 
- 
-SPDX-License-Identifier: BSD-3-Clause
```
mkorbel1, line 28: Can we keep the SPDX identifier and still be automatically detected?

chykon: Tried to place the id in different places - doesn't work. I also tried to remove only the identifier from the original version of the license - and it was recognized. The GitHub documentation has [recommendations](https://docs.github.com/en/repositories/managing-your-repositorys-settings-and-features/customizing-your-repository/licensing-a-repository#detecting-a-license:~:text=If%20your%20repository%20is%20using%20a%20license%20that%20is,complexity%20somewhere%20else%2C%20such%20as%20your%20repository%27s%20README%20file.). It will be necessary to decide whether license detection is needed or not.

mkorbel1: I don't think we need to keep the SPDX identifier, especially if it's making it harder to automatically detect the license.  Thanks for looking into it!

mkorbel1 (approved):
Looks great, thank you!

I'll be implementing similar changes over in https://github.com/intel/rohd-vf and https://github.com/intel/rohd-cosim

PR #267, opened 2023-02-13
Slightly fix the example files

chykon:
## Description & Motivation

Examples should be given special attention, as the quality of the examples can both attract and repel new users from the project.

This set of changes proposes:

* Remove `ignore_for_file: avoid_print` from examples, import `dart:io` and use `stdout.writeln('string')` instead of `print('string')`. Subjectively, ignoring the rules of the linter in the examples does not look very nice.
* Ignore `*.sv` files, because the `fir_filter.dart` example creates a `rtl.sv` file (maybe in the future there will be more examples that save the generated SystemVerilog to a file).
* Slightly tweak comments to match [Effective Dart](https://dart.dev/guides/language/effective-dart/documentation#comments) (dots, capital letters, etc).
* In the `fir_filter.dart` file, the declaration code and comments have been slightly rearranged to make the example more consistent.
* Other minor changes.

## Related Issue(s)

No.

## Testing

The only more or less significant change in the code is the use of the `stdout.writeln('string')` function instead of `print('string')`. You can check the functionality by running the examples.

## Backwards-compatibility

Unfortunately, since the `dart:io` library is now used, it becomes impossible to compile examples in JS to run in browsers, but this should not be a problem since ROHD itself also uses the `dart:io` library.

## Documentation

`README.md` contains the code from `example.dart`, but since `README.md` is under revision, it is better to make changes in the new version of `README.md`.

### DIFF
--- a/example/example.dart
+++ b/example/example.dart
@@ -8,12 +8,8 @@
 /// Author: Max Korbel <max.korbel@intel.com>
 ///
 
-// For greater clarity in the example, one of the rules of code analysis
-// will be suppressed. Sometimes it can be useful, but you should always
-// evaluate the consequences of deviations from the set of static analysis
-// rules used in the project. Additional information is available at
-// https://dart.dev/guides/language/analysis-options
-//
+// Though we usually avoid them, for this example,
+// allow `print` messages (disable lint):
 // ignore_for_file: avoid_print
 
 // Import the ROHD package.
--- a/example/example.dart
+++ b/example/example.dart
@@ -8,8 +8,13 @@
 /// Author: Max Korbel <max.korbel@intel.com>
 ///
 
-// Import the standard library for I/O.
-import 'dart:io';
+// For greater clarity in the example, one of the rules of code analysis
+// will be suppressed. Sometimes it can be useful, but you should always
+// evaluate the consequences of deviations from the set of static analysis
+// rules used in the project. Additional information is available at
+// https://dart.dev/guides/language/analysis-options
+//
+// ignore_for_file: avoid_print
 
 // Import the ROHD package.
 import 'package:rohd/rohd.dart';
@@ -78,7 +83,7 @@ Future<void> main({bool noPrint = false}) async {
   // to other tools.
   final systemVerilogCode = counter.generateSynth();
   if (!noPrint) {
-    stdout.writeln(systemVerilogCode);
+    print(systemVerilogCode);
   }
 
   // Now let's try simulating!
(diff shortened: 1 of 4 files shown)


mkorbel1:
> Remove ignore_for_file: avoid_print from examples, import dart:io and use stdout.writeln('string') instead of print('string'). Subjectively, ignoring the rules of the linter in the examples does not look very nice.

I feel like using `stdout.writeln` instead of `print` is just another way of avoiding the lint check, plus might make it look more annoying to print to the console.  Java's `System.out.println` is intimidating to new users compared to the `print` of Python, and I think `stdout.writeln` is a step in the wrong direction.  Perhaps a comment explaining the lint disable motivation would offer two benefits:
- Teach how easy it is to configure/disable lints (something annoying in many hardware dev stacks)
- Explain that `print` is easy, but the motivation for having the lint there in the first place.

example/example.dart
```diff
@@ -8,19 +8,21 @@
 /// Author: Max Korbel <max.korbel@intel.com>
 ///
 
-// ignore_for_file: avoid_print
+// Import the standard library for I/O.
+import 'dart:io';
```
mkorbel1, line 12: See other comment on PR

chykon: Agree. I didn't think from that position. I'll revert `print` back and add an explanatory comment.

chykon:
Doesn't it seem that the explanatory comment is excessively big?

mkorbel1:
> Doesn't it seem that the explanatory comment is excessively big?

Haha it does seem a bit large... brings a lot of attention to it which could distract from the actual example.

Maybe something closer to this (not necessarily exactly this) would be better?
"Though we usually avoid them, for this example, allow `print` messages (disable lint)."

example/fir_filter.dart
```diff
@@ -1,115 +1,129 @@
+/// Copyright (C) 2022-2023 Intel Corporation
```
mkorbel1, line 1: The majority of this code was written by @wswongat who chose to omit a copyright statement.  It doesn't seem right to add an Intel copyright onto his contribution after the fact unless it also includes him as one of the authors.

mkorbel1 (changes requested):
One small change on the copyright

example/fir_filter.dart
chykon: I'll fix it. I thought the Intel copyright applied to any contribution anyway, either explicitly or implicitly (for those cases where it is not specified in the code).

chykon: Although yes, there may be times when this is not the case. For example, the same file with an example, taken from some third-party project.

mkorbel1 (approved):
Looks good!  Thank you!

PR #277, opened 2023-02-20
Add recommended extensions for VSCode

chykon:
## Description & Motivation

It is suggested to add information about [recommended extensions](https://code.visualstudio.com/docs/editor/extension-marketplace#_workspace-recommended-extensions) to the project. This is similar to <https://github.com/intel/rohd/blob/da23c8bb3a222c26322c6474f2466093f60eb8b6/.devcontainer/devcontainer.json#L18-L20> in the `devcontainer.json` file, but is only a recommendation and works even without using containers:

Снимок экрана от 2023-02-20 11-15-05

The motivation for the change is to make it easier for new members to contribute.

## Related Issue(s)

No.

## Testing

Removed the `Dart` extension, restarted the IDE, opened the cloned ROHD repository, got a suggestion to install the `Dart` extension.

## Backwards-compatibility

No.

## Documentation

This can be mentioned, for example, in `README.md`, but these changes are not included in this PR.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,8 @@ doc/api/
 # Other stuff
 tmp_test/
 *.vcd
-.vscode/
+.vscode/*
 
 # Exceptions
+!.vscode/extensions.json
 !test/example_icarus_waves.vcd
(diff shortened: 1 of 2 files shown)


mkorbel1 (approved):
Great idea, thank you!

PR #294, opened 2023-03-01
First implementation for don't care scenarios

### LINKED ISSUE #93: An equality operator for `LogicValues` that supports "don't care"
## Motivation
There are cases where you want to compare two `LogicValues` values for some bits, but not all bits.  It would be nice to support an equality check that enables a comparison of two `LogicValues` where any `x` is treated as a "don't care if this bit is equal".  Some thought would need to be put into whether only one side of the comparison should consider `x` as don't care or both sides, or maybe that should be configurable.  This feature is similar to a `casez` but for values instead of hardware.

## Desired solution
Add a function for comparison with don't care to `LogicValues`.

## Alternatives considered  
Let users implement this functionality themselves as needed.

dmetis:
## Description & Motivation 

There are cases where you want to compare two LogicValues values for some bits, but not all bits. It would be nice to support an equality check that enables a comparison of two LogicValues where any x is treated as a "don't care if this bit is equal". Some thought would need to be put into whether only one side of the comparison should consider x as don't care or both sides, or maybe that should be configurable. This feature is similar to a casez but for values instead of hardware.

## Description of changes, and motivation for adding them.

fix #93 

## Backwards-compatibility

## Documentation


lib/src/values/logic_value.dart
```diff
@@ -928,7 +928,13 @@ abstract class LogicValue {
     ].swizzle();
   }
 
-  ///compare two LogicValues values for some bits, but not all bits.
+  ///The goal here is to compare two LogicValues that are the same width
```
mkorbel1, line 931: Check out some guidelines on good doc comment writing here:
https://dart.dev/guides/language/effective-dart/documentation#doc-comments

The top line should describe what the function _does_.  In this case, it should be something like:
```dart
Checks if `this` is equal to [other], except ignoring bits of either which are not valid.

Returns `true` if each bit of `this` which [isValid] is equal to each bit of [other] which [isValid].
```

The first line is clear, short, and describes what it's doing.  Further lines below it explain more formally what it is doing without describing the actual implementation.

It would be good then to include some sort of example showing how it works.

mkorbel1 (approved):
Looks great!  Thank you!

PR #302, opened 2023-03-07
[issue#137]: Add mechanism for reset all flop of Sequential

### LINKED ISSUE #137: Add a mechanism to reset all flops of a `Sequential`
## Motivation
"It's cumbersome and error prone to have to keep resetting flops I'm adding"

When creating lots of flops in a `Sequential`, often you want them all reset by the same reset signal.  It would be convenient to make that an option.

## Desired solution
An option in `Sequential` or a new type of sequential object that handles reset across all flops it creates.

## Alternatives considered  
Currently, each flop must be manually reset.

RPG-coder-intc:
## Related Issue(s)
Issue #137 

Fix #137 

## Testing
- TBA

## Backwards-compatibility
- This will modify Sequential class 
- This feature will be modified to make the effects backward compatible

## Documentation
- TBA


lib/src/modules/conditional.dart
```diff
@@ -267,14 +267,18 @@ class Sequential extends _Always {
...
+  /// This sets the Sequential conditions to reset
+  Logic? reset;
+  Map<Logic, Logic> resetVal = {};
+
   /// Constructs a [Sequential] single-triggered by [clk].
   Sequential(Logic clk, List<Conditional> conditionals,
-      {String name = 'sequential'})
-      : this.multi([clk], conditionals, name: name);
+      {String name = 'sequential', Logic? reset})
+      : this.multi([clk], conditionals, name: name, reset: reset);
```
chykon, line 277: Maybe put the `reset` parameter before `name`? This does not affect the functionality (named parameters), but will keep it consistent with other modules that have `name` at the end.

lib/src/modules/conditional.dart
```diff
@@ -267,14 +267,18 @@ class Sequential extends _Always {
...
 
   /// Constructs a [Sequential] multi-triggered by any of [clks].
   Sequential.multi(List<Logic> clks, List<Conditional> conditionals,
-      {String name = 'sequential'})
+      {String name = 'sequential', this.reset})
```
chykon, line 281: Maybe put the `reset` parameter before `name`? This does not affect the functionality (named parameters), but will keep it consistent with other modules that have `name` at the end.

lib/src/modules/conditional.dart
```diff
@@ -30,9 +32,33 @@ abstract class _Always extends Module with CustomSystemVerilog {
 
   final Uniquifier _portUniquifier = Uniquifier();
 
-  _Always(this.conditionals, {super.name = 'always'}) {
+  _Always(this.conditionals,
```
mkorbel1, line 35: We should add a doc comment for the constructor of  `_Always` explaining the functionality of the reset-related stuff.

lib/src/modules/conditional.dart
```diff
@@ -20,7 +20,9 @@ import 'package:rohd/src/utilities/uniquifier.dart';
 /// Represents a block of logic, similar to `always` blocks in SystemVerilog.
 abstract class _Always extends Module with CustomSystemVerilog {
   /// A [List] of the [Conditional]s to execute.
-  final List<Conditional> conditionals;
+  late List<Conditional> conditionals;
+  late Logic? reset;
```
mkorbel1, line 24: Why do we need to save `reset` and `resetVal` here?  Also, are you even using them?

lib/src/modules/conditional.dart
```diff
@@ -267,15 +293,27 @@ class Sequential extends _Always {
   /// The input clocks used in this block.
   final List<Logic> _clks = [];
 
+  @override
+
+  /// This sets the Sequential conditions to reset
```
mkorbel1, line 298: I don't think we need `reset` and `resetVal` as members of `Sequential`, do we?

lib/src/modules/conditional.dart
```diff
@@ -267,15 +293,27 @@ class Sequential extends _Always {
...
+
+  /// This sets a particular Sequential conditions receiver to resetVal
+  // ignore: overridden_fields
+  Map<Logic, Logic> resetVal = {};
+
   /// Constructs a [Sequential] single-triggered by [clk].
   Sequential(Logic clk, List<Conditional> conditionals,
-      {String name = 'sequential'})
-      : this.multi([clk], conditionals, name: name);
+      {Logic? reset, String name = 'sequential'})
```
mkorbel1, line 310: I think you want to put `resetVal` in here also?

mkorbel1, line 310: You can use `super.reset` instead of collecting it yourself then passing it to the super constructor.

mkorbel1 (changes requested):
Looking on the right path!

lib/src/modules/conditional.dart
RPG-coder-intc: resetVal is contains reset values for receiver logic

if the reset Logic is set and the resetVal for a receiver(Logic) 
   then resetVal gets assigned to receiver instead of Const(0)
```
if(resetVal[reveiver]):
  rec < resetVal[reveiver]
else
  rec < Const(0)
```

mkorbel1: I mean why do you need to declare them as class fields instead of just using the existing local variables?

lib/src/modules/conditional.dart
```diff
@@ -32,11 +30,16 @@ abstract class _Always extends Module with CustomSystemVerilog {
 
   final Uniquifier _portUniquifier = Uniquifier();
 
-  _Always(this.conditionals,
-      {Logic? reset, Map<Logic, Logic>? resetVal, super.name = 'always'}) {
+  /// Accepts [conditionals] and set maps receivers to output
+  ///
+  /// If [reset] is set then For a receiver assign default reset value
+  /// as `0` unless a reset value for is present [resetValues]
+  _Always(List<Conditional> conditionals,
```
chykon, line 37: Is it possible to replace `List<Conditional> conditionals` with `this.conditionals`?

lib/src/modules/conditional.dart
```diff
@@ -46,19 +49,22 @@ abstract class _Always extends Module with CustomSystemVerilog {
...
           ],
           orElse: conditionals,
         ),
       ];
-    } else {
-      conditionals = conditionals;
     }
 
+    // ignore: prefer_initializing_formals
+    this.conditionals = conditionals;
```
chykon, line 66: If `List<Conditional> conditionals` in the constructor is replaced with `this.conditionals`, then this is no longer needed.

lib/src/modules/conditional.dart
```diff
@@ -293,17 +299,13 @@ class Sequential extends _Always {
...
-
   /// This sets the Sequential conditions to reset
   // ignore: overridden_fields
   Logic? reset;
 
-  @override
-
-  /// This sets a particular Sequential conditions receiver to resetVal
+  /// This sets a particular Sequential conditions receiver to resetValues
   // ignore: overridden_fields
```
chykon, line 307: It is possible to remove `ignore: overriden_fields`, since `@override` has been removed.

lib/src/modules/conditional.dart
mkorbel1: Same for `resetValues`

test/conditionals_test.dart
```diff
@@ -13,6 +13,35 @@ import 'package:rohd/src/exceptions/sim_compare/sim_compare_exceptions.dart';
 import 'package:rohd/src/utilities/simcompare.dart';
 import 'package:test/test.dart';
 
+import 'counter_wintf_test.dart';
+
+class ResetFlipflops extends Module {
```
mkorbel1, line 18: Why not add this test module and test to the `coutner_wintf_test.dart` file instead, so both implementations of the same module are in the same place?  You could maybe even parameterize the same `Counter` module to conditionally generate the reset two different ways, then reuse the same set of vectors.  There's too much copy-paste here, I think.

test/conditionals_test.dart
```diff
@@ -594,4 +623,25 @@ void main() {
     final simResult = SimCompare.iverilogVector(mod, vectors);
     expect(simResult, equals(true));
   });
+
+  test('resetFlipflop', () async {
```
mkorbel1, line 627: We're missing a test that tests `resetValues` as well?

mkorbel1 (changes requested):

test/counter_wintf_test.dart
```diff
@@ -86,4 +97,24 @@ void main() {
       expect(simResult, equals(true));
     });
   });
+  test('resetFlipflop from root', () async {
+    final mod = Counter(CounterInterface(8), resetRoot: true);
```
mkorbel1, line 101: instead of copy-pasting the whole test, you could make a function that generates a `test` with vs without `resetRoot` set.

mkorbel1 (changes requested):

lib/src/modules/conditional.dart
```diff
@@ -30,9 +30,43 @@ abstract class _Always extends Module with CustomSystemVerilog {
...
+  /// associated with the driven signal will be set to that value instead upon
+  /// reset.
+  _Always(this.conditionals,
+      {Logic? reset, Map<Logic, dynamic>? resetValues, super.name = 'always'}) {
     // create a registration of all inputs and outputs of this module
     var idx = 0;
+
+    // Get all Receivers
+    final allReceivers =
+        conditionals.map((e) => e.getReceivers()).expand((e) => e).toList();
```
mkorbel1, line 49: Please pull from `main`, there's been some changes in `Conditional`, so you can use `receivers` instead of `getReceivers`, which is an API deprecation that also gives you some perf boost here.

lib/src/modules/conditional.dart
```diff
@@ -20,7 +20,7 @@ import 'package:rohd/src/utilities/uniquifier.dart';
 /// Represents a block of logic, similar to `always` blocks in SystemVerilog.
 abstract class _Always extends Module with CustomSystemVerilog {
   /// A [List] of the [Conditional]s to execute.
-  final List<Conditional> conditionals;
+  late List<Conditional> conditionals;
```
mkorbel1, line 23: Can we change this to use an `UnmodifiableListView` instead?

```dart
List<Conditional> get conditionals => UnmodifiableListView<Conditional>(_conditionals);
late List<Conditional> _conditionals;
```

Otherwise it opens the door to people re-assigning the list after construction, which is weird and might cause some unpredictable behavior.

lib/src/modules/conditional.dart
```diff
@@ -30,9 +30,43 @@ abstract class _Always extends Module with CustomSystemVerilog {
...
+    // This will reset the conditionals on setting the `reset` flag
+    if (reset != null) {
+      conditionals = [
+        If(
+          reset,
+          then: [
+            ...allReceivers.map((rec) {
+              final driver = resetValues?[rec] ?? 0;
+              return rec < driver;
+              // If resetValue for a receiver is defined,
```
mkorbel1, line 60: nit: would be nicer to put the comment _before_ the code instead of _after_

test/counter_wintf_test.dart
```diff
@@ -32,7 +34,9 @@ class CounterInterface extends Interface<CounterDirection> {
 
 class Counter extends Module {
   late final CounterInterface intf;
-  Counter(CounterInterface intf) {
+  late bool resetRoot;
+  Counter(CounterInterface intf,
+      {this.resetRoot = false, bool withResetVal = false}) {
```
mkorbel1, line 39: It's a little confusing how you have this, and unclear what `resetRoot` means.  There's 2 different things you can control:
- Should it be reset the "old" way, or the new way you implemented
- What should the reset value be?

It might be clearer to have those two arguments in the constructor for this `Counter` instead:
- `bool useBuiltInSequentialReset`: if `true`, then use the new way, otherwise use the old way
- `int resetValue`: the value that should be reset, regardless of the mode

Then you can use a single if/else clause based on `useBuiltInSequentialReset`, and in each of the two cases use `resetValue` to reset it.  What do you think?

mkorbel1 (changes requested):
Nice work!  Some suggestions for improvement

lib/src/modules/conditional.dart
```diff
@@ -267,15 +301,26 @@ class Sequential extends _Always {
...
+  Logic? reset;
+
+  /// This sets a particular Sequential conditions receiver to resetValues
+  Map<Logic, dynamic>? resetValues = {};
+
   /// Constructs a [Sequential] single-triggered by [clk].
   Sequential(Logic clk, List<Conditional> conditionals,
-      {String name = 'sequential'})
-      : this.multi([clk], conditionals, name: name);
+      {Logic? reset,
```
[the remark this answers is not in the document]
RPG-coder-intc: this seems to throw error as soon as we delete class members variables 
`Logic? reset` & `Map<Logic, dynamic>? resetValues`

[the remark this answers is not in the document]
mkorbel1: ok, this looks ok

test/counter_wintf_test.dart
```diff
@@ -32,7 +34,9 @@ class CounterInterface extends Interface<CounterDirection> {
 
 class Counter extends Module {
   late final CounterInterface intf;
-  Counter(CounterInterface intf) {
+  late bool resetRoot;
```
mkorbel1, line 37: i think you're not using this `resetRoot` anymore, delete?

mkorbel1 (changes requested):

mkorbel1 (approved):

mkorbel1:
Fix #137

PR #303, opened 2023-03-08
[Issue-281] port initialized with empty name

### LINKED ISSUE #281: Port can be initialized with empty name
### Describe the bug

Port of the Module can be initialized with empty name.

### To Reproduce

```dart
// system verilog generate: input logic ,
final signal1 = addInput('', unamedSignal); 

```

### Expected behavior

throw exception on empty string port initialization

### Actual behavior

The initialization passed and the generated system verilog contains errors.

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

_No response_

quekyj:
## Description & Motivation

Port of the Module can be initialized with an empty name.

## Related Issue(s)

Fix #281 

## Testing

Added Logic name and definition name test

## Backwards-compatibility

No

## Documentation

No


lib/src/module.dart
```diff
@@ -142,6 +142,8 @@ abstract class Module {
     } else if (reserveDefinitionName &&
         !Sanitizer.isSanitary(definitionName!)) {
       throw InvalidReservedNameException();
+    } else if (definitionName == '') {
+      throw NullReservedNameException();
```
mkorbel1, line 146: Good catch on this same bug for `definitionName` in module, but what about `name`?

mkorbel1, line 146: Shouldn't this be thrown `reserveDefinitionName` is set to true?  If it's not, we should probably silently rename it to something safe like some auto-generated name?

lib/src/logic.dart
```diff
@@ -432,7 +432,9 @@ class Logic {
   /// The default value for [width] is 1.  The [name] should be synthesizable
   /// to the desired output (e.g. SystemVerilog).
   Logic({String? name, int width = 1})
-      : name = name == null ? 's${_signalIdx++}' : Sanitizer.sanitizeSV(name),
+      : name = (name == null || name == '')
```
mkorbel1, line 435: All the places where you're checking `== ''`, can we use `isEmpty` instead?

mkorbel1 (changes requested):

quekyj:
I am not sure why the test failed but worked on my localhost.

chykon:
> I am not sure why the test failed but worked on my localhost.

As far as I understand, this issue should be fixed by [this commit](https://github.com/intel/rohd/pull/306/commits/a990ee336b9b23ae498a67b621762171d0688508) in [this PR](https://github.com/intel/rohd/pull/306).

mkorbel1:
Yes that's right, sorry just merged #306 in.

lib/src/module.dart
```diff
@@ -142,6 +142,8 @@ abstract class Module {
     } else if (reserveDefinitionName &&
         !Sanitizer.isSanitary(definitionName!)) {
       throw InvalidReservedNameException();
+    } else if (reserveDefinitionName && definitionName!.isEmpty) {
+      throw NullReservedNameException();
```
mkorbel1, line 146: I think this should be an exception for invalid, not null?  Also, can't this not get hit because the sanitizer will always catch it?

lib/src/module.dart
mkorbel1: Why resolved?

mkorbel1 (changes requested):

lib/src/module.dart
quekyj: My bad, I thought name and definition share on `_nameValidation` function. I added `_nameValidation` to the name as well. It should have the same behaviour now.

lib/src/module.dart
quekyj: Okay, I swap the priority now and create a new exception on empty string. Feel like better to separate the exception message to avoid confusion.

lib/src/module.dart
```diff
@@ -139,6 +139,8 @@ abstract class Module {
       String? definitionName, bool reserveDefinitionName) {
```
mkorbel1, line 139: Probably want to rename these to `name` and `reserveName` since they could be used for `definitionName` or `name`?

test/logic_name_test.dart
```diff
@@ -42,4 +42,20 @@ void main() {
...
+    final bus = Logic(name: '');
+    expect(bus.name, isNot(equals('')));
+  });
+
+  group('port name:', () {
+    test('GIVEN port name is empty string THEN expected to see exception',
+        () async {
+      expect(() async {
+        LogicTestModule('');
+      }, throwsA((dynamic e) => e is Exception));
```
mkorbel1, line 58: Why not the specific type of exception?

mkorbel1 (changes requested):
Looking good, just minor things

mkorbel1 (changes requested):
Very close, nice work!

mkorbel1 (approved):
Looks good!

PR #310, opened 2023-03-13
update chapter 2 bootcamp

quekyj:
## Description & Motivation

Update chapter 2 bootcamp by adding examples and fix assign statement does not show up. 

## Related Issue(s)

None

## Testing

None

## Backwards-compatibility

No

## Documentation

No


doc/tutorials/chapter_2/h_bus_range_swizzling.dart
```diff
@@ -31,6 +31,9 @@ void slicing(Logic a, Logic b, Logic c, Logic d, Logic e, Logic f) {
   // right swizzle: d = [1] (MSB), c = [00111], a = [1110] (LSB),
   // e = [1110 00111 1] - [a, c, d]
   f <= [d, c, a].rswizzle();
+
+  // Get a copy of f
```
mkorbel1, line 35: Confusing comment, it's a copy with one bit set to 0 right?

mkorbel1 (changes requested):

quekyj:
Why is this a dead link in Github action? 

```
FILE: ./CONTRIBUTING.md
[✖] https://dart.dev/guides/language/effective-dart

22 links checked.

ERROR: 1 dead links found!
[✖] https://dart.dev/guides/language/effective-dart → Status: 0
```

chykon:
> Why is this a dead link in Github action?

You can try restarting CI after a while (maybe the site was down or DDoS protection worked), or add it to the ignore list (<https://github.com/intel/rohd/blob/main/.github/configs/mlc_config.json>) as we are sure this is a false positive.

When adding to the ignore list, it is advisable to adhere to the pattern:

* first `^`
* then `LINK`
* at the end of `$`

We use regular expression syntax to mark the start (`^`) and end (`$`) of text, which allows us to ignore only the address that exactly matches the specified one, and not any addresses that contain such a substring. An example is already on the ignore list itself.

quekyj:
> You can try restarting CI after a while (maybe the site was down or DDoS protection worked), or add it to the ignore list (https://github.com/intel/rohd/blob/main/.github/configs/mlc_config.json) as we are sure this is a false positive.

Do you think the whitelist should be added to this pull request or better separate it into other pull request instead?

chykon:
It would probably be worth making the changes related to the ignore settings in a separate PR, but I personally did not think about it until you rightly pointed it out. It's probably best to check with the maintainer of the repository.

mkorbel1 (approved):
Looks good!

mkorbel1:
I think you need to pull master to fix the failing test

Issue #299, opened 2023-03-03
Logic operation fails on LogicValue wider than 127


ganewto:
https://github.com/intel/rohd/blob/344db0e18fa702eda056ed1ddbcda7717e154c17/lib/src/values/logic_value.dart#L783

The conversion to Int rather than BigInt means that logic operations can fail when LogicValues are wide.
The limit appears to be 128.

mkorbel1:
I think the issue is not that we should have `toBigInt()` here, but rather that the check directly above it is insufficient for all possible scenarios where it is not convertible to an int.  For example, it looks like if we have `_FilledLogicValue` of `width > 64` we may still hit `toInt()` in the `else` clause.

The check should probably check the `width` of `this` and `other` if they are `LogicValue`s instead of the type.

ganewto:
Here I have a testcase that replicates the problem:

`  test('crash_compare', () {
    final input = Const(BigInt.from(2).pow(128), width: 129);
    final output = Logic();
    Combinational([
      IfBlock([
        Iff(input.getRange(0, 128) > BigInt.from(0),
            [output < Const(1, width: 1)]),
        Else([output < Const(0, width: 1)]),
      ])
    ]);
  });`

ganewto:
I posted my first pull request.  Let me know if there are better ways to
commit these changes as well as if the changes I am suggesting are
sufficient.

There is still an outstanding issue on sign-extension when converting to
BigInt which causes a problem.

mkorbel1:
@Sanchit-kumar found a simple test case that reproduces this bug in #336:

```dart
LogicValue.ofBigInt(BigInt.zero, 128) + LogicValue.ofBigInt(BigInt.zero, 128);
```

PR #319, opened 2023-03-17
two tests, one fix for wide bitvector comparison

ganewto:
## Description & Motivation

Adds two tests for wide bitvector comparisons.  Adds a fix for one of them.
The remaining problem is that shifting a 1 into the signed position and then converting to bigint results in a negative number regardless.

## Related Issue(s)

None

## Testing

Created two tests that do comparisons.  One just breaks the length check.  The other puts a 1 in the sign location just before conversion to BigInt (the fix for the first).

## Backwards-compatibility

The change is in the core logic comparison operation so more tests may be needed.

## Documentation

No

### DIFF
--- a/lib/src/values/logic_value.dart
+++ b/lib/src/values/logic_value.dart
@@ -699,7 +699,11 @@ abstract class LogicValue {
       return LogicValue.filled(other.width, LogicValue.x);
     }
 
-    if (this is _BigLogicValue || other is BigInt || other is _BigLogicValue) {
+    if (this is _BigLogicValue ||
+        other is BigInt ||
+        other is _BigLogicValue ||
+        (this is _FilledLogicValue) && this.width >= 64 ||
+        (other is _FilledLogicValue) && other.width >= 64) {
       final a = toBigInt();
       final b = other is BigInt
           ? other
@@ -775,7 +779,11 @@ abstract class LogicValue {
 
     dynamic a;
     dynamic b;
-    if (this is _BigLogicValue || other is BigInt || other is _BigLogicValue) {
+    if (this is _BigLogicValue ||
+        other is BigInt ||
+        other is _BigLogicValue ||
+        (this is _FilledLogicValue) && this.width >= 64 ||
+        (other is _FilledLogicValue) && other.width >= 64) {
       a = toBigInt();
       b = other is BigInt
           ? other
--- a/test/bigint_test.dart
+++ b/test/bigint_test.dart
@@ -1,0 +1,36 @@
+//
+// bigint_test.dart
+// Test of very wide bitvector comparison failure
+//
+//
+
+import 'dart:math';
+import 'package:rohd/rohd.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('crash compare', () {
+    final input = Const(BigInt.from(2).pow(128), width: 129);
+    final output = Logic();
+    Combinational([
+      IfBlock([
+        Iff(input.getRange(0, 128) > BigInt.from(0),
+            [output < Const(1, width: 1)]),
+        Else([output < Const(0, width: 1)]),
+      ])
+    ]);
+  });
+  test('bad compare', () {
+    const i = 64;
+    final input = Const(BigInt.from(1) << (i - 1), width: i);
+    final output = Logic();
+    Combinational([
+      IfBlock([
+        Iff(input > BigInt.from(0), [output < Const(1, width: 1)]),
+        Else([output < Const(0, width: 1)]),
+      ])
+    ]);
+    final b = ~input.eq(0);
+    expect(output.value, equals(b.value));
+  });
+}


quekyj:
Maybe update your Related Issue to verb Fix to allow auto close of issue 299 once the pull request being merged. 

E.g.
## Related issue(s)

Fix #299

lib/src/values/logic_value.dart
```diff
@@ -699,7 +699,11 @@ abstract class LogicValue {
       return LogicValue.filled(other.width, LogicValue.x);
     }
 
-    if (this is _BigLogicValue || other is BigInt || other is _BigLogicValue) {
+    if (this is _BigLogicValue ||
```
mkorbel1, line 702: Why not just always check `this.width` and `other.width` and never check the type with `is`?

Also, let's use `_INT_BITS` instead of 64.  Eventually this may compile to javascript and then we'll only have 53 bits for ints.  Having fewer hard-coded 64's around will make things easier.

test/bigint_test.dart
```diff
@@ -0,0 +1,36 @@
+//
```
mkorbel1, line 1: Please include the SPDX identifier near the top of the file
```dart
```

test/bigint_test.dart
```diff
@@ -0,0 +1,36 @@
...
+// bigint_test.dart
+// Test of very wide bitvector comparison failure
+//
+//
+
+import 'dart:math';
+import 'package:rohd/rohd.dart';
+import 'package:test/test.dart';
+
+void main() {
```
mkorbel1, line 11: I think a test that could easily reproduce is just performing operations on a `LogicValue` directly without any actual hardware.  Worth adding that one?  I think any >64-bit `LogicValue` with any operation (e.g. `+`) against another `LogicValue` which is "filled" (all the same value, like all 1's) should be sufficient.

test/bigint_test.dart
```diff
@@ -0,0 +1,36 @@
+//
+// bigint_test.dart
```
mkorbel1, line 2: This file actually has tests related to `LogicValue` operations between large `LogicValue`s.  Naming it `bigint_test` is maybe not quite right.  Maybe `big_logic_value_test`?  Or we could throw it in the (already huge) `logic_value_test.dart`.

test/bigint_test.dart
```diff
@@ -0,0 +1,36 @@
...
+// Test of very wide bitvector comparison failure
+//
+//
+
+import 'dart:math';
+import 'package:rohd/rohd.dart';
+import 'package:test/test.dart';
+
+void main() {
+  test('crash compare', () {
```
mkorbel1, line 12: Good catch that this affects both `_doMath` and `_doCompare`, but don't these tests only test agains the `_doCompare` bug?

lib/src/values/logic_value.dart
```diff
@@ -699,7 +699,11 @@ abstract class LogicValue {
...
+        other is BigInt ||
+        other is _BigLogicValue ||
+        (this is _FilledLogicValue) && this.width >= 64 ||
```
mkorbel1, line 705: I don't think this should be `>=`, just `>`.  However, it appears there's a separate signed bug related to when the value is equal to 64 that you've discovered.  You could file it as a separate bug or just include some description of it added in the PR.

I thought a fix could be to use [`toUnsigned`](https://api.flutter.dev/flutter/dart-core/int/toUnsigned.html) in the `toInt` functions for LogicValue implementations, but it seems like that's limited to 63 bits.  So perhaps the `>=` is ok for the `_doCompare`, but the `_doMath` should support all 64 bits.  We still should add some additional tests around the fix to make sure it actually works for 63, 64, and >64 bit numbers (including ones that could be interpreted as negative) for both comparison and math.

mkorbel1 (changes requested):
Thank you for the contributions and tests!

ganewto:
I made a one-line change that fixes the original test.  I haven't made tests for the other observations on the 64-bit boundary,

test/big_logic_value_test.dart
```diff
@@ -0,0 +1,79 @@
...
+    for (final len in [63, 64, 65]) {
+      final sslv = LogicValue.ofInt(3, len); // small Int hold Big
+      final bslv = LogicValue.ofInt(-0xFFFF, len); // 18446744073709486081
+      final fslv = LogicValue.ofInt(-2, len); // 18446744073709551614
+
+      final sblv = LogicValue.ofBigInt(BigInt.from(3), len);
+      final bblv = LogicValue.ofBigInt(BigInt.from(-0xFFFF), len);
+      final fblv = LogicValue.ofBigInt(BigInt.from(-2), len);
+
+      test('small Int storage', () {
```
mkorbel1, line 45: might want to/need to name the test (here and below) based on the `len`?  Like 
```dart
      test('small Int storage len=$len', () {
```

test/big_logic_value_test.dart
```diff
@@ -0,0 +1,79 @@
...
+        expect(fblv < sblv, LogicValue.zero);
+        expect(sblv > fblv, LogicValue.zero);
+        expect(fblv > sblv, LogicValue.one);
+
+        expect(bblv < fblv, LogicValue.one);
+        expect(fblv < bblv, LogicValue.zero);
+        expect(bblv > fblv, LogicValue.zero);
+        expect(fblv > bblv, LogicValue.one);
+      });
+    }
```
mkorbel1, line 77: I think we're missing tests for math stuff.  I think `+`, `-`, `*` should work with two's complement (would be good to have a test anyways for them), but I'm not sure about `/`.  Effectively we're not fully covering the `_doMath` portion of stuff

mkorbel1 (changes requested):
thank you this is looking great so far!

lib/src/values/logic_value.dart
```diff
@@ -552,9 +553,9 @@ abstract class LogicValue implements Comparable<LogicValue> {
   String toString({bool includeWidth = true}) {
     if (isValid && includeWidth) {
       // for ==_INT_BITS, still use BigInt so we don't get negatives
-      final hexValue = width >= _INT_BITS
+      final hexValue = width > _INT_BITS
           ? toBigInt().toUnsigned(width).toRadixString(16)
-          : toInt().toRadixString(16);
+          : toBigInt().toRadixString(16);
```
mkorbel1, line 558: what's the point of having different logic for different widths here then?  Maybe just reverting to `width >= _INT_BITS` is sufficient so that for 63 and less we dont have to build a BigInt?

test/big_logic_value_test.dart
```diff
@@ -74,6 +74,31 @@ void main() {
...
+        expect(sblv - fblv, LogicValue.ofInt(6, len));
+        expect(fblv - sblv, LogicValue.ofInt(-6, len));
+
+        expect(sblv * fblv, LogicValue.ofInt(-8, len));
+
+        expect(sblv + fblv, LogicValue.ofBigInt(BigInt.from(2), len));
+        expect(sblv - fblv, LogicValue.ofBigInt(BigInt.from(6), len));
+        expect(fblv - sblv, LogicValue.ofBigInt(BigInt.from(-6), len));
+
+        expect(fblv * sblv, LogicValue.ofBigInt(BigInt.from(-8), len));
```
mkorbel1, line 100: Nice, what about division and modulo though?

Also, with these tests perhaps we can remove the "WARNING: Signed math is not fully tested." messages in logic_value.dart?

lib/src/values/logic_value.dart
desmonddak: This is string output, so likely performance is not a big issue. if the width==64, the sign gets output, when I don't think we want that, we want to see the raw bits in radix form.

test/big_logic_value_test.dart
desmonddak: division doesn't seem to work properly for negative numbers.
8 / -2  should be -4, but it is 0.
I felt I needed to understand Dart ints and LogicValue a bit more before I figure out the problem.  
Currently this is not a big/little problem, but an int problem.

mkorbel1: Ok, got it, I filed a separate issue to track here so we're not hung up on this PR
https://github.com/intel/rohd/issues/411

lib/src/values/logic_value.dart
mkorbel1: This code seems redundant and simple to optimize.  Since `toBigInt()` in all cases returns an unsigned `BigInt` now (there's a `toUnsigned` beforehand for all three cases), then both of these lines are effectively the same:
```dart
toBigInt().toUnsigned(width).toRadixString(16)
toBigInt().toRadixString(16)
```

While performance for simulation is not impacted by this, it is relevant for SystemVerilog generation (constants), which is somewhat performance sensitive, although much less so.

I guess what I'm saying is one of these two options makes more sense to me:

Option 1: (only extra perf impact for 64 bit, since >64 already must do the `toBigInt`)
```dart
      final hexValue = width >= _INT_BITS // include 64-bit
          ? toBigInt().toRadixString(16)
          : toInt().toRadixString(16);
```

Option 2: (simpler solution, always perf impact)
```dart
      final hexValue = toBigInt().toRadixString(16);
```

I think option 1 is better?

desmonddak: I tested version 1and pushed.

mkorbel1: Did the push go through?  Still looks the same?

lib/src/values/logic_value.dart
```diff
@@ -881,7 +882,7 @@ abstract class LogicValue implements Comparable<LogicValue> {
     }
     if (this is BigInt ||
         this is _BigLogicValue ||
-        (this is _FilledLogicValue && width >= _INT_BITS)) {
+        (this is _FilledLogicValue && width > _INT_BITS)) {
```
mkorbel1, line 885: I didn't notice before that you caught and fixed this one too, thanks! We should add some unit test that hits the unary math on these borders as well?

lib/src/values/logic_value.dart
desmonddak: I just pushed the toString performance fix (not converting to BigInt unless >= _INT_MAX.
This case is special (I guess I should comment on it) because we really DO want to convert to BigInt at 64 in this case only to ignore the sign.

lib/src/values/logic_value.dart
desmonddak: unary math?  Can you give an example?
I'm looking at the divide issue as well.

mkorbel1: You can see the complete set of things that use it by finding "find all references" to `_doUnaryMath`, which I'm surprised to find (forgot) is only the `clog2` function.  It looks like there's a lot of tests already in "logic_value_test.dart", I dont know if the existing tests cover the types of scenarios being addressed in this PR though.  The other "unary" math like and/or/xor reduction use their own custom implementations per-flavor already.

lib/src/values/logic_value.dart
```diff
@@ -18,6 +18,7 @@ typedef LogicValues = LogicValue;
 ///
 /// Each bit of [LogicValue] can be represented as a [LogicValue]
 /// of `0`, `1`, `x` (contention), or `z` (floating).
+/// LogicValue is unsigned
```
mkorbel1, line 21: Would be better to make this a new line and a sentence with a period, and put [] around it. 

 i.e.

```dart
/// An immutable 4-value representation of an arbitrary number of bits.
///
/// Each bit of [LogicValue] can be represented as a [LogicValue]
/// of `0`, `1`, `x` (contention), or `z` (floating).
///
/// [LogicValue] is unsigned.
```

test/big_logic_value_test.dart
mkorbel1, line 1: The name of the file doesn't really match what we're doing in here.  Maybe we should call it "logic_value_math_test.dart" or something?

test/big_logic_value_test.dart
```diff
@@ -0,0 +1,149 @@
+import 'package:rohd/rohd.dart';
```
mkorbel1, line 1: Please add a file header for this new file (see another file as an example)

test/big_logic_value_test.dart
```diff
@@ -0,0 +1,149 @@
...
+    final output = Logic();
+    Combinational([
+      If.block([
+        Iff(input.getRange(0, 128) > BigInt.from(0),
+            [output < Const(1, width: 1)]),
+        Else([output < Const(0, width: 1)]),
+      ])
+    ]);
+  });
+  test('bad compare', () {
```
mkorbel1, line 16: nit: would be nice to put some space (blank line) between `test`s

lib/src/values/logic_value.dart
mkorbel1, line 1: There's a number of places in this file with the comment `WARNING: Signed math is not fully tested.`.  I think you've now resolved a lot of that, so let's remove those?

mkorbel1 (changes requested):
I think this all looks great to me, thank you so much!

Just a handful of minor cosmetic/comment changes before merging it in

mkorbel1 (approved):
Looks great, thank you!

Issue #324, opened 2023-03-20
`WaveDumper` cannot create missing directories in the `outputPath` path


chykon:
### Describe the bug

`WaveDumper` cannot create missing directories in the `outputPath` path.

### To Reproduce

1. Remove `tmp_test`
2. Run `benchmark/wave_dump_benchmark.dart`

### Expected behavior

The missing directories will be created.

### Actual behavior

Missing directories are not created, resulting in an exception.

### Additional: Dart SDK info

Dart SDK version: 2.19.3 (stable) (Tue Feb 28 15:52:19 2023 +0000) on "linux_x64"

### Additional: Context

ROHD v0.4.2

mkorbel1:
Thank you for filing!

PR #326, opened 2023-03-20
[Fix 324] Add `WaveDumper` ability to create missing output directories

chykon:
## Description & Motivation

These changes fix the issue https://github.com/intel/rohd/issues/324

## Related Issue(s)

Fix https://github.com/intel/rohd/issues/324

## Testing

Added test to check if missing directories can be created.

## Backwards-compatibility

No.

## Documentation

No.


mkorbel1 (approved):
Looks great, thank you for fixing this!

PR #327, opened 2023-03-21
Small improvements for `Architecture.md`, `TreeExample.md`

chykon:
## Description & Motivation

It is proposed to make small changes to the `Architecture.md` and `TreeExample.md` files:

* Rename files to follow `snake_case`
* Fix typos, remove extra spaces
* Update code in `TreeExample.md` (updated code is taken from `example/tree.dart`)
* Update generated SystemVerilog in `TreeExample.md`
* Update links to renamed files

## Related Issue(s)

No.

## Testing

Not required.

## Backwards-compatibility

Old links to renamed files will no longer work.

## Documentation

The proposed changes for `Architecture.md` and `TreeExample.md` do not affect the actual content.


chykon:
Need to restart CI after some time due to external links check failure.

chykon:
Because `README.md` currently cannot use relative links (this results in an error when checking the documentation) (<https://github.com/dart-lang/dartdoc/issues/1096>), there was a chicken-and-egg problem with checking links. It is proposed to temporarily add the path to the new file to the ignore list so that the PR will complete successfully.

If such a solution is suitable, then we will need to remember to remove the path from the ignore list.

mkorbel1 (approved):
Looks good to me, thanks!

The workaround with ignore seems acceptable to me

PR #333, opened 2023-03-27
Update chapter 3 bootcamp

quekyj:
## Description & Motivation

Added exercises and answers for chapter 3. 

## Related Issue(s)

None

## Testing

None.

## Backwards-compatibility

Nope.

## Documentation

Nope


mkorbel1 (approved):
This looks good!

PR #340, opened 2023-04-11
[Issue 246] SynthBuilder should check for Module.build()

### LINKED ISSUE #246: `SynthBuilder` should check modules are already built
### Describe the bug

It's possible to generate outputs using a `Synthesizer` and `SynthBuilder` rather than through `Module.generateSynth`.  However, the `SynthBuilder` does not check that `Module.build` is called, only `generateSynth` does.  This can produce unexpected outputs (e.g. SystemVerilog with no submodules) without any helpful error message indicating why.

### To Reproduce

Use `SynthBuilder` to generate output verilog without calling `build` on the `Module`.

### Expected behavior

`ModuleNotBuiltException` is thrown

### Actual behavior

Silent incorrect results

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

_No response_

quekyj:
## Description & Motivation

It's possible to generate outputs using a Synthesizer and SynthBuilder rather than through Module.generateSynth. However, the SynthBuilder does not check that Module.build is called, only generateSynth does. This can produce unexpected outputs (e.g. SystemVerilog with no submodules) without any helpful error message indicating why.

## Related Issue(s)

Fix #246 

## Testing

Tested with module not built and catch exception. 

## Backwards-compatibility

Yes. 

## Documentation

No. Just added validation and test.

### DIFF
--- a/lib/src/synthesizers/synth_builder.dart
+++ b/lib/src/synthesizers/synth_builder.dart
@@ -10,6 +10,7 @@
 
 import 'package:collection/collection.dart';
 import 'package:rohd/rohd.dart';
+import 'package:rohd/src/exceptions/module/module_not_built_exception.dart';
 import 'package:rohd/src/utilities/uniquifier.dart';
 
 /// A generic class which can convert a module into a generated output using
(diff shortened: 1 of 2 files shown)


mkorbel1 (approved):
Looks good, thank you!

PR #350, opened 2023-04-27
Fix #148: Implemented neq

### LINKED ISSUE #148: Implement `neq`
## Motivation
Equality checks with `eq` are simple, but you have to use `~...eq` for not-equal-to.  It would be nice to be able to do not-equal-to using `neq`.

## Desired solution
Implement `neq` everywhere that `eq` exists (for values and signals).

Sanchit-kumar:
## Description & Motivation
Equality checks with eq are simple, but need to write ~...eq for not-equal-to. It would be nice to be able to do not-equal-to using neq.

## Testing

Updated the test/comparison_test.dart  file for neq w.r.t already used values for the other supported operations.

## Documentation

Yes. I have updated both README.md and /doc/tutorials/chapter_2/00_basic_logic.md for neq operation.

### DIFF
--- a/lib/src/values/logic_value.dart
+++ b/lib/src/values/logic_value.dart
@@ -725,6 +725,14 @@ abstract class LogicValue {
   /// [eq] will return `x`.
   LogicValue eq(dynamic other) => _doCompare(other, (a, b) => a == b);
 
+  /// Not equal-to operation.
+  ///
+  /// This is different from [!=] because it returns a [LogicValue] instead
+  /// of a [bool].  It does a logical comparison of the two values, rather
+  /// than exact inequality.  For example, if one of the two values is invalid,
+  /// [neq] will return `x`.
+  LogicValue neq(dynamic other) => _doCompare(other, (a, b) => a != b);
+
   /// Less-than operation.
   ///
   /// WARNING: Signed math is not fully tested.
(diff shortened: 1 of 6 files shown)


lib/src/values/logic_value.dart
```diff
@@ -725,6 +725,14 @@ abstract class LogicValue {
   /// [eq] will return `x`.
   LogicValue eq(dynamic other) => _doCompare(other, (a, b) => a == b);
 
+  /// Not equal-to operation.
+  ///
+  /// This is different from != operator because it returns a [LogicValue] instead
+  /// of a [bool].  It does a logical comparison of the two values, rather
+  /// than exact inequality.  For example, if one of the two values is invalid,
+  /// [neq] will return `x`.
+  LogicValue neq(dynamic other) => _doCompare(other, (a, b) => a != b);
```
mkorbel1, line 734: The `neq` tests for `Logic` look great, but can we add some for `LogicValue` as well?  There are some tests in `logic_value_test.dart` that you could use for reference.

mkorbel1 (changes requested):

mkorbel1 (approved):
Great work, thank you for contributing this!

For the future, it's a good idea to check if issues have an assignee already in case someone else is already making progress on it

Issue #140, opened 2022-07-27
Add `gt` and `gte` for consistency to `Logic`


mkorbel1:
## Motivation
For equality, less than, and less than or equal to, `Logic` has `eq`, `lt`, and `lte`, respectively, since those operators have different meanings (object equality, conditional assignment, and assignment, respectively).  Since greater than and greater than or equal to do not use those operators for something else, they were overloaded for `>` and `>=`.  However, this is a little inconsistent and confusing to new users.  It would be nice if you could also optionally use `gt` and `gte` for consistency.

## Desired solution
Add `gt` and `gte` implementations to `Logic`.  Do not remove `>` and `>=` overloaded operators.

IdiBetzSch:
Hey. 

Picked this one up while familiarizing myself with Dart. Seems like I have it done(wasn't much to it), but running myself through the test and such getting a feel for the setup right now.

Sanchit-kumar:
Picking this one up.

PR #351, opened 2023-04-28
Issue#140: Added gt and gte for for consistency to Logic

Sanchit-kumar:
## Description & Motivation

As per #140 , for equality, less than, and less than or equal to, Logic has eq, lt, and lte, respectively, since those operators have different meanings (object equality, conditional assignment, and assignment, respectively). Since greater than and greater than or equal to do not use those operators for something else, they were overloaded for > and >=. However, this is a little inconsistent and confusing to new users. It would be nice if you could also optionally use gt and gte for consistency. 

## Testing

Added test cases in test/comparison_test.dart 

## Documentation

Yes. I have updated README.md and doc/tutorials/chapter_2/00_basic_logic.md for the support of _gt_ and _gte._

### DIFF
--- a/lib/src/logic.dart
+++ b/lib/src/logic.dart
@@ -581,6 +581,12 @@ class Logic {
   Logic lte(dynamic other) => LessThanOrEqual(this, other).out;
 
   /// Greater-than.
+  Logic gt(dynamic other) => GreaterThan(this, other).out;
+
+  /// Greater-than-or-equal-to.
+  Logic gte(dynamic other) => GreaterThanOrEqual(this, other).out;
+
+  /// Greater-than.
   Logic operator >(dynamic other) => GreaterThan(this, other).out;
 
   /// Greater-than-or-equal-to.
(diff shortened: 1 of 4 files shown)


lib/src/logic.dart
```diff
@@ -580,6 +580,12 @@ class Logic {
   /// Less-than-or-equal-to.
   Logic lte(dynamic other) => LessThanOrEqual(this, other).out;
 
+  /// Greater-than.
+  Logic gt(dynamic other) => GreaterThan(this, other).out;
```
mkorbel1, line 584: Looks like this PR is missing implementations for `GreaterThan` and `GreaterThanOrEqual`?

mkorbel1 (changes requested):
Thank you for taking this up!

lib/src/logic.dart
mkorbel1: whoops nevermind, forgot it already exists!

mkorbel1 (approved):

Issue #336, opened 2023-03-31
Add power functionality


mkorbel1:
### Motivation

In SystemVerilog, power (`a**b`) is a useful function.

It would be nice to have a synthesizable `Module` and associated functions for generating hardware and SystemVerilog with powers.

### Desired solution

- Create a new `Module` that generates `a**b` in output SystemVerilog and appropriately simulates powers in the ROHD simulator.
- Add functions (probably to `Logic`) to automatically generate this module in a convenient API.

PR #356, opened 2023-05-06
Issue #336: Added power functionality

Sanchit-kumar:
## Description & Motivation
In SystemVerilog, power (a**b) is a useful function.

It would be nice to have a synthesizable Module and associated functions for generating hardware and SystemVerilog with powers as per issue #336 

## Testing

Added necessary test cases to test/logic_value_test.dart and test/math_test.dart

## Documentation

Yes. Updated both README.md and doc/tutorials/chapter_2/00_basic_logic.md


lib/src/values/logic_value.dart
```diff
@@ -733,6 +733,13 @@ abstract class LogicValue {
   /// invalid, [neq] will return `x`.
   LogicValue neq(dynamic other) => _doCompare(other, (a, b) => a != b);
 
+  /// Power operation.
+  ///
+  /// This will return a [LogicValue] of some input 'base' to the power of other
+  /// input 'exponent'. If one of the two input values is invalid, [pow] will
+  /// return ‘x’ of input size width.
+  LogicValue pow(dynamic other) => _doMath(other, _binPower);
```
mkorbel1, line 741: Should we change the name to `exponent` instead of `other`?  You can also reference in the comment as `[exponent]`

test/logic_value_test.dart
```diff
@@ -551,6 +551,43 @@ void main() {
...
+              256)));
+      expect(
+          //test string
+          LogicValue.ofString('000010').pow(LogicValue.ofString('000100')),
+          equals(LogicValue.ofString('010000')));
+      expect(
+          //test invalid input
+          LogicValue.ofString('0001').pow(LogicValue.ofString('000x')),
+          equals(LogicValue.filled(4, LogicValue.x)));
+    });
```
mkorbel1, line 589: add test that if the base is invalid, then output is also invalid?

lib/src/values/logic_value.dart
```diff
@@ -761,6 +768,35 @@ abstract class LogicValue {
       // ignore: avoid_dynamic_calls
       _doCompare(other, (a, b) => (a >= b) as bool);
 
+  /// Binary Power
+  ///
+  /// Both inputs [a] and [b] are either of type [int] or [BigInt].
+  /// Returns [a] raise to the power [b] of input type.
+  dynamic _binPower(dynamic a, dynamic b) {
```
mkorbel1, line 775: can we leverage built-in `pow` methods for `int` and `BigInt` since they are more efficient and already well-tested?
https://api.dart.dev/stable/2.19.6/dart-math/pow.html
https://api.flutter.dev/flutter/dart-core/BigInt/pow.html

mkorbel1 (changes requested):
Looks very good, thank you for the contribution!  A few suggestions

test/logic_value_test.dart
Sanchit-kumar: Yes. Added.

lib/src/values/logic_value.dart
```diff
@@ -761,6 +767,20 @@ abstract class LogicValue {
       // ignore: avoid_dynamic_calls
       _doCompare(other, (a, b) => (a >= b) as bool);
 
+  /// Power operation
+  ///
+  /// Both inputs [a] and [b] are either of type [int] or [BigInt].
+  /// Returns [a] raise to the power [b] of same input type.
+  dynamic _powerOperation(dynamic a, dynamic b) {
+    if (a is BigInt) {
+      // ignore: avoid_dynamic_calls
```
mkorbel1, line 776: We can make some casts on `a` and `b` to highlight the assumption that either both are `BigInt` or both are `int`.  Then we can avoid these lint ignores.

lib/src/values/logic_value.dart
```diff
@@ -761,6 +767,20 @@ abstract class LogicValue {
...
+      return a.pow(b.toInt() as int);
```
mkorbel1, line 777: We should add an explicit exception if the value of `b` as a BigInt won't fit in an `int`.  I think this should work: https://api.flutter.dev/flutter/dart-core/BigInt/isValidInt.html.  Otherwise, we just truncate and silently do the wrong thing.  (even though it's unlikely someone needs more than 64 bits for their exponent intentionally).

mkorbel1 (changes requested):
Looking good!  Just a couple minor things

Sanchit-kumar:
Hi @mkorbel1 

There is a test case where  `LogicValue.ofBigInt(BigInt.zero, width)` for any `width > 64`  leads to throw an error in case of arithmetic `+,-,*` operations when applied to the same LogicValue. It throws an error **only when both are valued zero.** 

E.g., `LogicValue.ofBigInt(BigInt.zero, 128) + (LogicValue.ofBigInt(BigInt.zero, 128))` throws an error.

Can you please confirm if this is an issue or I misunderstood something. If it's an issue, we need to handle `_FilledLogicValue` with width>64 also for arithmetic operations.

mkorbel1:
> There is a test case where LogicValue.ofBigInt(BigInt.zero, width) for any width > 64 leads to throw an error in case of arithmetic +,-,* operations when applied to the same LogicValue. It throws an error only when both are valued zero.
> 
> E.g., LogicValue.ofBigInt(BigInt.zero, 128) + (LogicValue.ofBigInt(BigInt.zero, 128)) throws an error.
> 
> Can you please confirm if this is an issue or I misunderstood something. If it's an issue, we need to handle _FilledLogicValue with width>64 also for arithmetic operations.

@Sanchit-kumar, yes this looks like a bug, thanks for raising it!  I think this is a manifestation of #299 which is WIP being addressed in PR #319 (which has been WIP for a while, should probably poke on that).  Your test case that reproduces the issue is pretty simple, so I'll make a note on that issue to suggest using that as an additional test case.  Sorry you ran into this!

lib/src/values/logic_value.dart
```diff
@@ -7,6 +7,7 @@
 /// 2021 August 2
 /// Author: Max Korbel <max.korbel@intel.com>
 ///
+// ignore_for_file: prefer_single_quotes
```
mkorbel1, line 10: Let's not add this lint ignore, especially for the whole file.

If you're trying to split a line across multiple lines, you can do 
```dart
'''multi-line
string
'''
```
(i.e. it doesn't have to be double-quotes)

Also, you can just split strings like this, if you're just trying to make it fit on a line:
```dart
'this string needs'
' to be split across multiple lines'
' even though it is not a mult-line string'
```

You don't even need any `+`'s between them

lib/src/values/logic_value.dart
```diff
@@ -761,6 +769,29 @@ abstract class LogicValue {
...
+    if ((base is int) && (exponent is int)) {
+      return math.pow(base, exponent);
+    }
+    if ((base is BigInt) && (exponent is BigInt)) {
+      if (base == BigInt.one) {
+        return BigInt.one;
+      } else if (base == BigInt.zero && exponent > BigInt.zero) {
+        return BigInt.zero;
+      } else if (!exponent.isValidInt) {
+        throw Exception("""
```
mkorbel1, line 787: We are gradually trying to make ROHD throw more specific exceptions instead of `Exception`.  Can we add a new type of exception in the `exceptions/logic_value/` directory covering this flavor of exception?  It doesn't have to be super-specific to this exact function if you can come up with a more generic one (e.g. if you want to make one that a LogicValue operation couldn't be performed with a configurable message for why)

test/math_test.dart
```diff
@@ -79,6 +83,22 @@ void main() {
...
+        Vector({'a': 1043, 'b': 0}, {'a_power_b': 1}),
+        Vector({'a': 3, 'b': 3}, {'a_power_b': 27}),
+        Vector({'a': 2, 'b': 1}, {'a_power_b': 2}),
+        Vector({'a': 2, 'b': 7}, {'a_power_b': 128}),
+        Vector({'a': 2}, {'a_power_const': 32}),
+        Vector({'a': 5, 'b': 3}, {'a_power_b': 125}),
+        Vector({'a': 7, 'b': 2}, {'a_power_b': 49}),
+        Vector({'a': 0, 'b': 0}, {'a_power_b': 1}),
+        Vector({'a': 0, 'b': 3}, {'a_power_b': 0}),
+        Vector({'a': 1, 'b': 10000000}, {'a_power_b': 1}),
```
mkorbel1, line 98: The width of `b` is 8 bits, so this value is getting truncated down.  It doesn't really matter a lot since the base is 1, but just in case you were trying to test large numbers on it here.  I feel like it's not a big deal to change since we're covering the math part of it in `LogicValue` tests already, will leave to your judgement.

mkorbel1 (changes requested):
Changes look good!  A few more things

mkorbel1 (approved):

Issue #334, opened 2023-03-27
Add an enable signal to `FlipFlop`


mkorbel1:
### Motivation

Along the same lines as https://github.com/intel/rohd/issues/69

It would be nice to have an enable signal optionally provided as part of the built-in `FlipFlop`.

### Desired solution

Add an optional enable signal to `FlipFlop`.

gantanikhilraj:
Hello @mkorbel1 , could you brief me the issue a bit more?

mkorbel1:
Sure!  There's a `Module` called `FlipFlop` in `lib/src/modules/conditional.dart`.  The current constructor looks like this:
```dart
FlipFlop(Logic clk, Logic d, ...)
```

This represents a simple flip-flop sequential element which samples `d` on each rising edge and drives the last cycle's input on `q`.  An `enable` on a flip-flop is a way to control whether it should sample on the rising edge or continue to hold the previous value sampled when `enable` *was* high.

This enhancement is to add an optional `enable` signal (probably a nullable, optional named argument to the constructor) which then generates an enableable flip-flop.

Let me know if that helped clarify and if you'd like to take up the issue!

Sanchit-kumar:
I can take up this issue.

PR #357, opened 2023-05-07
Issue#334: Added enable signal to flipflop

Sanchit-kumar:
## Description & Motivation

It would be nice to have an enable signal optionally provided as part of the built-in FlipFlop as per #334 

## Testing

Update test/flop_test.dart for the same.

## Documentation

Update required at official [documentation](https://intel.github.io/rohd/rohd/FlipFlop-class.html). Not included by me.

### DIFF
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -1240,6 +1240,9 @@ ${padding}end ''');
 
 /// Represents a single flip-flop with no reset.
 class FlipFlop extends Module with CustomSystemVerilog {
+  /// Enable signal for flip-flop
+  late final String _enName;
+
   /// Name for the clk of this flop.
   late final String _clkName;
 
@@ -1253,19 +1256,30 @@ class FlipFlop extends Module with CustomSystemVerilog {
   late final Logic _clk = input(_clkName);
 
   /// The input to the flop.
+  late final Logic _en = input(_enName);
+
+  /// The input to the flop.
   late final Logic _d = input(_dName);
 
   /// The output of the flop.
   late final Logic q = output(_qName);
 
   /// Constructs a flip flop which is positive edge triggered on [clk].
-  FlipFlop(Logic clk, Logic d, {super.name = 'flipflop'}) {
+  FlipFlop(Logic clk, Logic d, {Logic? en, super.name = 'flipflop'}) {
     if (clk.width != 1) {
       throw Exception('clk must be 1 bit');
     }
+    if (en == null) {
+      en ??= Logic(name: 'en');
+      en.put(1);
+    }
+
+    _enName = Module.unpreferredName('en');
     _clkName = Module.unpreferredName('clk');
     _dName = Module.unpreferredName('d');
     _qName = Module.unpreferredName('q');
+
+    addInput(_enName, en);
     addInput(_clkName, clk);
     addInput(_dName, d, width: d.width);
     addOutput(_qName, width: d.width);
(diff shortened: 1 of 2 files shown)


lib/src/modules/conditional.dart
```diff
@@ -1274,18 +1288,21 @@ class FlipFlop extends Module with CustomSystemVerilog {
 
   /// Performs setup for custom functional behavior.
   void _setup() {
-    Sequential(_clk, [q < _d]);
+    Sequential(_clk, [
```
mkorbel1, line 1291: The design here is correct when `en` is provided, but we don't need to generate hardware that conditionally checks if `en` is high when it was _not_ provided.  Here, we want to conditionally generate the `If` around the conditional assignment only if `en` was provided.

This will make the ROHD simulation (slightly) more efficient by not needing to redundantly check the condition.  It will also make the generated outputs like SystemVerilog cleaner (even though the synthesis tools will probably optimize the if(1) away).

lib/src/modules/conditional.dart
```diff
@@ -1274,18 +1288,21 @@ class FlipFlop extends Module with CustomSystemVerilog {
...
-      throw Exception('FlipFlop has exactly two inputs and one output.');
+    if (inputs.length != 3 || outputs.length != 1) {
+      throw Exception('FlipFlop has exactly three inputs and one output.');
     }
     final clk = inputs[_clkName]!;
+    final en = inputs[_enName]!;
     final d = inputs[_dName]!;
     final q = outputs[_qName]!;
-    return 'always_ff @(posedge $clk) $q <= $d;  // $instanceName';
+    return 'always_ff @(posedge $clk) if($en) $q <= $d;  // $instanceName';
```
mkorbel1, line 1306: Here again, the SV generation should be dependent on whether `en` was provided

mkorbel1 (changes requested):
Thanks for contributing this!  This has been a requested feature from multiple users.

A few notes on the implementation, but the tests look great!

lib/src/modules/conditional.dart
```diff
@@ -1252,40 +1255,83 @@ class FlipFlop extends Module with CustomSystemVerilog {
...
 
+  /// To track if optional enable is provided or not.
+  late final bool _isEnableProvided;
+
   /// Constructs a flip flop which is positive edge triggered on [clk].
-  FlipFlop(Logic clk, Logic d, {super.name = 'flipflop'}) {
+  FlipFlop(Logic clk, Logic d, {Logic? en, super.name = 'flipflop'}) {
     if (clk.width != 1) {
       throw Exception('clk must be 1 bit');
     }
```
mkorbel1, line 1278: Probably want to add another check on `en.width` to make sure it's `1` if it's provided?

lib/src/modules/conditional.dart
```diff
@@ -1252,40 +1255,83 @@ class FlipFlop extends Module with CustomSystemVerilog {
...
   /// The input to the flop.
   late final Logic _d = input(_dName);
 
   /// The output of the flop.
   late final Logic q = output(_qName);
 
+  /// To track if optional enable is provided or not.
+  late final bool _isEnableProvided;
+
   /// Constructs a flip flop which is positive edge triggered on [clk].
```
mkorbel1, line 1274: We should discuss the impact of `en` in the doc comment of the constructor as well

mkorbel1 (changes requested):
Couple more tiny things, but looks great other than that!

Also, great job on testing!

mkorbel1 (approved):

Issue #337, opened 2023-03-31
Add clog2 functionality


mkorbel1:
### Motivation

In SystemVerilog, `$clog2` is a useful function.

It would be nice to have a synthesizable `Module` and associated functions for generating hardware and SystemVerilog with `$clog2`

### Desired solution

- Create a new `Module` that generates `$clog2` in output SystemVerilog and appropriately simulates in the ROHD simulator.
- Add functions (probably to `Logic`) to automatically generate this module in a convenient API.

mkorbel1:
Closed in #361

PR #361, opened 2023-05-14
Issue:#337 Added clog2

Sanchit-kumar:
## Description & Motivation

In SystemVerilog, $clog2 is a useful function. It is better to have clog2 as per #337 

## Testing

Added required test cases to 

> test/logic_value_test.dart 
> test/math_test.dart

## Documentation

Yes. I have update the following documentations:
> README.md
> doc/tutorials/chapter_2/00_basic_logic.md

### DIFF
--- a/lib/src/logic.dart
+++ b/lib/src/logic.dart
@@ -568,6 +568,9 @@ class Logic {
   /// Logical right-shift.
   Logic operator >>>(dynamic other) => RShift(this, other).out;
 
+  /// Unary clog2
+  Logic clog2() => Clog2Unary(this).out;
+
   /// Unary AND.
   Logic and() => AndUnary(this).out;
 
--- a/lib/src/logic.dart
+++ b/lib/src/logic.dart
@@ -568,9 +568,6 @@ class Logic {
   /// Logical right-shift.
   Logic operator >>>(dynamic other) => RShift(this, other).out;
 
-  /// Unary clog2
-  Logic clog2() => Clog2Unary(this).out;
-
   /// Unary AND.
   Logic and() => AndUnary(this).out;
 
--- a/lib/src/modules/gates.dart
+++ b/lib/src/modules/gates.dart
@@ -122,84 +122,6 @@ class _OneInputUnaryGate extends Module with InlineSystemVerilog {
   }
 }
 
-/// A generic one-input bitwise gate [Module].
-///
-/// It always takes one input and has one output. Output port have the
-/// same width as of input.
-abstract class _OneInputBitwiseGate extends Module with InlineSystemVerilog {
-  /// Name for a  input port of this module.
-  late final String _inName;
-
-  /// Name for the output port of this module.
-  late final String _outName;
-
-  /// An input to this gate.
-  late final Logic _in = input(_inName);
-
-  /// The output of this gate.
-  late final Logic out = output(_outName);
-
-  /// The output of this gate.
-  ///
-  /// Deprecated: use [out] instead.
-  @Deprecated('Use `out` instead.')
-  Logic get y => out;
-
-  /// The functional operation to perform for this gate.
-  final LogicValue Function(LogicValue in_) _op;
-
-  /// The `String` representing the operation to perform in generated code.
-  final String _opStr;
-
-  /// Constructs a one-input bitwise gate for an abitrary custom functional
-  /// implementation.
-  ///
-  /// The function [_op] is executed as the custom functional behavior.  When
-  /// this [Module] is in-lined as SystemVerilog, it will use [_opStr] as a
-  /// String of one input signal names (e.g. if [_opStr] was "clog2",
-  /// generated SystemVerilog may look like "$clog2(a)").
-  _OneInputBitwiseGate(this._op, this._opStr, Logic in_,
-      {String name = 'gate3'})
-      : super(name: name) {
-    _inName = Module.unpreferredName('in_${in_.name}');
-    _outName = Module.unpreferredName('${name}_${in_.name}');
-
-    addInput(_inName, in_, width: in_.width);
-    addOutput(_outName, width: in_.width);
-
-    _setup();
-  }
-
-  /// Performs setup steps for custom functional behavior.
-  void _setup() {
-    _execute(); // for initial values
-    _in.glitch.listen((args) {
-      _execute();
-    });
-  }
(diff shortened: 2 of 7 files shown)


lib/src/logic.dart
```diff
@@ -562,6 +562,9 @@ class Logic {
   /// Logical right-shift.
   Logic operator >>>(dynamic other) => RShift(this, other).out;
 
+  /// Unary clog2
```
mkorbel1, line 565: The doc comment here could be more helpful.  Somthing more like you had in the README.  Like "Ceiling of log base 2."  The fact that it's unary isn't so interesting like it is for `and()` where you may perform an and on one bus or two inputs.

lib/src/logic.dart
```diff
@@ -562,6 +562,9 @@ class Logic {
...
+  Logic clog2() => Clog2Unary(this).out;
```
mkorbel1, line 566: Upon doing some googling, it is unclear to me whether `$clog2` is a synthesizable function in SystemVerilog.  It seems legal for constant value inputs and for width calculations, but I don't see any dynamic signals fed into `$clog2` in a synthesizable design.

If it turns out to _not_ be synthesizable, then at the very least we need to document here that it should not be used for cases where it is desired to be synthesized, or else we need to generate synthesizable hardware that actually calculates the right value.  If we leave it as a non-synthesizable option, do we want to keep it in ROHD?  It could be useful for testbench-related module generation perhaps.

Thoughts?

lib/src/modules/gates.dart
```diff
@@ -122,6 +122,84 @@ class _OneInputUnaryGate extends Module with InlineSystemVerilog {
   }
 }
 
+/// A generic one-input bitwise gate [Module].
+///
+/// It always takes one input and has one output. Output port have the
+/// same width as of input.
+abstract class _OneInputBitwiseGate extends Module with InlineSystemVerilog {
```
mkorbel1, line 129: This doesn't really feel like it's a "bitwise" gate, right?  It's like a "function" gate or something, since it always adds `$` and `()`?

lib/src/values/logic_value.dart
```diff
@@ -679,6 +679,43 @@ abstract class LogicValue {
...
+      return LogicValue.ofInt(op(num) as int, width);
+    }
+  }
+
+  /// Ceil of log base 2 Operation
+  ///
+  /// Input [num] will be either of type [int] or [BigInt].
+  /// Returns ceil of log base 2 of [num]  having same type as of input.
+  dynamic _clog2(dynamic num) {
+    if (num is int) {
```
mkorbel1, line 711: Does this perform correctly for `int`s that can be interpreted as negative numbers?  Perhaps we need to cast to BigInt all the way up to `_INT_BITS-1` width?  Or else perform some other special consideration?

mkorbel1 (changes requested):

lib/src/logic.dart
Sanchit-kumar: Upon googling, I found out that clog2 is ‘synthesizable with limitations’.
Input to `$clog2` must be static-elaboration time constant and cannot be signal. Hence it cannot be used to model actual hardware. 
Reference: [https://www.csl.cornell.edu/courses/ece5745/handouts/ece5745-verilog-rules.pdf](url)

Considering these, maybe it's not required in ROHD. As a new user, it may be confusing to have clog2 option in ROHD which is actually non-synthesizable for signal. But if it is really required, we can keep by explicitly mentioning that it is synthesizable for constant value inputs only.

Let me know if we need to proceed with this PR.

lib/src/values/logic_value.dart
Sanchit-kumar: Yes, I did test a few values and cross checked online with $clog2 (with 64 bits) which is working correctly as per clog2 implemented. Actual $clog2 always converts the negative input to an equivalent positive number using the standard way (2^bits + (input negative number)). Hence input is always a positive equivalent.

The implemented clog2 does the same.

E.g., `LogicValue.ofInt(-128, 8).clog2()` outputs 7
         `LogicValue.ofInt(-127, 8).clog2()` outputs 8

Works similarly for BigInt.

I will add negative cases also if we need to proceed with PR.

lib/src/logic.dart
mkorbel1: Thanks for the link, that seems like a good reference.

Since there's not a legal way to pass a signal to it, perhaps it's not appropriate to have a `Module` that accepts `Logic`s to compute it.  However, the functionality still seems valuable for `LogicValue` to have.  So I think it's worth proceeding with the PR, but unfortunately I think we should omit the part that applies to synthesizable hardware generation (`Module` and `Logic` stuff) and just keep the `LogicValue` addition.  How does that sound to you?

lib/src/values/logic_value.dart
mkorbel1: The case I was thinking of was negative numbers which are at the int-width boundary, since that's the point where they get interpreted as negative by the int libraries that use `bitLength`.  For example, it seem this test is failing:
```dart
expect(LogicValue.ofInt(-1, 64).clog2().toInt(), 64);
```

lib/src/logic.dart
Sanchit-kumar: Yes, it would better to proceed with the PR without synthesizable `Module` that accepts `Logic`s.

lib/src/values/logic_value.dart
```diff
@@ -678,6 +678,45 @@ abstract class LogicValue {
...
+  /// Ceil of log base 2 operation.
+  ///
+  /// Returns ceil of log base 2 of valid input.
+  /// Returns `x` if any bit of input is invalid.
+  LogicValue clog2() => _doUnaryMath(_clog2);
+
+  /// Executes mathematical operations on single [LogicValue]
+  ///
+  /// Handles invalid input and do proper type conversion to required types
+  LogicValue _doUnaryMath(dynamic Function(dynamic num, int width) op) {
```
mkorbel1, line 690: `num` is a built-in type in Dart, so it's a little confusing to have variables throughout named `num`.  Can we try a different variable name?

mkorbel1 (changes requested):
Getting close, this looks great, thank you!  Very nice test suite here too

mkorbel1 (approved):

PR #362, opened 2023-05-15
232 add a mechanism for generating random logic values

### LINKED ISSUE #232: Add a mechanism for generating random `LogicValue`s
### Motivation

`LogicValue` is a useful type since it supports arbitrary width in an efficient representation.  For testing purposes, it might be desirable to generate a random value on a bus of arbitrary width.  The built-in Dart randomization only randomizes up to 2^32, which means users have to invent their own little code snippets to generate random numbers bigger than 32 bits.

### Desired solution

A way to generate large random numbers, especially for `LogicValue`.

One nice option might be to add an `extension` on `Random`.

It might be nice to support both valid values or a mix of valid and invalid values.

It might be nice to support random `BigInt`s as well.

### Alternatives considered

_No response_

### Additional details

_No response_

quekyj:
## Description & Motivation

LogicValue is a useful type since it supports arbitrary width in an efficient representation. For testing purposes, it might be desirable to generate a random value on a bus of arbitrary width. The built-in Dart randomization only randomizes up to 2^32, which means users have to invent their own little code snippets to generate random numbers bigger than 32 bits.

## Related Issue(s)

Fix #232 

## Testing

Add unit test to test random generation of the `LogicValue`.

## Backwards-compatibility

No

## Documentation

Included in the public documentation.


lib/src/values/values.dart
```diff
@@ -11,3 +13,61 @@ part 'logic_value.dart';
 part 'small_logic_value.dart';
 part 'big_logic_value.dart';
 part 'filled_logic_value.dart';
+
+/// Allows random generation of [LogicValue] for [BigInt] and [int].
```
mkorbel1, line 17: Let's move all the random stuff into a new file

lib/src/values/values.dart
```diff
@@ -11,3 +13,61 @@ part 'logic_value.dart';
...
+extension RandLogicValue on Random {
+  /// Generate non-negative random [BigInt] value that consists of
```
mkorbel1, line 19: I think more interesting than "non-negative" is "unsigned" (in all comments where you mention this)

lib/src/values/values.dart
```diff
@@ -11,3 +13,61 @@ part 'logic_value.dart';
...
+    var result = BigInt.zero;
+    for (var i = 0; i < numBits; i += 32) {
+      result = (result << 32) | BigInt.from(nextInt(1 << 32));
+    }
+    return result & ((BigInt.one << numBits) - BigInt.one);
+  }
+
+  /// Generate non-negative random [LogicValue] based on [width] and [max] num.
+  /// The random number can be mixed in invalid bits x and z by set
+  /// [hasInvalidBits] to `false`.
```
mkorbel1, line 38: This doc seems backwards?  Setting it to `true` will add invalid bits, right?

lib/src/values/values.dart
```diff
@@ -11,3 +13,61 @@ part 'logic_value.dart';
...
+  /// Example:
+  ///
+  /// ```dart
+  /// // generate 100 bits of random BigInt
+  /// final bigInt = Random(10).nextBigInt(numBits: 100);
+  /// ```
+  LogicValue nextLogicValue({
+    required int width,
+    int? max,
+    bool hasInvalidBits = false,
```
mkorbel1, line 49: I think it might be better to call this `includeInvalidBits` since we're including it, it doesn't already have it?

lib/src/values/values.dart
```diff
@@ -11,3 +13,61 @@ part 'logic_value.dart';
...
+  }) {
+    final bitString = StringBuffer();
```
mkorbel1, line 51: define the `bitString` inside the `if` statement where it's used since we don't need it elsewhere.

lib/src/values/values.dart
```diff
@@ -11,3 +13,61 @@ part 'logic_value.dart';
...
+      for (var i = 0; i < width; i++) {
+        bitString.write(const ['1', '0', 'x', 'z'][nextInt(4)]);
+      }
+
+      return LogicValue.ofString(bitString.toString());
+    } else {
+      // Generate the random value of range between min and max
+      // that are still within width
+
+      // 2 ^ 5 = 32 = 100000 = 1 << 5
```
mkorbel1, line 63: Remove this comment

lib/src/values/values.dart
```diff
@@ -11,3 +13,61 @@ part 'logic_value.dart';
...
+      // 32 - 1 = 31 =011111
+
+      if (width <= LogicValue._INT_BITS) {
```
mkorbel1, line 66: We're not considering `max` here.  I think a simple modulo should be sufficient?
e.g. (if `max` is provided)
```dart
LogicValue.ofInt(nextInt(1 << width), width) % max;
```

lib/src/values/values.dart
```diff
@@ -11,3 +13,61 @@ part 'logic_value.dart';
...
+  /// ```
+  BigInt _nextBigInt({required int numBits}) {
+    var result = BigInt.zero;
+    for (var i = 0; i < numBits; i += 32) {
+      result = (result << 32) | BigInt.from(nextInt(1 << 32));
+    }
+    return result & ((BigInt.one << numBits) - BigInt.one);
+  }
+
+  /// Generate non-negative random [LogicValue] based on [width] and [max] num.
```
mkorbel1, line 36: Need to document that `max` has no impact when there are invalid bits

test/logic_value_test.dart
```diff
@@ -947,4 +949,44 @@ void main() {
           equals(LogicValue.filled(0, LogicValue.zero).hashCode));
     });
   });
+
+  group('random value generation ', () {
+    test(
+        'should return random integer logic value if min and max value '
+        'is in integer', () {
+      final lvInt = Random(10).nextLogicValueInt(min: 20, max: 100);
```
mkorbel1, line 957: does this function exist?

test/logic_value_test.dart
```diff
@@ -947,4 +949,44 @@ void main() {
...
+      print(lvRand);
+    });
+
+    test(
+        'should return random logic value without invalid bits '
+        'if hasInvalidBits is false', () {
+      final lvRand = Random().nextLogicValue(width: 10, max: 20);
+      print(lvRand);
+    });
+  });
```
mkorbel1, line 991: add tests that exercise different number of bits (0, 1-31, 32, 31-63, 64, >64) to hit edge cases.

mkorbel1, line 991: Add tests that `max` works properly

mkorbel1, line 991: Add tests that invalid bits get added properly

test/logic_value_test.dart
```diff
@@ -947,4 +949,44 @@ void main() {
...
+      final lvBigInt = Random(10).nextLogicValueBigInt(numBits: 100);
+
+      expect(lvBigInt, isA<LogicValue>());
+      expect(lvBigInt.toBigInt(), isA<BigInt>());
+      expect(lvBigInt.width, 100);
+    });
+    test(
+        'should return random logic value with invalid bits '
+        'if hasInvalidBits is true', () {
+      final lvRand = Random().nextLogicValue(width: 10, hasInvalidBits: true);
```
mkorbel1, line 980: Use fixed seeds in tests with randomness

mkorbel1 (changes requested):

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,60 @@
...
+    var result = BigInt.zero;
+    for (var i = 0; i < numBits; i += 32) {
+      result = (result << 32) | BigInt.from(nextInt(1 << 32));
+    }
+    return result & ((BigInt.one << numBits) - BigInt.one);
+  }
+
+  /// Generate unsigned random [LogicValue] based on [width] and [max] num.
+  /// The random number can be mixed in invalid bits x and z by set
+  /// [includeInvalidBits] to `true`. [max] only work when [includeInvalidBits]
```
mkorbel1, line 27: Why can't we use `max` when it's bigger than 64 bits?  `BigInt` supports modulo too (and add a test for it)

mkorbel1 (changes requested):

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,79 @@
...
+    LogicValue? max,
+    bool includeInvalidBits = false,
+  }) {
+    if (width == 0) {
+      throw RangeError('width size must be larger than 0.');
+    }
+
+    if (includeInvalidBits) {
+      if (max != null) {
+        throw Exception(
```
mkorbel1, line 40: Let's throw a ROHD-specific exception

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,79 @@
...
+  /// The random number can be mixed in invalid bits x and z by set
+  /// [includeInvalidBits] to `true`. [max] only work when [includeInvalidBits]
+  /// is set to false else an exception will be thrown.
+  LogicValue nextLogicValue({
+    required int width,
+    LogicValue? max,
+    bool includeInvalidBits = false,
+  }) {
+    if (width == 0) {
+      throw RangeError('width size must be larger than 0.');
```
mkorbel1, line 35: Can we just return a `LogicValue` of width 0 instead of throwing an exception?

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,79 @@
...
+      final bitString = StringBuffer();
+      for (var i = 0; i < width; i++) {
+        bitString.write(const ['1', '0', 'x', 'z'][nextInt(4)]);
+      }
+
+      return LogicValue.ofString(bitString.toString());
+    } else {
+      if (width <= LogicValue._INT_BITS) {
+        LogicValue ranNum;
+        if (width < 32) {
```
mkorbel1, line 53: shouldn't this be `<= 32`?

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,79 @@
...
+          ranNum = LogicValue.ofInt(nextInt(1 << width), width);
+        } else {
+          ranNum = LogicValue.ofInt(_nextBigInt(numBits: width).toInt(), width);
+        }
+
+        return LogicValue.ofInt(
```
mkorbel1, line 59: If `max` is null, we don't need to convert it to an `int` and back again.  Let's move that ternary up a level, so that if `max` is null we just return `ranNum` directly.

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,79 @@
...
+    return result & ((BigInt.one << numBits) - BigInt.one);
+  }
+
+  /// Generate unsigned random [LogicValue] based on [width] and [max] num.
+  /// The random number can be mixed in invalid bits x and z by set
+  /// [includeInvalidBits] to `true`. [max] only work when [includeInvalidBits]
+  /// is set to false else an exception will be thrown.
+  LogicValue nextLogicValue({
+    required int width,
+    LogicValue? max,
```
mkorbel1, line 31: It might make more sense to make `max` a `dynamic` that supports either `int` or `BigInt`?  In which case, need to check that it's not negative.  Otherwise, if we keep `LogicValue`, we should check that it's valid.

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,79 @@
...
+      if (width <= LogicValue._INT_BITS) {
+        LogicValue ranNum;
+        if (width < 32) {
+          ranNum = LogicValue.ofInt(nextInt(1 << width), width);
+        } else {
+          ranNum = LogicValue.ofInt(_nextBigInt(numBits: width).toInt(), width);
+        }
+
+        return LogicValue.ofInt(
+            ranNum.toInt() % (max == null ? (1 << width) - 1 : max.toInt()),
```
mkorbel1, line 60: if `max` is `0`, then doesn't the `%` cause an error?  We should add tests for `max=0` in both `int` and `BigInt` widths.

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,79 @@
...
+      } else {
+        var ranNum = _nextBigInt(numBits: width);
+
+        if (max == null || ranNum.bitLength < max.width) {
+          return LogicValue.ofBigInt(ranNum, width);
+        } else {
+          final minWidth =
+              max.width < ranNum.bitLength ? max.width : ranNum.bitLength;
+
+          ranNum = (ranNum & ((BigInt.one << minWidth) - BigInt.one)) %
```
mkorbel1, line 71: I don't think the masking is necessary here (or in the `int` case) since `LogicValue` will already truncate to `width`?

mkorbel1 (changes requested):

lib/src/values/random_logic_value.dart
```diff
@@ -50,28 +59,31 @@ extension RandLogicValue on math.Random {
...
         }
 
-        return LogicValue.ofInt(
-            ranNum.toInt() % (max == null ? (1 << width) - 1 : max.toInt()),
-            width);
+        if (max == null) {
+          return ranNum;
+        } else {
+          if (max is BigInt) {
+            throw InvalidRandomLogicValueException(
```
mkorbel1, line 72: I don't think we have to throw an exception here.  Just because it's a bigint doesn't mean we cant use it, right?

lib/src/values/random_logic_value.dart
```diff
@@ -50,28 +59,31 @@ extension RandLogicValue on math.Random {
...
+          }
+          return LogicValue.ofInt(ranNum.toInt() % max, width);
+        }
       } else {
-        var ranNum = _nextBigInt(numBits: width);
+        final ranNum = _nextBigInt(numBits: width);
 
-        if (max == null || ranNum.bitLength < max.width) {
+        if (max == null ||
+            (max is BigInt && ranNum.bitLength < max.bitLength)) {
```
mkorbel1, line 83: Maybe we need to be careful with `bitLength`, since I think negative numbers act differently than positive numbers (e.g. -1 bitlength is 1 i think, even though it's 0xffffffff)?  I wonder if we should try to make some mocks or something to test specific outputs from the random generator?

lib/src/values/random_logic_value.dart
```diff
@@ -50,28 +59,31 @@ extension RandLogicValue on math.Random {
...
           return LogicValue.ofBigInt(ranNum, width);
         } else {
```
mkorbel1, line 85: what if `max` is `int`?

test/logic_value_test.dart
```diff
@@ -1033,10 +1034,25 @@ void main() {
       expect(lvRand.toString(), contains('x'));
     });
 
-    test('should throw RangeError Exception if width is 0', () {
-      expect(() {
-        Random(5).nextLogicValue(width: 0);
-      }, throwsA((dynamic e) => e is RangeError));
+    test('should return empty LogicValue if width is 0.', () {
+      expect(Random(5).nextLogicValue(width: 0).toInt(), equals(0));
```
mkorbel1, line 1038: instead of `toInt() == 0`, let's check that it equals `LogicValue.empty`?

mkorbel1 (changes requested):

lib/src/values/random_logic_value.dart
quekyj: Ya, integer -1 bitLength is 1. But since the `Random()` generator only generates positive numbers, I think it should be fine right? Just need to add a check that thrown exception when max is negative.

https://api.flutter.dev/flutter/dart-math/Random-class.html

test/logic_value_test.dart
```diff
@@ -1022,4 +1025,97 @@ void main() {
...
+            .nextLogicValue(width: i, max: BigInt.parse('9999999999999999999'));
+        final randMaxInt =
+            Random(5).nextLogicValue(width: i, max: BigInt.from(30));
+        expect(randMaxBigInt, equals(Random(5).nextLogicValue(width: i)));
+        expect(randMaxInt.toInt(), lessThan(30));
+      }
+    });
+    test(
+        'should return valid results when using int as max '
+        'and BigInt width.', () {
```
mkorbel1, line 1077: what about `int` max when width <= 64?

mkorbel1 (changes requested):

test/logic_value_test.dart
```diff
@@ -1022,4 +1025,97 @@ void main() {
...
+    test('should return empty LogicValue when max is 0 for int and big int.',
+        () {
+      final maxBigInt = BigInt.zero;
+      const maxInt = 0;
+      expect(Random(5).nextLogicValue(width: 10, max: maxInt).toInt(),
+          equals(maxInt));
+
+      expect(Random(5).nextLogicValue(width: 80, max: maxBigInt).toBigInt(),
+          equals(maxBigInt));
+    });
```
mkorbel1, line 1092: can you add some spacing between `test`s?

lib/src/values/random_logic_value.dart
```diff
@@ -0,0 +1,93 @@
...
+      if (width <= LogicValue._INT_BITS) {
+        final ranNum = width <= 32
+            ? LogicValue.ofInt(nextInt(1 << width), width)
+            : LogicValue.ofInt(_nextBigInt(numBits: width).toInt(), width);
+
+        if (max == null) {
+          return ranNum;
+        } else {
+          final maxBigInt = max is int ? BigInt.from(max) : max as BigInt;
+          return LogicValue.ofInt(ranNum.toInt() % maxBigInt.toInt(), width);
```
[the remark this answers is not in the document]
quekyj: Its seems like the output will be converted to `1e+41` when attempt to convert `BigInt` to `Int`. Probably I can just return the ranNum when max is BigInt and the width only support `int`.

test/logic_value_test.dart
```diff
@@ -1022,4 +1025,97 @@ void main() {
...
+      final lvRand =
+          Random(5).nextLogicValue(width: 10, includeInvalidBits: true);
+
+      expect(lvRand.toString(), contains('x'));
+    });
+    test('should return empty LogicValue when width is 0.', () {
+      expect(Random(5).nextLogicValue(width: 0), equals(LogicValue.empty));
+    });
+    test(
+        'should return valid results when using BigInt as max '
```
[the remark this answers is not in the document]
quekyj:
```dart
test(
        'should return random big integer logic value with width '
        'greater than 64 when having different width and max constraint.', () {}
```
This is tested in another test. Probably I should merge it together

test/logic_value_test.dart
quekyj: Ya, its in here:

```dart
    test(
        'should return random int logic value without invalid bits when'
        ' having different width and max constraint.', () {
```

Merged to avoid confusion.

mkorbel1 (approved):
Looks great!  Nice work!

Issue #316, opened 2023-03-15
Prevent `Logic` from connecting to itself


chykon:
### Describe the bug

`Logic` can be connected to itself, which is rather odd.

### To Reproduce

```dart
import 'package:rohd/rohd.dart';

void main() {
  final a = Logic();
  final b = Logic();

  a <= a;
  a <= b; // exception
}
```

### Expected behavior

Connecting `Logic` to itself is not allowed.

### Actual behavior

Connecting `Logic` to itself is allowed.

### Additional: Context

ROHD v0.4.2

mkorbel1:
This represents a trivial case of a combinational loop.  It should be quite easy to detect statically, especially given `_Wire`s are shared among connected `Logic`s.  I am having trouble thinking of any valid use-case where a ring of wires connected directly to each other (including one to itself) would ever be valuable, so throwing an exception immediately seems fair to me.

Thank you for filing!

Issue #315, opened 2023-03-15
Add more informative looped connection error message


chykon:
### Motivation

Might be worth adding a more descriptive looped connection error message. Currently showing `Unhandled exception: LateInitializationError: Field '_preTickSubscription@22087482' has already been initialized.`.

Code:

```dart
import 'package:rohd/rohd.dart';

void main() {
  final a = Logic();
  final b = Logic();

  a <= b;
  b <= a;
}
```

### Desired solution

Add more informative looped connection error message.

PR #368, opened 2023-05-20
Issue#316 Prevent self connecting logic

Sanchit-kumar:
## Description & Motivation
Prevent Logic from connecting to itself #316 

## Related Issue(s)
#315 (Covered)

## Testing

Added testcase to `test/bus_test.dart` and created an exception `lib/src/exceptions/logic/invalid_connection_exception.dart`

## Documentation

Not required.


lib/src/exceptions/logic/invalid_connection_exception.dart
```diff
@@ -0,0 +1,20 @@
...
+///
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd/src/exceptions/rohd_exception.dart';
+
+/// An exception that thrown when a [Logic] is trying for an invalid connection.
+class InvalidConnectionException extends RohdException {
+  /// Creates an exception when a logic is trying for an invalid connect.
+  InvalidConnectionException(Logic a, Logic b)
+      : super('Signal "$a" trying to connect with signal "$b"'
```
mkorbel1, line 18: This error message is true, but might be confusing to someone unintentionally connecting a signal to itself.  Can we either make this message more informative/specific that it's connecting a signal to itself?  Or else add an additional message that can be provided with details for the reason?

lib/src/logic.dart
```diff
@@ -493,6 +493,9 @@ class Logic {
 
   /// Handles the actual connection of this [Logic] to [other].
   void _connect(Logic other) {
+    if (this == other) {
+      throw InvalidConnectionException(this, other);
+    }
     _unassignable = true;
     _updateWire(other._wire);
```
mkorbel1, line 500: I think we should be able to also/instead check that `this._wire` is the same as `other._wire`.  `Logic`s that are connected without any gates in between should share an underlying `Wire`.  This way we can detect far-off `Logic`s that are actually the same wire connected to itself as well.

test/bus_test.dart
```diff
@@ -269,6 +269,17 @@ void main() {
...
+
+    group('invalid connection exceptions', () {
+      test('self-connecting logic', () {
+        try {
+          final a = Logic();
+          a <= a;
+        } on Exception catch (e) {
+          expect(e.runtimeType, InvalidConnectionException);
+        }
+      });
```
mkorbel1, line 281: We can add a test where `a <= b` and `b <= c` and `c <= a` with the `Wire` updates and it should be able to detect that too.

mkorbel1 (changes requested):
Thank you for your contribution!

It looks like this is addressing both #315, #316?  In what way do you think #315 isn't?

lib/src/logic.dart
Sanchit-kumar: Yes, it will work. Got the clarity.

mkorbel1 (approved):

Issue #371, opened 2023-05-30
Make `LogicValue` comparable


mkorbel1:
### Motivation

Calling `sort` on a `List<LogicValue>` results in an error since they are not comparable.  We should implement functionality in `LogicValue` so that they can be naturally compared.

If the value is not valid, we should probably just throw an exception (like we do for other number-based operations).

### Desired solution

Make `LogicValue` implement [Comparable](https://api.dart.dev/stable/3.0.2/dart-core/Comparable-class.html)

Throw exception if not valid.

Consider width properly, since we need to use `int` vs `BigInt` depending on size.

### Alternatives considered

Current workaround is to sort based on `toInt()` or `toBigInt()`

mkorbel1:
Closed in #373

PR #373, opened 2023-06-05
Issue#371 Made LogicValue Comparable

Sanchit-kumar:
## Description & Motivation

Calling `sort` on a `List<LogicValue>` results in an error since they are not comparable. We should implement functionality in `LogicValue` so that they can be naturally compared as per #371 

## Testing

Added required test cases to file `test/logic_value_test.dart`

## Documentation

Not included by me on `README.md`. Need list this addition functionality.


test/logic_value_test.dart
```diff
@@ -1022,4 +1022,201 @@ void main() {
...
+        expected[i] = values[i].toInt();
+      }
+
+      values.sort();
+      expected.sort();
+      for (var i = 0; i < values.length; i++) {
+        expect(values[i].toInt(), expected[i]);
+      }
+    });
+    test('signed values - int 64 bits', () {
```
mkorbel1, line 1051: good test, but wrong result.  the `LogicValue`s should always compare as unsigned numbers.

test/logic_value_test.dart
```diff
@@ -1022,4 +1022,201 @@ void main() {
...
+      final h64 = LogicValue.ofInt(-1, 64);
+      final invalidLogicX = LogicValue.filled(8, LogicValue.x);
+      final invalidLogicZ = LogicValue.filled(8, LogicValue.z);
+
+      <LogicValue>[a64, b64, c64, d64, e64, f64, g64, h64].sort();
+
+      final excBigIntWidth = <LogicValue>[a64, b64, a128];
+      final excIntWidth = <LogicValue>[e64, e8, f64];
+
+      try {
```
mkorbel1, line 1198: These try/catches are good if an exception is caught, but they dont test that the exception is actually being thrown.  If none of them threw any exception the test would still pass.

mkorbel1 (changes requested):
Thanks for taking this up!

test/logic_value_test.dart
```diff
@@ -1252,4 +1252,190 @@ void main() {
...
+
+      for (var i = 0; i < values.length; i++) {
+        expect(BigInt.from(values[i].toInt()).toUnsigned(values[i].width),
+            expected[i]);
+      }
+    });
+    test('unsigned  BigInt(invalid int)  & int ', () {
+      final a = LogicValue.ofBigInt(BigInt.parse('3'), 64);
+      final b = LogicValue.ofBigInt(BigInt.zero, 64);
+      final c = LogicValue.ofBigInt(BigInt.from(-1), 64); //Invalid Int
```
mkorbel1, line 1336: what makes this "invalid"?

mkorbel1 (changes requested):

lib/src/values/logic_value.dart
```diff
@@ -565,6 +565,58 @@ abstract class LogicValue {
...
+  ///
+  /// Returns a negative number if `this` is less than `other`, zero if they are
+  /// equal, and a positive number if `this` is greater than `other`.
+  ///
+  /// It will throw an exception if `this` or [other] is not [isValid] or for
+  /// non-equal `LogicValue` [width].
+  @override
+  int compareTo(Object other) {
+    if (other is! LogicValue) {
+      throw TypeError();
```
mkorbel1, line 578: Perhaps we should extend this or something, or go back to an exception, because this has no helpful message

Sanchit-kumar: Adding `ValueTypeError` to the exceptions folder by extending `TypeError` looks like a good option but `ValueType`Error may confuse the developers. So going back to Exception() seems a good option, but let me know if `ValueTypeError() ` with message serves better in this case.

test/logic_value_test.dart
Sanchit-kumar: Since `-1 BigInt` of 64 bits is an `invalid` int and we are comparing `BigInt` and `int`, so
 I kept sanity checks as I was accepting the signed number input in the `-2^(n-1) to 2^(n-1)-1` range, but processing it like an unsigned int.

Since we are dealing only with unsigned values, maybe it is redundant.

mkorbel1 (approved):
Looks good to me, thank you for the contribution!

PR #380, opened 2023-06-09
Chapter 7 bootcamp

quekyj:
## Description & Motivation

Chapter 7 of bootcamp material - Sequential Logic. 

## Related Issue(s)

None

## Testing

None

## Backwards-compatibility

Yes

## Documentation

Nope


doc/tutorials/chapter_7/00_sequential_logic.md
```diff
@@ -0,0 +1,291 @@
...
+
+## Learning Outcome
+
+In this chapter:
+
+- You will learn how to create a sequential logic that are equivalent to System Verilog `always_ff`. We will also review on what is `Simulator` and how to create unit test along with ROHD Simulator.
+
+## What is Sequential Logic?
+
+Sequential Logic is a circuit that consists of flip-flops as memory to generate output. In our previous chapter, we talked about combinational logic which input is only depend on the current state. But in sequential logic, past input will also be taken account before generate the current output.
```
mkorbel1, line 19: > Sequential Logic is a circuit that consists of flip-flops as memory to generate output.

Sequential logic consists of elements that store state (such as flip-flops).

doc/tutorials/chapter_7/00_sequential_logic.md
```diff
@@ -0,0 +1,291 @@
...
+void main() async {
+  final shiftReg = ShiftRegister();
+  await shiftReg.build();
+  print(shiftReg.generateSynth());
+}
+```
+
+Next, let define our inputs to the shift register. So, in our shift register we will need a reset pin `reset`, shift in pin `sin` and a clock `clk`. As for the output, there is one output pin shift out `sout`.
+
+Let register or add the inputs and output to our ShiftRegister module. Oh ya, we can also add the module name for our module as well. You can doing so by adding `super.name='shift_register'` to the constructor.
```
mkorbel1, line 53: > Oh ya, we can also add the module name for our module as well. You can doing so by adding `super.name='shift_register'` to the constructor.

Haha funny, but maybe not the right spot for this tidbit.  Maybe this should be in the modules section?

doc/tutorials/chapter_7/00_sequential_logic.md
```diff
@@ -0,0 +1,291 @@
...
+Simulator.setMaxSimTime(100); // set maximum time to 100
+```
+
+- To immediately end the simulation at the end of the current timestamp, use `Simulator.endSimulation`
+
+```dart
+Simulator.registerAction(75, Simulator.endSimulation); 
+```
+
+- To run just the next timestamp, use `Simulator.tick`
```
mkorbel1, line 149: Let's dissuade users from calling `Simulator.tick` directly since it can be a little confusing.  There can be multiple `tick`s in one timestamp.  Just don't mention `Simulator.tick`.

doc/tutorials/chapter_7/00_sequential_logic.md
```diff
@@ -0,0 +1,291 @@
...
+    injectedActionExecuted = true;
+  });
+});
+```
+
+## Unit Test in Sequential Logic
+
+Let see how we can actually build a unit test in `Sequential`. Before we start the simulation, let inject value of 1 to signals `reset` and `sin` to prevent our signal from being `z` value at the start.
+
+We also can create a local function that print the flop of the clock. We can access the Simulator time using `.time()` function. Then, let us kick start the `Simulator` by setting its maximum simulation time and `.run()` the Simulator. Notice that we use `unawaited()` function instead of `await` because we want to do something with the positive edges.
```
mkorbel1, line 189: `time` is a getter, not `time()`

mkorbel1, line 189: It might be good to explain the `unawaited ` little more, that it's telling Dart not to wait for it to complete before continuing.

mkorbel1 (changes requested):
Nice job

doc/tutorials/chapter_7/00_sequential_logic.md
quekyj: Ya, let me pull this out for now.

mkorbel1 (approved):

PR #386, opened 2023-06-16
Cases

### LINKED ISSUE #304: Implement a shorthand for case statements
### Motivation

Sometimes it's desirable to assign/create some signal based on some condition, similar to a `mux`, but there is more than one case to consider.  The current easy options could be either a chain of `mux`s, which implies priority, or building a `Case` in a `Conditional`, which is a little heavy-weight for a simple scenario.

It would be nice to have some automation that creates a `Conditional` and `Case`, optionally with the usual set of settings like `unique`.

### Desired solution

A function which builds a `Conditional` and `Case` statement with some simple set of inputs.  Probably it should only be capable of applying to a single signal (i.e. one output, like `mux`).  For more complex scenarios, people can just create the full thing themselves.

### Alternatives considered

_No response_

### Additional details

_No response_

mjayasim9:
## Description & Motivation

Implemented a shorthand for creating a Conditional and Case. This is for the scenario where it is needed to assign/create a signal based on a condition, similar to a mux. The output is a single `Logic` signal. The width of the output signal is optional and if not specified, is inferred from the width of the conditions.

## Related Issue(s)

Fix #304

## Testing

Added 5 new testcases which test the expression and condition match as well as width mismatch exceptions.

## Backwards-compatibility

No

## Documentation

Yes, API docs

### DIFF
--- a/lib/src/exceptions/logic/signal_width_mismatch_exception.dart
+++ b/lib/src/exceptions/logic/signal_width_mismatch_exception.dart
@@ -18,4 +18,12 @@ class SignalWidthMismatchException extends RohdException {
       : super('Signal ${signal.name} has the wrong width.'
             ' Expected $expectedWidth but found ${signal.width}.'
             ' $additionalMessage');
+
+  /// Constructs a new [Exception] for when a dynamic has a wrong width.
+  SignalWidthMismatchException.forDynamic(
+      dynamic val, int expectedWidth, int actualWidth,
+      {String additionalMessage = ''})
+      : super('Value $val has the wrong width.'
+            ' Excepted $expectedWidth but found $actualWidth.'
+            ' $additionalMessage');
 }
--- a/lib/src/exceptions/logic/signal_width_mismatch_exception.dart
+++ b/lib/src/exceptions/logic/signal_width_mismatch_exception.dart
@@ -24,6 +24,6 @@ class SignalWidthMismatchException extends RohdException {
       dynamic val, int expectedWidth, int actualWidth,
       {String additionalMessage = ''})
       : super('Value $val has the wrong width.'
-            ' Excepted $expectedWidth but found $actualWidth.'
+            ' Expected $expectedWidth but found $actualWidth.'
             ' $additionalMessage');
 }
--- a/lib/src/exceptions/logic/signal_width_mismatch_exception.dart
+++ b/lib/src/exceptions/logic/signal_width_mismatch_exception.dart
@@ -30,6 +30,6 @@ class SignalWidthMismatchException extends RohdException {
   /// Constructs a new [Exception] for when a dynamic has no width or it could
   /// not be inferred.
   SignalWidthMismatchException.forNull(dynamic val)
-      : super('Could not infer width of value $val.'
+      : super('Could not infer width of input $val.'
             ' Please provide a valid width.');
 }
--- a/lib/src/exceptions/logic/signal_width_mismatch_exception.dart
+++ b/lib/src/exceptions/logic/signal_width_mismatch_exception.dart
@@ -26,4 +26,10 @@ class SignalWidthMismatchException extends RohdException {
       : super('Value $val has the wrong width.'
             ' Expected $expectedWidth but found $actualWidth.'
             ' $additionalMessage');
+
+  /// Constructs a new [Exception] for when a dynamic has no width or it could
+  /// not be inferred.
+  SignalWidthMismatchException.forNull(dynamic val)
+      : super('Could not infer width of value $val.'
+            ' Please provide a valid width.');
 }
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -908,14 +908,10 @@ Logic cases(Logic expression, Map<Logic, dynamic> conditions, {int? width}) {
     width ??= inferredWidth;
 
     if (width != inferredWidth && inferredWidth != null) {
-      //throw Exception('Width of (${condition.value}) '
-      //    'must match width ($width)');
       throw SignalWidthMismatchException.forDynamic(
           condition.value, width!, inferredWidth);
     }
     if (expression.width != condition.key.width) {
-      //  throw Exception('Width of (${condition.key}) '
-      //      'must match with width of ($expression)');
       throw SignalWidthMismatchException.forDynamic(
           condition.key, expression.width, condition.key.width);
     }
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -933,8 +933,6 @@ Logic cases(Logic expression, Map<dynamic, dynamic> conditions,
     }
   }
 
-  if (defaultValue == null) {}
-
   for (final condition in conditions.entries) {
     if (condition.key is Logic) {
       if (expression.width != (condition.key as Logic).width) {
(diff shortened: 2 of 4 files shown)


lib/src/exceptions/logic/signal_width_mismatch_exception.dart
```diff
@@ -18,4 +18,12 @@ class SignalWidthMismatchException extends RohdException {
       : super('Signal ${signal.name} has the wrong width.'
             ' Expected $expectedWidth but found ${signal.width}.'
             ' $additionalMessage');
+
+  /// Constructs a new [Exception] for when a dynamic has a wrong width.
+  SignalWidthMismatchException.forDynamic(
+      dynamic val, int expectedWidth, int actualWidth,
+      {String additionalMessage = ''})
+      : super('Value $val has the wrong width.'
+            ' Excepted $expectedWidth but found $actualWidth.'
```
mkorbel1, line 27: Typo: "expected"

lib/src/modules/conditional.dart
```diff
@@ -890,6 +889,43 @@ enum ConditionalType {
   priority
 }
 
+/// Shorthand for a [Conditional] inside a [Case] block.
```
mkorbel1, line 892: Backwards, this is a `Case` inside of a `Conditional`.    Also, "Returns a [Logic] signal based on the expression and conditions." doesn't really elaborate on how to use it or what it will do.  Imagine a user is reading this doc comment in order to learn what this function does and what to expect it will do.  You can even include some example code in the doc comment.

Also, it's usually best to keep 1 line empty after the first line.

Here's some tips on writing doc comments:
https://dart.dev/effective-dart/documentation

lib/src/modules/conditional.dart
```diff
@@ -890,6 +889,43 @@ enum ConditionalType {
...
+Logic cases(Logic expression, Map<Logic, dynamic> conditions, {int? width}) {
+  for (final condition in conditions.entries) {
+    int? inferredWidth;
+    if (condition.value is Logic) {
+      inferredWidth = (condition.value as Logic).width;
+    } else if (condition.value is LogicValue) {
+      inferredWidth = (condition.value as LogicValue).width;
+    }
+
+    width ??= inferredWidth;
```
mkorbel1, line 905: Feels like something's not right here.  If the `inferredWidth` doesn't match a provided `width` that is not null, we should throw an exception.  After this line, we're overriding the value in `width`.

lib/src/modules/conditional.dart
```diff
@@ -890,6 +889,43 @@ enum ConditionalType {
   priority
 }
 
+/// Shorthand for a [Conditional] inside a [Case] block.
+/// Returns a [Logic] signal based on the expression and conditions.
+/// The width of the result is determined by the width of the value of the
+/// conditions.
+Logic cases(Logic expression, Map<Logic, dynamic> conditions, {int? width}) {
```
mkorbel1, line 896: Should we support inferring widths for the keys of `conditions` also?  E.g. then you could put `LogicValue`, `Logic`, `int`, etc. as the match for the `expression`.

mkorbel1 (changes requested):

lib/src/modules/conditional.dart
mkorbel1, line 896: Add `conditionalType` as well

lib/src/modules/conditional.dart
```diff
@@ -890,6 +889,71 @@ enum ConditionalType {
...
+
+    if (condition.key is LogicValue) {
+      if (expression.width != (condition.key as LogicValue).width) {
+        throw SignalWidthMismatchException.forDynamic(condition.key,
+            expression.width, (condition.key as LogicValue).width);
+      }
+    }
+  }
+
+  final result = Logic(name: 'result', width: width!);
```
mkorbel1, line 938: why can't `width` be null here?  what if none of the inputs provided something to infer the width?  we probably need a check before this to ensure that `width` is provided if it couldnt be inferred.  also means we're missing a test

lib/src/modules/conditional.dart
```diff
@@ -890,6 +889,71 @@ enum ConditionalType {
...
+          condition.value, width, inferredWidth);
+    }
+
+    width ??= inferredWidth;
+
+    if (width != inferredWidth && inferredWidth != null) {
+      throw SignalWidthMismatchException.forDynamic(
+          condition.value, width!, inferredWidth);
+    }
+    if (condition.key is Logic) {
```
mkorbel1, line 923: minor: spacing between these if statements would be nice?

lib/src/modules/conditional.dart
```diff
@@ -890,6 +889,71 @@ enum ConditionalType {
...
+        expression,
+        [
+          for (final condition in conditions.entries)
+            CaseItem(
+                condition.key is Logic
+                    ? condition.key as Logic
+                    : Const(condition.key, width: expression.width),
+                [result < condition.value])
+        ],
+        conditionalType: conditionalType)
```
mkorbel1, line 951: We're missing a default, can we add that to the API?  It's already built-in to `Case`, just need to add a default value to the function.  Probably need to do some additional checks on the width of it and add tests for it too

test/gate_test.dart
```diff
@@ -187,6 +188,104 @@ void main() {
...
+    control.put(2);
+
+    expect(result.value, LogicValue.ofInt(2, 8));
+
+    control.put(3);
+
+    expect(result.value, LogicValue.ofInt(3, 8));
+  });
+
+  test('Cases test Logic', () {
```
mkorbel1, line 231: You could put all these `test`s in a `group`

mkorbel1 (changes requested):
Looking good! few more changes

mkorbel1 (changes requested):

test/gate_test.dart
```diff
@@ -187,6 +188,133 @@ void main() {
...
+    control.put(0);
+
+    expect(result.value, LogicValue.zero);
+
+    control.put(1);
+
+    expect(result.value, LogicValue.one);
+  });
+
+  group('cases', () {
```
mkorbel1, line 208: Missing test cases where the keys of the `conditions` are not `Logic`s

mkorbel1, line 208: Missing test cases where the values of `conditions` are `Logic`s

mkorbel1 (changes requested):

mkorbel1:
Your markdown failure is my fault, fixing in https://github.com/intel/rohd/pull/398

test/gate_test.dart
```diff
@@ -187,6 +188,145 @@ void main() {
...
+
+    test('test Null width Exception', () {
+      final control = Logic();
+      const d0 = 2;
+      const d1 = 4;
+
+      control.put(LogicValue.zero);
+      expect(() => cases(control, {d0: 2, d1: 4}, defaultValue: 3),
+          throwsA(isA<SignalWidthMismatchException>()));
+    });
```
mkorbel1, line 327: Missing a test where a provided `width` and `inferredWidth` don't match?

mkorbel1, line 327: Missing test where inferred LogicValue widths for values are wrong and throws an exception?

mkorbel1, line 327: Missing a test where LogicValue widths for keys don't match expression and throws an exception?

mkorbel1 (changes requested):

mkorbel1 (approved):
I think this looks good to me, great work!

Just need to get the CI checks passing, looks like a minor lint issue.

Issue #290, opened 2023-02-27
`Combinational` is overly optimistic in equivalent `always_comb` inferred sensitivities


mkorbel1:
### Describe the bug

In SystemVerilog, the `always_comb` sensitivities are only signals referenced within the `always_comb` block.  In ROHD, currently, any combinational signal that may modify the input to a `Combinational` is treated as a sensitivity.  When the `always_comb` is created, it can generate a non-equivalent implementation relative to the ROHD simulation.

When inlineable modules are used, things are generally equivalent thanks to the fix for https://github.com/intel/rohd/issues/158.

However, if the module is *not* inlined (e.g. from another generator, see https://github.com/steveicarus/iverilog/issues/872) or the module itself is non-inlineable (easy to create with the existing ROHD generator), then the behavior of the simulation in ROHD will not necessarily match the SystemVerilog simulation.

Perhaps a modification from `always_comb` to `always @(list, of, sensitivities)` could help, but this has two problems:
1. The sensitivity could be outside of scope of the current module.  XMRs might help, but that's messy and tricky to implement.
2. Assignments and blocks in other locations (other than the "current" `always_comb`) can be treated as separate processes by the SystemVerilog simulator and lead to non-deterministic race conditions in simulation behavior.

### To Reproduce

This new test (based off others in `comb_math_test.dart`) will fail due to simulation behavior mismatch in ROHD and Icarus Verilog because the `&` gate has been moved into a non-inlineable module.

```dart
class Ander extends Module {
  Logic get c => output('c');
  Ander(Logic a, Logic b) {
    a = addInput('a', a, width: a.width);
    b = addInput('b', b, width: b.width);
    addOutput('c', width: a.width) <= a & b;
  }
}

class ExtraModuleSensitivitySimplerExample extends Module {
  Logic get b => output('b');
  ExtraModuleSensitivitySimplerExample(Logic a) {
    a = addInput('a', a, width: 8);
    addOutput('b', width: 8);

    final inner = Logic(name: 'inner', width: 8);

    Combinational([
      inner < 0xf,
      b < Ander(a, inner).c,
      inner < 0,
    ]);
  }
}

void main() {
  test('extra module sensitivity simpler example', () async {
    final a = Logic(name: 'a', width: 8);
    final mod = ExtraModuleSensitivitySimplerExample(a);
    await mod.build();

    final vectors = [
      Vector({'a': 0xff}, {'b': bin('00001111')})
    ];
[...]

PR #387, opened 2023-06-17
Fix bug with SSA reuse of signals

mkorbel1:
## Description & Motivation

In some cases of logic reuse for SSA, there was a scenario where signals could attempt to reconnect to themselves causing false failures.  This PR fixes the bug by skipping over those cases.

This bug was discovered via the integer pipeline multiplier in ROHD-HCL with an updated dependency.

## Related Issue(s)

N/A

## Testing

Added 2 tests:
- one is based on the pipeline code from ROHD-HCL pipelined integer multiplier
- one is a distilled minimal example to reproduce the issue in just SSA

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -697,6 +697,17 @@ abstract class Conditional {
         throw UninitializedSignalException(ssaDriver._ref.name);
       }
 
+      // if these are already connected, just skip it, we're fine already
+      if (ssaDriver.srcConnection != null &&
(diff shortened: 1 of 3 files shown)


quekyj:
What is SSA?

mkorbel1:
> What is SSA?

https://en.wikipedia.org/wiki/Static_single-assignment_form

See https://github.com/intel/rohd/pull/344 and https://github.com/intel/rohd/issues/290 for how we got to using it.  I am going to write something up about it soon.

PR #388, opened 2023-06-19
Refactor docs and README for website

### LINKED ISSUE #374: Documents in `user_guide` aren't ideal for non-website viewing
### Describe the bug

The "user guide" isn't in an ideal format for users to peruse without referencing the official website.

For example, the links reference `{{ site.baseurl }}` which doesn't exist until Jekyll parses it:
https://github.com/intel/rohd/blob/main/doc/user_guide/_docs/A11-interfaces.md

We should do a sweep through to see if there's any other issues like this.

Also, the file naming with "A11" etc. makes it annoying to add new documents inserted not at the end.  Is there another convention or mechanism (table of contents instead?) that we could try?

### To Reproduce

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

_No response_

mkorbel1:
## Description & Motivation

As we move to using the ROHD Website as the main landing area, this PR migrates most of the README into separate documents for the website.  Mostly this is deleting stuff from README and including updates from it into the changes from #359.

## Related Issue(s)

Fix #374

## Testing

CI should catch things, will also test with rohd-website fork

## Backwards-compatibility

No

## Documentation

Only docs


quekyj:
The readme is good. But do you think that we should add more details like a reference to the tutorials / bootcamp or the upcoming blog? Make it more user friendly by listing some of the features on the website. Maybe also link the installation page to the website in the readme? I think we can follow what flutter did on their page https://github.com/flutter/flutter.

doc/user_guide/_docs/A03-constant.md
```diff
@@ -1,14 +1,12 @@
 ---
-title: "Constant"
+title: "Constants"
```
quekyj, line 2: I am not sure if changing the word here will break the link in https://github.com/intel/rohd-website/blob/main/docs/_data/navigation.yml.

doc/user_guide/_docs/A03-constant.md
```diff
@@ -1,14 +1,12 @@
...
 permalink: /docs/constant/
-excerpt: "Constant"
+excerpt: "Constants"
```
quekyj, line 4: Here as well.

doc/user_guide/_docs/B01-comparisons-with-alternatives.md
```diff
@@ -0,0 +1,86 @@
+---
+title: "Comparison with Alternatives"
```
quekyj, line 2: Probably need to add to the https://github.com/intel/rohd-website/blob/main/docs/_data/navigation.yml to let rohd-website pick this up.

doc/user_guide/_get-started/03-development-recommendations.md
```diff
@@ -0,0 +1,16 @@
+---
+title: "Development Recommendations"
```
quekyj, line 2: Add to the https://github.com/intel/rohd-website/blob/main/docs/_data/navigation.yml as well.

quekyj (changes requested):
Loog good to me. But maybe need a little changes on the https://github.com/intel/rohd-website/blob/main/docs/_data/navigation.yml to let rohd-website pick up the pages route correctly

doc/user_guide/_docs/A03-constant.md
mkorbel1: I was under the impression that as long as the permalink didn't change it would be safe?

doc/user_guide/_docs/B01-comparisons-with-alternatives.md
mkorbel1: Working on adding stuff here: https://github.com/intel/rohd-website/pull/23

doc/user_guide/_docs/A03-constant.md
quekyj: Ya, I think this should be okay.

doc/user_guide/_docs/A03-constant.md
quekyj: I think this should be fine too

quekyj (approved):

PR #393, opened 2023-07-20
Gates should output X (never Z) when inputs are invalid

### LINKED ISSUE #235: Gates should output `X` (never `Z`) when inputs are invalid
### Describe the bug

Some gates (e.g. the `Mux`) will potentially output a `Z` if the input (e.g. the control) is a `Z` instead of outputting an `X`.  We should fix the `Mux` and also search for any other gates or built-in modules that may be outputting a `Z` instead of an `X` when inputs are invalid.

### To Reproduce

Set the control on `Mux` to `Z`, and observe output value.

### Expected behavior

Output is `X`

### Actual behavior

Output is `Z`

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

_No response_

dmetis:
## Description & Motivation

Some gates (e.g. the Mux) will potentially output a Z if the input (e.g. the control) is a Z instead of outputting an X. We should fix the Mux and also search for any other gates or built-in modules that may be outputting a Z instead of an X when inputs are invalid.
## Related Issue(s)

fix #235 

## Backwards-compatibility

## Documentation


mkorbel1:
Fixed bad tests here: https://github.com/intel/rohd/pull/397

mkorbel1:
If you pull from main those tests should be fixed now

lib/src/modules/conditional.dart
```diff
@@ -831,7 +831,12 @@ class ConditionalAssign extends Conditional {
...
-    _receiverOutput.put(driverValue(driver));
+    final currentValue = driverValue(driver);
+    if (!currentValue.isValid) {
+      _receiverOutput.put(LogicValue.x);
+    } else {
+      _receiverOutput.put(currentValue);
+    }
 
     if (!drivenSignals.contains(receiver) || receiver.value.isValid) {
       drivenSignals.add(receiver);
```
mkorbel1, line 842: I think there's another case in the `If` where we need to make sure to prop a X instead of Z:
```dart
      } else if (driverValue(iff.condition) != LogicValue.zero) {
        // x and z propagation
        for (final receiver in receivers) {
          receiverOutput(receiver).put(driverValue(iff.condition)[0]);
          if (!drivenSignals.contains(receiver) || receiver.value.isValid) {
            drivenSignals.add(receiver);
          }
        }
        break;
      }
```

lib/src/modules/conditional.dart
```diff
@@ -831,7 +831,12 @@ class ConditionalAssign extends Conditional {
       guard(driver);
     }
 
-    _receiverOutput.put(driverValue(driver));
+    final currentValue = driverValue(driver);
```
mkorbel1, line 834: We're missing a test that checks this `ConditionalAssign` change

mkorbel1 (changes requested):

mkorbel1 (approved):
Looks good to me!

PR #403, opened 2023-09-05
Unique case with multiple match behavior

### LINKED ISSUE #107: Unique `Case` with multiple match behavior
## Motivation
When multiple cases match on a unique case statement, it currently causes an exception to be thrown.  Perhaps this should be a configurable option allowing X's to be driven instead.

## Desired solution
Always drive X when there's a violation of unique rules, and raise exception on it as well in a configurable way (default enabled) on a per-`Case` basis.

## Alternatives considered  
A global switch for this type of check could also be an option.

## Additional details
SystemVerilog by default just prints a warning message (http://www.verilogpro.com/systemverilog-unique-priority/) which seems not great.

dmetis:
## Description & Motivation

When multiple cases match on a unique case statement, it currently causes an exception to be thrown. Perhaps this should be a configurable option allowing X's to be driven instead. 

## Related Issue(s)

Fix #107

## Backwards-compatibility

## Documentation

### DIFF
--- a/lib/src/modules/conditional.dart
+++ b/lib/src/modules/conditional.dart
@@ -740,6 +740,16 @@ abstract class Conditional {
   /// This is used for [Combinational.ssa].
   Map<Logic, Logic> _processSsa(Map<Logic, Logic> currentMappings,
       {required int context});
+
+  /// Drives X to all receivers.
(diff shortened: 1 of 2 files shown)


mkorbel1 (approved):
Looks good!

PR #442, opened 2023-12-01
Absolute value

### LINKED ISSUE #360: Add absolute
### Motivation

It would be useful to be able to calculate the absolute value of a signal or value (convert negative to positive, or leave positive as is).

### Desired solution

Add a function to `Logic` and `LogicValue` that allows computation of absolute value.  For `Logic`, it should generate a `Module` like other operators.  Assume 2's complement for negative values.

### Alternatives considered

_No response_

### Additional details

_No response_

dmetis:
## Description & Motivation

It would be useful to be able to calculate the absolute value of a signal or value (convert negative to positive, or leave positive as is).

## Related Issue(s)

Fix #360 

## Backwards-compatibility

## Documentation

### DIFF
--- a/lib/src/signals/logic.dart
+++ b/lib/src/signals/logic.dart
@@ -1,4 +1,4 @@
-// Copyright (C) 2021-2023 Intel Corporation
+// Copyright (C) 2021-2024 Intel Corporation
 // SPDX-License-Identifier: BSD-3-Clause
 //
 // logic.dart
@@ -638,10 +638,8 @@ class Logic {
     ].swizzle();
   }
 
-  /// This function calculates the absolute value of a signal. It first checks
-  /// the size of the signal. Then checks to see if upper bits is a positive or
-  /// negative value. In a postive scenario, it returns the same value. In a
-  /// negative scenario, it  inverts the values and add 1 to it.
+  /// Calculates the absolute value of a signal, assuming that the
+  /// number is a two's complement.
   Logic abs() {
     if (width == 0) {
       return this;
--- a/lib/src/signals/logic.dart
+++ b/lib/src/signals/logic.dart
@@ -633,7 +633,10 @@ class Logic {
     ].swizzle();
   }
 
-  ///This function calculate the value of a signal or value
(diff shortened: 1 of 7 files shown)


lib/src/signals/logic.dart
```diff
@@ -633,7 +633,15 @@ class Logic {
...
-  /// Returns a new [Logic] with width [newWidth] where new bits added are sign
+  ///This function calculate the value of a signal or value
+  Logic abs() {
+    if (width == 0) {
+      return this;
+    }
+    return mux(this[-1], ~this + 1, this);
+  }
+
+  /// Returns a new [Logic] wi}th width [newWidth] where new bits added are sign
```
mkorbel1, line 644: Extra `}` got thrown into this comment

lib/src/signals/logic.dart
```diff
@@ -633,7 +633,15 @@ class Logic {
     ].swizzle();
   }
 
-  /// Returns a new [Logic] with width [newWidth] where new bits added are sign
+  ///This function calculate the value of a signal or value
```
mkorbel1, line 636: Let's make this a little more descriptive.  Some important topics to hit:
- It performs an "absolute value" on the signal
- Treats the signal as a two's complement signed number

Also, style-wise, put a space between the `///` and your documentation

lib/src/values/logic_value.dart
```diff
@@ -863,6 +863,18 @@ abstract class LogicValue implements Comparable<LogicValue> {
     return op(this, other);
   }
 
+  LogicValue abs() {
```
mkorbel1, line 866: Add a similar doc comment here as well

lib/src/signals/logic.dart
mkorbel1, line 1: Reminder to test the `abs` function for `Logic`

lib/src/values/logic_value.dart
mkorbel1, line 1: Reminder to test the `abs` function for `LogicValue`

mkorbel1 (approved):
Looks great!

mkorbel1 (approved):

PR #461, opened 2024-02-09
chore(devtool): build devtool artifact and commit to other branch

quekyj:
## Description & Motivation

To test out the unreleased version of the devtool_extension, we can allow github action to commit the build artifact and modified the .gitignore. So, user can point to this package in pubspec.yaml via the use of:

```yaml

rohd:
     git:
       url: https://github.com/quekyj/rohd.git
       ref: artifacts
```

Branch that have the artifact:
- artifacts

## Testing

Tested on fork.

## Backwards-compatibility

no

## Documentation

No.

### DIFF
--- a/.github/workflows/build_devtool.yml
+++ b/.github/workflows/build_devtool.yml
@@ -30,10 +30,6 @@ jobs:
         - name: Build Static Web
           run: tool/gh_actions/devtool/build_web.sh
 
-        # 1. checkout to branch artifacts
-        # 2. Do not ignore build/ folder (just for artifacts branch)
-        # 3. Commit everythings including the build/ folder
-        # 4. Push the new branch artifacts
         - name: Create artifact branch and commit
           run: |
             git config --global user.name "${GITHUB_ACTOR}"
@@ -41,8 +37,8 @@ jobs:
             git fetch origin
             git checkout -b artifacts
             git pull origin main
-            sed -i '/build\//d' .gitignore  # Remove the line that ignores the build directory
             git add .
+            git add -f extension/*
             git commit -m "Add build artifacts and modify .gitignore"
 
         - name: Push changes
--- a/.github/workflows/build_devtool.yml
+++ b/.github/workflows/build_devtool.yml
@@ -1,0 +1,53 @@
+name: Build ROHD devtool
+on:
+  push:
+    branches:
+      - main
+
+# Top-level default, no permissions
+permissions: {}
+
+jobs:
+  build-devtool:
+      name: Build Devtools
+      permissions:
+        contents: write
+        pull-requests: write
+      timeout-minutes: 30
+      runs-on: ubuntu-22.04
+      steps:
+        - uses: actions/checkout@v2
+
+        - name: Setup Flutter SDK
+          uses: flutter-actions/setup-flutter@v2
+          with:
+            channel: beta
+            version: 3.18.0-0.2.pre
+
+        - name: Run Flutter Test
+          run: tool/gh_actions/devtool/run_devtool_test.sh
+
+        - name: Build Static Web
+          run: tool/gh_actions/devtool/build_web.sh
+
+        # 1. checkout to branch artifacts
+        # 2. Do not ignore build/ folder (just for artifacts branch)
+        # 3. Commit everythings including the build/ folder
+        # 4. Push the new branch artifacts
+        - name: Create artifact branch and commit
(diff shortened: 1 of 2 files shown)


.github/workflows/general.yml
```diff
@@ -112,4 +115,26 @@ jobs:
...
+      # 3. Commit everythings including the build/ folder
+      # 4. Push the new branch artifacts
+      - name: Create artifact branch and commit
+        run: |
+          git config --global user.name "${GITHUB_ACTOR}"
+          git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+          git fetch origin
+          git checkout -b artifacts
+          git pull origin main
+          sed -i '/build\//d' .gitignore  # Remove the line that ignores the build directory
```
mkorbel1, line 129: doesn't `git add -f` or something work fine here rather than trying to modify .gitignore?

.github/workflows/general.yml
```diff
@@ -112,4 +115,26 @@ jobs:
...
+          git add .
+          git commit -m "Add build artifacts and modify .gitignore"
+
+      - name: Push changes
```
mkorbel1, line 133: the push should only happen if there's a push to main, not for pull requests!  otherwise every PR will push immediately to this branch

mkorbel1 (changes requested):

.github/workflows/build_devtool.yml
```diff
@@ -0,0 +1,49 @@
...
+on:
+  push:
+    branches:
+      - main
+
+# Top-level default, no permissions
+permissions: {}
+
+jobs:
+  build-devtool:
```
mkorbel1, line 11: I think it's still important to have the build-devtool as part of the CI for every branch, we just want to limit the *push to the other branch* to only when there's an actualy merge into main.

One way I think you can do this is leaving it in the general.yml, but adding some conditions on when it should run vs. be skipped: https://docs.github.com/en/actions/using-jobs/using-conditions-to-control-job-execution . The push can be a separate step which depends on the previous step.

An easier solution might just be to keep this file as-is, but add back the build alone into the general.yml CI flow.  There's a little duplication, but it's simpler to set up and see what's happening I think.

mkorbel1 (changes requested):
Looks great!

We should include information about this branch in the user guide section we write about the devtools extension

mkorbel1:
Were you able to get the extension running by pointing at the `artifacts` branch?

I tried adding this into a `pubspec.yaml` (not in the `rohd` package itself):
```yaml
dependency_overrides:
  rohd:
    git:
      url: https://github.com/quekyj/rohd.git
      ref: artifacts
```

It looks like all the build artifacts are there in the downloaded dependency, but I don't see it showing up in the devtools?

quekyj:
> Were you able to get the extension running by pointing at the `artifacts` branch?
> 
> I tried adding this into a `pubspec.yaml` (not in the `rohd` package itself):
> 
> ```yaml
> dependency_overrides:
>   rohd:
>     git:
>       url: https://github.com/quekyj/rohd.git
>       ref: artifacts
> ```
> 
> It looks like all the build artifacts are there in the downloaded dependency, but I don't see it showing up in the devtools?

Yes, I can get it working. Are you using the beta version of flutter?

.github/workflows/build_devtool.yml
quekyj: Yup. I think I will just go with the easier option for now. We can add this as a good first issue.

mkorbel1:
> Yes, I can get it working. Are you using the beta version of flutter?

Really?  I tried in WSL using latest Dart version:
`Dart SDK version: 3.3.0 (stable) (None) on "linux_x64"`

I don't see it showing up

.github/workflows/build_devtool.yml
```diff
@@ -0,0 +1,49 @@
...
+        - name: Create artifact branch and commit
+          run: |
+            git config --global user.name "${GITHUB_ACTOR}"
+            git config --global user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+            git fetch origin
+            git checkout -b artifacts
+            git pull origin main
+            git add .
+            git add -f extension/*
+            git commit -m "Add build artifacts and modify .gitignore"
```
mkorbel1, line 42: this actually doesn't modify .gitignore?

mkorbel1 (approved):
looks good!  thank you!

PR #474, opened 2024-03-12
fix: update flutter version to the latest

quekyj:
## Description & Motivation

Fix the build fail on flutter devtool extension.

## Related Issue(s)

NA

## Testing

Tested on forked main branch.

## Backwards-compatibility

no

## Documentation

no

### DIFF
--- a/.github/workflows/build_devtool.yml
+++ b/.github/workflows/build_devtool.yml
@@ -22,7 +22,7 @@ jobs:
           uses: flutter-actions/setup-flutter@v2
           with:
             channel: stable
-            version: 3.16.9
+            version: 3.19.3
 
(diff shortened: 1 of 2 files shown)


.github/workflows/build_devtool.yml
```diff
@@ -22,7 +22,7 @@ jobs:
           uses: flutter-actions/setup-flutter@v2
           with:
             channel: stable
-            version: 3.16.9
+            version: 3.19.3
```
mkorbel1, line 25: can you make the version in `general.yml` exactly the same as the one in`build_devtool.yml` so that we are always certain this will pass if it passes CI?

mkorbel1 (approved):

PR #478, opened 2024-03-26
Transition to using a different addition syntax for lint avoidance

### LINKED ISSUE #444: Improve lint avoidance for width expansion
### Motivation

The current implementation (https://github.com/intel/rohd/issues/298, https://github.com/intel/rohd/pull/423) for handling width expansion (e.g. from adding, shift left, multiplication, etc.) avoids lint errors successfully, but still hits some lint warnings and is also somewhat ugly.

A nicer way to generate it might look something like this:
```SystemVerilog
assign {cout, out} = a + b + cin;
```

### Desired solution

Upgrade modules to generate prettier SV that avoids both lint errors and warnings

### Alternatives considered

_No response_

### Additional details

_No response_

mkorbel1:
## Description & Motivation

See #444

## Related Issue(s)

Fix #444

## Backwards-compatibility

## Documentation


lib/src/modules/gates.dart
```diff
@@ -469,13 +469,99 @@ class Power extends _TwoInputBitwiseGate {
...
+
+    _in0Name = Naming.unpreferredName('in0_${in0.name}');
+    _in1Name = Naming.unpreferredName('in1_${in1Logic.name}');
+    _outName = Naming.unpreferredName('${in0.name}_${name}_${in1Logic.name}');
+    _carryName = Naming.unpreferredName('carry_$_outName');
+
+    addInput(_in0Name, in0, width: width);
+    addInput(_in1Name, in1Logic, width: width);
+    addOutput(_outName, width: width);
+    addOutput(_carryName); // just for SV generation, not used
```
mkorbel1, line 531: we could expose the carry as well in this module, just not use it for `+` operator

PR #481, opened 2024-04-05
Avoid Module creation for Constant Gate scenarios

### LINKED ISSUE #429: Avoid module creation for simple constant scenarios in gates
### Motivation

Sometimes it's easy to initialize some variable to 0 or 1, and then do multiple operator functions on it afterwards like `&`, `~`, and `|`.  There are some common rules that we can use to avoid creating gates when not necessary.  For example:
- `~0 == 1` (inversion of a constant is constant of inversion)
- `(x & 0) == 0`
- `(x & 1) == x`
- `(x | 0) == x`
- `(x | 1) == 1`

Skipping over these scenarios would allow for boosts in simulation performance and cleaner generated verilog.

### Desired solution

Modify the operator overrides in `Logic` to avoid creation of `Module`s if it can instead directly compute a `Logic` to return (like in above situations).

Note that this won't completely avoid all the applicable scenarios alone since there could be things like 0 assigned to another signal which is then used in a gate.  Perhaps some optimization in the generation of verilog after assignment collapsing could help even more.

### Alternatives considered

_No response_

### Additional details

_No response_

mjayasim9:
## Description & Motivation

#429 

## Related Issue(s)

#429 

## Testing

Added test cases to check if the result of '&', '|' and '~' operations are as expected for constant input scenarios.

## Backwards-compatibility

No.

## Documentation

No.


lib/src/signals/logic.dart
```diff
@@ -319,13 +319,59 @@ class Logic {
   void operator <=(Logic other) => gets(other);
 
   /// Logical bitwise NOT.
-  Logic operator ~() => NotGate(this).out;
+  Logic operator ~() {
+    if (this is Const) {
+      if (value.isValid) {
```
mkorbel1, line 324: I think we can rely on the operators within `LogicValue` for this?  i.e. something like `Const(~this.value)`?

lib/src/signals/logic.dart
```diff
@@ -319,13 +319,59 @@ class Logic {
...
+        if (value == LogicValue.of(0, width: width) ||
+            other.value == LogicValue.of(0, width: width)) {
+          return Const(LogicValue.of(0, width: width));
+        } else if (value == LogicValue.filled(width, LogicValue.one)) {
+          return other;
+        } else if (other.value == LogicValue.filled(width, LogicValue.one)) {
+          return this;
+        }
+      } else {
+        return Const(LogicValue.x, width: width);
```
mkorbel1, line 350: This isn't quite right, consider 11x11 & 11111 == 11x11

If we just cover the cases of all 0's and all 1's that's probably enough and let the rest fall through to the and gate?

lib/src/signals/logic.dart
```diff
@@ -319,13 +319,59 @@ class Logic {
...
   /// Logical bitwise AND.
-  Logic operator &(Logic other) => And2Gate(this, other).out;
+  Logic operator &(Logic other) {
+    if (this is Const || other is Const) {
+      if (value.isValid && other.value.isValid) {
+        if (value == LogicValue.of(0, width: width) ||
+            other.value == LogicValue.of(0, width: width)) {
+          return Const(LogicValue.of(0, width: width));
+        } else if (value == LogicValue.filled(width, LogicValue.one)) {
+          return other;
```
mkorbel1, line 345: What if `other is Const` and `this` is not `Const` but has a value of all 1s currently?  Need to be cautious to only consider the value if it's a `Const`.

test/gate_test.dart
```diff
@@ -353,6 +353,115 @@ void main() {
...
+    });
+
+    test('And2Gate Single bit Constant input', () async {
+      final a = Logic();
+      final b = Const(LogicValue.zero, width: 1);
+      final c = Const(LogicValue.x, width: 1);
+      final d = Const(LogicValue.one, width: 1);
+      a.put(LogicValue.one);
+
+      expect(a & b, isA<Logic>());
```
mkorbel1, line 392: anything `& 0` should be a constant `0`

test/gate_test.dart
```diff
@@ -353,6 +353,115 @@ void main() {
...
+      a.put(LogicValue.one);
+
+      expect(a | b, isA<Logic>());
+      expect((a | b).value, equals(a.value));
+
+      expect(a | c, isA<Logic>());
+      expect((a | c).value, equals(LogicValue.one));
+
+      expect(a | d, isA<Logic>());
+      expect((a | d).value, equals(LogicValue.one));
```
mkorbel1, line 446: anything `| 1` should be `1`

mkorbel1 (changes requested):

mkorbel1 (commented):
This is looking very good, and the tests make sense to me for &, |, and ~

mkorbel1:
superseded by https://github.com/intel/rohd/pull/682

Issue #42, opened 2021-11-05
Support multiple triggers on flops


mkorbel1:
**Is your feature request related to a problem? Please describe.**
Support multiple triggers for flops, like in SystemVerilog:
```SystemVerilog
always_ff @(posedge clk or posedge reset)
```

**Describe the solution you'd like**
The `FF` and `FlipFlop` should have a syntax that accepts a list of signals instead of a single `clk` signal.

**Describe alternatives you've considered**
Using something like `clk | reset` is insufficient because there could be a rising edge of reset during a clock high period.

wswongat:
@mkorbel1 Is it possible to support negative edge trigger as well?
It seems like the rohd only support the positive edge single trigger with current version.
https://github.com/intel/rohd/blob/eb4fd660e9f4bb646c76f1bb7b900a445cc9ec79/lib/src/modules/conditional.dart#L222

mkorbel1:
@wswongat it wouldn't be too hard to add that, but I figured it wasn't really needed as a separate feature since you could always just use the posedge of an inverted signal, for example:
```dart
FF(clk, [...]); //posedge of clk
FF(~clk, [...]); // negedge of clk
```

Do you think that meets your needs or can you elaborate why you think a separate negedge trigger would be beneficial beyond this?

wswongat:
@mkorbel1 Your suggestion looks good. Looks like we need to find some way to support multiple triggers on flops

mkorbel1:
@wswongat #65 should address this issue and provide support for multiple triggers on flops

PR #483, opened 2024-04-17
Support negative edge-triggered `Sequential`s

mkorbel1:
## Description & Motivation

Adds support for negative edge-triggered `Sequential`s.  In some cases, downstream tools may treat an inverted clock fed into a sequential differently than a negative edge triggered sequential.  Functionally, in simulation, the behavior is the same for either option. Generators may choose to generate these differently (e.g. using `negedge` vs. `posedge` in SystemVerilog generation).

Additionally fixed a bug where a multi-triggered `Sequential` may not generate X's if one trigger is valid and another (later specified) trigger is invalid.

The changes also motivated some refactoring of trigger handling in `Sequential`, as well as renaming of the first positional argument in `Sequential.multi`.

## Related Issue(s)

This is related to the discussion on #42, where previously it was decided not to include `negedge` due to lack of motivation.  Downstream tool differences now motivate adding the feature.

## Testing

Added new tests, existing tests pass.

## Backwards-compatibility

No, but the signature of `Sequential.multi` has been slightly expanded and modified:
- first positional argument changed name from `clks` to `posedgeTriggers`
- added `negedgeTriggers` as a named optional argument

## Documentation

This is a relatively uncommon feature, so documentation added onto the API docs should be sufficient (extra example in user guide not really needed).


mkorbel1:
Superseded by #533

Issue #482, opened 2024-04-15
Make full array assignments compressed to 1 line (or in-lined) in generated SystemVerilog


mkorbel1:
### Motivation

For compatibility purposes, ROHD currently does per-element assignments between arrays.  A common case of assignment between two entire arrays of the same dimensions makes many more lines than is necessary.  This makes the generated SV more verbose than it needs to be.

### Desired solution

Intelligently compress assignments between arrays with the same dimensions to a single-line assignment.

### Additional details

This is not a functional change, just makes the generated SV nicer.

mkorbel1:
Closed by #487

PR #487, opened 2024-06-06
RTL to have one array-array assignment instead of bit blasted assignments 

sshankar4:
## Description & Motivation

For compatibility purposes, ROHD currently does per-element assignments between arrays. A common case of assignment between two entire arrays of the same dimensions makes many more lines than is necessary. This makes the generated SV more verbose than it needs to be.

## Related Issue(s)

#482 
## Testing

Added new tests

## Backwards-compatibility

No

## Documentation

Documentation is not required for this change.


test/array_collapsing_test.dart
```diff
@@ -0,0 +1,112 @@
+// Copyright (C) 2022-2023 Intel Corporation
```
mkorbel1, line 1: copyright 2024

test/array_collapsing_test.dart
```diff
@@ -0,0 +1,112 @@
...
+        contains('net_connect #(.WIDTH(9)) net_connect_0 (b, intermediate);'));
+
+    final vectors = [
+      Vector({'a': 0}, {'b': 0}),
+      Vector({'a': 123}, {'b': 123}),
+    ];
+    await SimCompare.checkFunctionalVector(mod, vectors);
+    SimCompare.checkIverilogVector(mod, vectors);
+
+    //TODO: what if unpacked?
```
mkorbel1, line 70: ditch the TODO's in this file

test/array_collapsing_test.dart
```diff
@@ -0,0 +1,112 @@
...
+
+    final vectors = [
+      Vector({'a': 0}, {'b': 0}),
+      Vector({'a': 123}, {'b': 123}),
+      Vector({'c': 6}, {'d': 6}),
+    ];
+
+    await SimCompare.checkFunctionalVector(mod, vectors);
+    SimCompare.checkIverilogVector(mod, vectors);
+  });
```
mkorbel1, line 111: are there any other more complex conditions to consider to test?  im not thinking of any off the top of my head

mkorbel1 (changes requested):

test/array_collapsing_test.dart
```diff
@@ -87,8 +86,6 @@ void main() {
       Vector({'a': 123}, {'b': 123}),
     ];
     await SimCompare.checkFunctionalVector(mod, vectors);
-    SimCompare.checkIverilogVector(mod, vectors);
```
mkorbel1: this unpacked fails with iverilog? just checking

mkorbel1 (approved):

test/array_collapsing_test.dart
mkorbel1: actually, does it fail in iverilog due to tool support or is it illegal in SystemVerilog?  If it's illegal, then we should be doing a check to *not* merge assignments for unpacked arrays instead of just removing this test.

sshankar4: yes it fails

sshankar4: Hmm.. I haven't checked if if it's illegal. I was just under the impression that its not supported in iverilog. Will not merge until I make sure.

mkorbel1: looks like it IS legal from LRM 7.4.3 to do unpacked to unpacked assignments

PR #526, opened 2024-10-05
Improved port already exists error message with module name

### LINKED ISSUE #521: Improve error message for when a port already exists on a module
### Motivation

If exceptions are not being debugged, it can be difficult to identify where the `UnavailableReservedNameException` is coming from (e.g. which module).

### Desired solution

In throwing the `UnavailableReservedNameException` in `_checkForSafePortName`, also communicate the module name.

### Alternatives considered

_No response_

### Additional details

_No response_

c0d3-br3ak9r:
## Description & Motivation

It is difficult to debug and identify from where (which module) `UnavailableReservedNameException` is getting thrown

## Related Issue(s)

Fixes #521 

## Testing

Added duplicate port names in example files and verified that the exception message includes the module name

## Backwards-compatibility

NA

## Documentation

NA

### DIFF
--- a/lib/src/module.dart
+++ b/lib/src/module.dart
@@ -617,7 +617,7 @@ abstract class Module {
         inputs.containsKey(name) ||
         inOuts.containsKey(name)) {
       throw UnavailableReservedNameException.withMessage(
-          'Already defined a port with name "$name".');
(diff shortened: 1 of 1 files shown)


mkorbel1 (approved):
Looks good, thank you for your contribution!

PR #529, opened 2024-10-30
RadixString code

desmonddak:
## Description & Motivation

Added LogicValue radix string conversion routines.

## Related Issue(s)

None

## Testing

Added tests that include roundtrip conversions and corner cases.

## Backwards-compatibility

No.

## Documentation

Yes:  updated the user-guide with an example of conversion.  Documentation in the source can eventually be put in the user-guide.


mkorbel1 (approved):
Awesome!

Issue #534, opened 2024-12-23
LogicValue radixStrings should use underscore as separator, similar to Dart 3.6


desmonddak:
### Motivation

Dart has moved to allowing underscores in digit strings to make them more readable.
LogicValue radixStrings were using spaces as separators, not a great choice.
It woud be nice to match Dart with underscores but also allow for other separators upon request.

### Desired solution

Add an optional seperatorChar for ofRadixString and toRadixString, defaulting to '_'.

PR #535, opened 2024-12-23
LogicValue radixString uses underscore separators by default

desmonddak:
## Description & Motivation

Added an option to LogicValue.toRadixString and .ofRadixString to use '_' as a default separator, but allow
for the user to specify a different separator.

## Related Issue(s)

#534 

## Testing

Converted tests to use _ as separator after testing with sepChar argument.

## Backwards-compatibility

Yes.  Previously output radixStrings will not parse by default as they will have spaces.

## Documentation

Documentation in the class code is now up to date with this change.

### DIFF
--- a/lib/src/exceptions/logic_value/logic_value_conversion_exception.dart
+++ b/lib/src/exceptions/logic_value/logic_value_conversion_exception.dart
@@ -1,0 +1,19 @@
+// Copyright (C) 2023 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// logic_value_conversion_exception.dart
+// An exception that is thrown when a conversion from a [LogicValue] fails
+// (such as to a [String]).
+//
+// 2024 December 26
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:rohd/rohd.dart';
+
+/// An exception that is thrown when a [LogicValue] cannot be
+/// properly converted.
+class LogicValueConversionException extends RohdException {
+  /// Creates an exception for when conversion of a `LogicValue` fails.
+  LogicValueConversionException(String message)
+      : super('Failed to convert `LogicValue`: $message');
+}
--- a/lib/src/exceptions/logic_value/logic_value_conversion_exception.dart
+++ b/lib/src/exceptions/logic_value/logic_value_conversion_exception.dart
@@ -1,4 +1,4 @@
-// Copyright (C) 2023 Intel Corporation
+// Copyright (C) 2024 Intel Corporation
 // SPDX-License-Identifier: BSD-3-Clause
 //
 // logic_value_conversion_exception.dart
--- a/lib/src/values/logic_value.dart
+++ b/lib/src/values/logic_value.dart
@@ -605,7 +605,7 @@ abstract class LogicValue implements Comparable<LogicValue> {
   }
 
   /// Legal characters in a radixString representation.
-  static String get radixStringChars => "'0123456789aAbBcCdDeEfFqohzZxX";
+  static const String radixStringChars = "'0123456789aAbBcCdDeEfFqohzZxX";
 
   /// Reverse a string (helper function)
   static String _reverse(String inString) =>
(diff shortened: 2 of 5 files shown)


lib/src/values/logic_value.dart
```diff
@@ -622,7 +625,9 @@ abstract class LogicValue implements Comparable<LogicValue> {
   /// - Decimal: `10'd1492`
   /// - Hex: `15'h05d4`
   ///
-  /// Spaces are output according to [chunkSize] starting from the LSB(right).
+  /// Separators are output according to [chunkSize] starting from the
+  /// LSB(right).  The default separator is '_'.  [sepChar] can be set to
+  /// another character, but not in [radixStringChars].
   ///  - [chunkSize] = default:  `61'h2 9ebc 5f06 5bf7`
```
mkorbel1, line 631: update examples to use `_` also?

lib/src/values/logic_value.dart
```diff
@@ -604,6 +604,9 @@ abstract class LogicValue implements Comparable<LogicValue> {
     }
   }
 
+  /// Legal characters in a radixString representation.
+  static String get radixStringChars => "'0123456789aAbBcCdDeEfFqohzZxX";
```
mkorbel1, line 608: nit: this could be `static const String` instead

lib/src/values/logic_value.dart
```diff
@@ -640,8 +645,11 @@ abstract class LogicValue implements Comparable<LogicValue> {
   /// matter what the length. When leading, 'Z' indicates one or more 'z'
   /// bits to fill the first radix character.
   /// - `9'bz zzzz zzz = 9'hZZZ`
-  ///
-  String toRadixString({int radix = 2, int chunkSize = 4}) {
+  String toRadixString(
+      {int radix = 2, int chunkSize = 4, String sepChar = '_'}) {
+    if (radixStringChars.contains(sepChar)) {
+      throw Exception('separation character invalid');
```
mkorbel1, line 651: We're trying to (at least in ROHD) move to more specific types of exception (see `exceptions` directory) than a generic `Exception`.  Can you see if one meets your needs or add a new one?

test/logic_value_test.dart
mkorbel1, line 1: some more tests to cover new scenarios?
e.g.:
- generate string with different separator
- parse string with different separator
- illegal separator selected -> exception
- wrong separator selected for parsing -> exception

mkorbel1 (changes requested):
nice changes!

test/logic_value_test.dart
desmonddak: Done.  Roundtrip check with different separator, new exception type added.

lib/src/exceptions/logic_value/logic_value_conversion_exception.dart
```diff
@@ -0,0 +1,19 @@
+// Copyright (C) 2023 Intel Corporation
```
mkorbel1, line 1: year :)

lib/src/values/logic_value.dart
```diff
@@ -724,6 +727,10 @@ abstract class LogicValue implements Comparable<LogicValue> {
   /// radix format.  Space-separation is for ease of reading and is often
   /// in chunks of 4 digits.
   ///
+  /// If the format of then length/radix-encoded string is not completely parsed
+  /// an exception will be thrown.  This can be caused by illegal characters
+  /// in the string or too short or too long of a value string.
```
mkorbel1, line 732: what would be an example of too short?  wouldn't it 0-pad?

test/logic_value_test.dart
```diff
@@ -2079,6 +2079,44 @@ void main() {
...
+    });
+    test('radixString space separators', () {
+      final lv = LogicValue.ofRadixString("10'b10 0010 0111", sepChar: ' ');
+      expect(lv.toInt(), equals(551));
+    });
+    test('radixString bad separator', () {
+      try {
+        LogicValue.ofRadixString("10'b10 0010_0111");
+      } on Exception catch (e) {
+        expect(e.runtimeType, LogicValueConstructionException);
```
mkorbel1, line 2100: i think the typical pattern for this is `expect(e, isA<LogicValueConstructionException>())`

lib/src/values/logic_value.dart
desmonddak: I'm updating this documentation and fixing the code to make "10'b"  legal.  before it had to have at least one bit.

mkorbel1 (approved):
Looks great!  Thank you!

Issue #542, opened 2025-01-13
toRadixString should work for radix:10 even if the Logic is X or Z


desmonddak:
### Describe the bug

If you have a Logic that has not been set yet, and you use toRadixString(radix:10)  (an alternative to .toInt(), with more features to support different radices and separators), the code throws due to the Z value.

### To Reproduce

Create a Logic:   final l = Logic();
Print it:    print (l.value.toRadixString(radix:10));

This will throw.

### Expected behavior

We should see the value printed out in radix form with X or Z.  With radix10, this is challenging as bitfields don't line up.
But the number of actual digits can be printed as X or Z.   Like 10'dZZZZ for a 4-digit or 11-bit Logic.

### Actual behavior

We see a throw on unassigned Logic.

mkorbel1:
Fixed in #543

PR #543, opened 2025-01-13
fix radix10 issue in toRadixString

desmonddak:
## Description & Motivation

Bug fix to enable LogicValue.toRadixString(radix:10) to return a string with X or Z instead of throwing.

## Related Issue(s)

#542 

## Testing

Tests added with X and Z in their bit values, tested to ensure a string with appropriate X or Z is returned.

## Backwards-compatibility

No.  Throwing was not an expected behavior.

## Documentation

we don't have documentation on toRadixString except in code.  This will have to be included as we add more extensive documentation to the user guide:  best ways of printing LogicValue for debug.


mkorbel1 (approved):
Looks great!

PR #550, opened 2025-01-18
Add `Logic.named` and broaden `clone`

mkorbel1:
## Description & Motivation

The main things in this PR are:
- adding `Logic.clone`, which behaves similarly to the existing ones for `LogicStructure`
- adding `Logic.named`, which automates the creation of a named clone of a signal (usually for the purpose of making generated outputs prettier)

## Related Issue(s)

N/A

## Testing

Added new tests

## Backwards-compatibility

No

## Documentation

Not really, API docs cover it, maybe some general examples and guidance would be nice in the future

### DIFF
--- a/lib/src/signals/logic.dart
+++ b/lib/src/signals/logic.dart
@@ -255,6 +255,24 @@ class Logic {
           naming: naming,
         );
 
+  Logic _clone({String? name, Naming? naming}) =>
+      (isNet ? LogicNet.new : Logic.new)(
+          name: name ?? this.name,
+          naming: Naming.chooseCloneNaming(
+              originalName: this.name,
+              newName: name,
+              originalNaming: this.naming,
+              newNaming: naming),
+          width: width);
+
+  /// Makes a copy of `this`, optionally with the specified [name] and [naming],
+  /// but the same [width].
+  Logic clone({String? name}) => _clone(name: name);
+
+  /// Makes a [clone] with the provided [name] and optionally [naming].
+  Logic named(String name, {Naming? naming}) =>
+      _clone(name: name, naming: naming); //TODO: NETS!
+
   /// An internal constructor for [Logic] which additional provides access to
   /// setting the [wire].
   Logic._({
--- a/lib/src/signals/logic.dart
+++ b/lib/src/signals/logic.dart
@@ -255,6 +255,7 @@ class Logic {
           naming: naming,
         );
 
+  /// A cloning utility for [clone] and [named].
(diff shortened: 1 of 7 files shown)


ganewto (commented):
These changes look good and the naming is effective, especially with the merging options.

- you may want to reference the Naming option in .named description so people check that out.
- Logic(name: 'my_important_signal').named('my_really_important_signal'); makes it so we can rename signals, but what is really happening is we are getting a new Logic?  We should clarify.
- We have a slight inconsistency with instances now where we cannot do a .named() on a module instantiation similar to above.
- Is there _any_ way to pickup the name of the Dart variable.   If so, we can use decent names for Dart variables and get them for free in the Verilog (and guaranteed to match).  Then we would only used .named() for the merging options or if we have a Verilog conflict or want a different naming convention in Verilog than in Dart.  I find I only use the merging options in detailed components, but not in larger blocks where the names are generally unique.

mkorbel1:
> - you may want to reference the Naming option in .named description so people check that out.
> - Logic(name: 'my_important_signal').named('my_really_important_signal'); makes it so we can rename signals, but what is really happening is we are getting a new Logic? We should clarify.

I think the doc comment covers these?

```dart
  /// Makes a [clone] with the provided [name] and optionally [naming], then
  /// assigns it to be driven by `this`.
  ///
  /// This is a useful utility for naming the result of some hardware
  /// construction without separately declaring a new named signal and then
  /// assigning.  For example:
```

> - We have a slight inconsistency with instances now where we cannot do a .named() on a module instantiation similar to above.

I don't think it makes sense to have a `named` on a `Module` instance, since it doesn't make sense to "copy and assign" like we can for `Logic`s.  Instance names of modules have to be defined per-module.  We could perhaps add some method for overriding instance names, but I'm not sure that's necessary or sufficiently useful?

> - Is there any way to pickup the name of the Dart variable. If so, we can use decent names for Dart variables and get them for free in the Verilog (and guaranteed to match). Then we would only used .named() for the merging options or if we have a Verilog conflict or want a different naming convention in Verilog than in Dart. I find I only use the merging options in detailed components, but not in larger blocks where the names are generally unique.

This would bring reflection into ROHD, which implies the structure and naming of the generator software has an impact on the generated hardware (or at least its naming).  Philosophically, this has been a strength of ROHD vs. other new HDL approaches like Chisel.  I think the Dart macros feature will provide a good mechanism in the future for saving on redundancy for naming things identically without forcing reflection more broadly.

PR #561, opened 2025-02-06
Bug fix for withSet in LogicStructure.

kimmeljo:
## Description & Motivation

Bug in LogicStructure's withSet method where if the "update" value's width is less than the given overlapping field's width within the structure (per the start index of the set), build would crash because the out of bounds guard on the "update" value was incorrectly ignoring the "update" field's width and instead just using the overlapping field's width.

## Related Issue(s)

N/A

## Testing

Added a single unit test to logic_structure_test that exercises the bug.

## Backwards-compatibility

Backwards compatible.

## Documentation

No need for doc.

### DIFF
--- a/lib/src/signals/logic_structure.dart
+++ b/lib/src/signals/logic_structure.dart
@@ -1,4 +1,4 @@
-// Copyright (C) 2023-2024 Intel Corporation
+// Copyright (C) 2024-2025 Intel Corporation
 // SPDX-License-Identifier: BSD-3-Clause
 //
 // logic_structure.dart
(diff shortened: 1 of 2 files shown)


lib/src/signals/logic_structure.dart
mkorbel1, line 1: Can you update the copyright years in edited files?

mkorbel1 (commented):
Looks great, thank you!

Can you add a brief summary of the bug in the PR description? It helps for paper trail of bugs and their fixes and writing the CHANGELOG

test/logic_structure_test.dart
```diff
@@ -1,4 +1,4 @@
-// Copyright (C) 2023 Intel Corporation
+// Copyright (C) 2024-2025 Intel Corporation
```
mkorbel1, line 1: nit: 2023-2025 (on both of these files)

mkorbel1 (approved):
Great!

Issue #506, opened 2024-09-11
Deprecate `Port`, move to `Logic.port` for API consistency


mkorbel1:
### Motivation

The "port" style constructors include:
```
Logic --> Port
LogicArray --> LogicArray.port
LogicNet --> LogicNet.port
LogicArray.net --> LogicArray.netPort
```

One of these is not like the others.  Moving to `Logic.port` would be more verbose, but create a more consistent API.

### Desired solution

Deprecate `Port` in favor of `Logic.port`.

lucasphillips:
As part of the 2025 OSS Hackathon, ROHD_Team1 is working on this issue.

PR #575, opened 2025-03-04
Deprecate `Port` and move to `Logic.port` for API consistency

lucasphillips:
## Description & Motivation

Deprecate `Port` and replace it with a new `Logic.port` constructor with the same functionality to make the API more consistent. Other classes such as `LogicArray` and `LogicNet` already have this style of port constructor.

## Related Issue(s)

#506 

## Testing

Updated existing tests to use `Logic.port`

## Backwards-compatibility

No. The `Port` class is deprecated but still present.

## Documentation

Updated documentation to use the new `Logic.port` constructor instead of `Port`

### DIFF
--- a/doc/user_guide/_docs/A11-interfaces.md
+++ b/doc/user_guide/_docs/A11-interfaces.md
@@ -10,7 +10,7 @@ Interfaces make it easier to define port connections of a module in a reusable w
 
 [`Interface`](https://intel.github.io/rohd/rohd/Interface-class.html) takes a generic parameter for direction type.  This enables you to group signals so make adding them as inputs/outputs easier for different modules sharing this interface.
 
-The [`Logic.port`](https://intel.github.io/rohd/rohd/Logic/Logic.port.html) facory returns `Logic`, but has a constructor that takes width as a positional argument to make interface port definitions a little cleaner.
+The [`Logic.port`](https://intel.github.io/rohd/rohd/Logic-class.html) constructor makes interface port definitions a little cleaner by taking the width as a positional argument.
 
 When connecting an `Interface` to a `Module`, you should always create a new instance of the `Interface` so you don't modify the one being passed in through the constructor.  Modifying the same `Interface` as was passed would have negative consequences if multiple `Module`s were consuming the same `Interface`, and also breaks the rules for `Module` input and output connectivity.
 
--- a/doc/user_guide/_docs/A11-interfaces.md
+++ b/doc/user_guide/_docs/A11-interfaces.md
@@ -10,7 +10,7 @@ Interfaces make it easier to define port connections of a module in a reusable w
 
 [`Interface`](https://intel.github.io/rohd/rohd/Interface-class.html) takes a generic parameter for direction type.  This enables you to group signals so make adding them as inputs/outputs easier for different modules sharing this interface.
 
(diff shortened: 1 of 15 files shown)


lib/src/signals/port.dart
mkorbel1, line 1: Be sure to update the copyright year in the header at the top of files modified (e.g. from `2021-2023` to `2021-2025`)

doc/user_guide/_docs/A11-interfaces.md
```diff
@@ -10,7 +10,7 @@ Interfaces make it easier to define port connections of a module in a reusable w
 
 [`Interface`](https://intel.github.io/rohd/rohd/Interface-class.html) takes a generic parameter for direction type.  This enables you to group signals so make adding them as inputs/outputs easier for different modules sharing this interface.
 
-The [`Port`](https://intel.github.io/rohd/rohd/Port-class.html) class extends `Logic`, but has a constructor that takes width as a positional argument to make interface port definitions a little cleaner.
+The [`Logic.port`](https://intel.github.io/rohd/rohd/Logic/Logic.port.html) facory returns `Logic`, but has a constructor that takes width as a positional argument to make interface port definitions a little cleaner.
```
mkorbel1, line 13: nit: a factory constructor is a constructor, so "has a constructor" doesn't quite make sense in this sentence (similar sentence exists in the tutorial)

lib/src/interfaces/interface.dart
```diff
@@ -25,13 +25,14 @@ import 'package:rohd/rohd.dart';
 /// [Module] input and output connectivity.
 class Interface<TagType> {
   /// Internal map from the [Interface]'s defined port name to an instance
-  /// of a [Port].
+  /// of a [Logic.port].
```
mkorbel1, line 28: nit: technically, it's just an instance of a `Logic` (and should have always said that, in fact)

mkorbel1 (changes requested):
Looks great!  Just a few minor things

doc/user_guide/_docs/A11-interfaces.md
curtisan-intel: Would something like this be better?

[`Logic`](https://intel.github.io/rohd/rohd/Logic-class.html)s can also be ports by using the `Logic.port` constructor, which takes width as a positional argument to make interface port definitions a little cleaner.

mkorbel1: Maybe this?

> The `Logic.port` constructor makes interface port definitions a little cleaner by taking the width as a positional argument.

All ports are `Logic`s

lib/src/interfaces/interface.dart
lucasphillips: Changed to `Logic`

lib/src/signals/port.dart
lucasphillips: Done.

doc/user_guide/_docs/A11-interfaces.md
lucasphillips: We used your suggestion verbatim, since we had no better ideas.

Also changed the link to Logic-class.html instead of Logic.port.html to keep some kind of useful link while avoiding the markdown check issue.

mkorbel1 (approved):
Looks great!  Will merge it in once latest commit passes all CI checks, awesome job!

awmoore-intel:
Should there be a `LogicStructure.port` too?

Issue #582, opened 2025-03-07
radixString should have an option to produce fixed-length strings


desmonddak:
### Motivation

In some formatting cases, a fixed length string makes sense for alignment or for future parsing.

So instead of 16h'1,  we may need to have preceding zeros to pad the string like 16h'0001.
Furthermore, this may allow for printing just the raw data in a string like "0001" when you know it will be parsed with a radix argument.  This creates a situation where output can be aligned independent of value.
For binary, this is often the case where numeric values (mantissas) are examined in aligned form.

### Desired solution

Modify LogicValue.toRadixString to have an option that allows for preceding zero padding.
LogicValue.ofRadixString can already process padded strings.

PR #583, opened 2025-03-07
Enhance LogicValue.toRadixString to enable fixed-width output

desmonddak:
## Description & Motivation

Enable printing of leading zeros to get fixed-width output to help with alignment during printing and for parsing
from strings without having to use the size prefix, just passing the radix along with a set of strings.

## Related Issue(s)

#582 

## Testing

Added tests that enabled the leadingZero option and checked that they match as well as not breaking existing tests.

## Backwards-compatibility

Yes.  This option is off by default.

## Documentation

Yes.  An example was added to one of the user guide sections and the API for the method was updated.

### DIFF
--- a/lib/src/values/logic_value.dart
+++ b/lib/src/values/logic_value.dart
@@ -721,8 +721,8 @@ abstract class LogicValue implements Comparable<LogicValue> {
         buf.write(_reverse(s));
       }
       reversedStr = _reverse(buf.toString());
-      // }
     }
+
     final spaceString = _reverse(reversedStr
         .replaceAllMapped(
             RegExp('((>(.){$chunkSize}<)|([a-zA-Z0-9])){$chunkSize}'),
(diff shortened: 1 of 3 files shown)


desmonddak:
@mkorbel1 I think this may need review.

lib/src/values/logic_value.dart
```diff
@@ -695,9 +712,16 @@ abstract class LogicValue implements Comparable<LogicValue> {
...
+        if (s != '0') {
+          haveLeadingZeros = false;
+        }
+        if ((s == '0') & !leadingZeros & haveLeadingZeros) {
+          continue;
+        }
         buf.write(_reverse(s));
       }
       reversedStr = _reverse(buf.toString());
+      // }
```
mkorbel1, line 724: left over commented code, probably meant to remove this?

test/logic_value_test.dart
```diff
@@ -2052,11 +2052,30 @@ void main() {
   group('RadixString', () {
     test('radixString roundTrip', () {
       final lv = LogicValue.ofBigInt(BigInt.from(737481838713847), 61);
+      LogicValue.ofRadixString(lv.toRadixString());
```
mkorbel1, line 2055: what's the point of these new lines?  aren't they redundant with the loop below?

mkorbel1 (commented):
Looks like a nice addition!

test/logic_value_test.dart
desmonddak: Those were for debug and will be deleted.

desmonddak:
I think this is ready to merge.

mkorbel1 (approved):
Looks great!

Issue #559, opened 2025-01-29
Make assignments between packed LogicArrays and Logics in generated SystemVerilog avoid a swizzle


mkorbel1:
### Motivation

Sometimes there are assignments between a `LogicArray` with 1 dimension (packed), and a `Logic` of the same width.  It would be nice to generate that SystemVerilog as a simple assignment instead of first going through a swizzle conversion.

### Desired solution

Somehow transform assignments in generated SystemVerilog to avoid unnecessary swizzling prior to assignments between 1-D packed `LogicArray`s and `Logic`s.

PR #584, opened 2025-03-08
Adding mechanisms to avoid unnecessary swizzle conversions.

gus-bonilla-86:
## Description & Motivation

The use of swizzle conversions could be unnecessary in the generated SystemVerilog code when there are assignments between a LogicArray with 1 dimension (packed), and a Logic of the same width. This changes add mechanisms to identify those cases and replace the swizzle conversions with simple assignments.

## Related Issue(s)

#559 

## Testing

A unit testing script was created to validate the [SystemVerilogSwizzleOptimizer] functionality.
rohd/test/swizzle_opt_test.dart

## Backwards-compatibility

No.

## Documentation

The [SystemVerilogSwizzleOptimizer] class was added with its code documentation, no need to add more.
rohd/lib/src/utilities/swizzle_opt.dart

### DIFF
--- a/lib/src/module.dart
+++ b/lib/src/module.dart
@@ -19,6 +19,7 @@ import 'package:rohd/src/utilities/config.dart';
 import 'package:rohd/src/utilities/sanitizer.dart';
 import 'package:rohd/src/utilities/timestamper.dart';
 import 'package:rohd/src/utilities/uniquifier.dart';
+import 'package:rohd/src/utilities/swizzle_opt.dart';
 
 /// Represents a synthesizable hardware entity with clearly defined interface
 /// boundaries.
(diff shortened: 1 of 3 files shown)


mkorbel1 (commented):
I'll take a deeper look through this as soon as possible, but at first review, I'm a little concerned about the approach of post-processing generated SystemVerilog with regular expressions rather than modifying the generation mechanism before SystemVerilog is generated.

For example, the `_collapseArrays` flow in the `SystemVerilogSynthesizer` stack handles collapsing of assignments more generically by reference before SystemVerilog is emitted:
https://github.com/intel/rohd/blob/44139254838b0a2b3c80b0efa7d2914477410654/lib/src/synthesizers/systemverilog.dart#L1284

Is there a reasonable way to transform your algorithm to apply more generically pre-SV generation?  My intuition is that it would become more robust and less brittle to specific SV generation styling, since these regular expressions are not as robust as a full SystemVerilog parser anyways.

Another consideration is that generated SV may contain "custom" SystemVerilog modules written by users, and we wouldn't want to modify those at all.

mkorbel1:
I think this PR is not on a trend where it will be merged into ROHD, so I'm closing it.  Thank you for the time you've spent on this, and the original issue is still worth addressing, but this approach isn't quite right.  If you'd like to continue on this development with an adjusted approach, we can either re-open this PR or open a new one.

Issue #146, opened 2022-08-11
Support tagging of generated SystemVerilog `begin` and `end`


mkorbel1:
## Motivation
Some users have extra preferences about the readability of generated SystemVerilog for their applications.  It would be nice to use the tagging mechanism for if, case, etc. in SV to help along those lines.

## Desired solution
Every ‘begin’ should have the ability to add a name for it.

```Verilog
if (a) begin : if_name
   .
   .
end else begin : else_name
   .
   .
end // if_name
```

This should be simple to add to things like `Iff`, `ElseIf`, `Else`, as well as `If` and `Case`.

mkorbel1:
Some lint tool checks need this feature, so there is some urgency to get this implemented to make life easier for users producing SystemVerilog that passes through those tools.

lucasphillips:
As part of the 2025 OSS Hackathon, ROHD_Team1 is working on this issue.

mkorbel1:
Awesome, assigned!

PR #585, opened 2025-03-08
SystemVerilog block labels

lucasphillips:
## Description & Motivation

For ROHD classes which use begin/end to define blocks in generated SystemVerilog, support tagging the block with an optional label.

This is helpful to generate SystemVerilog which passes certain strict linting checks, and for compatibility with compilers which require labeled blocks.

## Related Issue(s)

#146

## Testing

Added a new set of tests which generate different types of blocks with labels and compile/compare the generated SystemVerilog.

## Backwards-compatibility

No. The new parameters are optional.

## Documentation

Yes. Requires description of new parameters and explanation of the feature.

### DIFF
--- a/lib/src/modules/conditionals/always.dart
+++ b/lib/src/modules/conditionals/always.dart
@@ -21,6 +21,9 @@ abstract class Always extends Module with SystemVerilog {
       UnmodifiableListView<Conditional>(_conditionals);
   List<Conditional> _conditionals;
 
+  /// Optional block label
+  final String? label;
+
   /// A mapping from internal receiver signals to designated [Module] outputs.
   @protected
   @internal
@@ -47,7 +50,10 @@ abstract class Always extends Module with SystemVerilog {
   /// driven by any other [Conditional] in this block, it will be driven to the
   /// specified reset value.
   Always(this._conditionals,
-      {Logic? reset, Map<Logic, dynamic>? resetValues, super.name = 'always'}) {
+      {Logic? reset,
+      Map<Logic, dynamic>? resetValues,
+      super.name = 'always',
+      this.label}) {
     // create a registration of all inputs and outputs of this module
     var idx = 0;
 
@@ -105,8 +111,10 @@ abstract class Always extends Module with SystemVerilog {
           reset,
           // then use it for assigning receiver
           then: allResetCondAssigns,
+          ifLabel: '${label}_if',
           // else assign zero as resetValue
           orElse: conditionals,
+          elseLabel: '${label}_else',
         ),
       ];
     }
@@ -176,12 +184,14 @@ abstract class Always extends Module with SystemVerilog {
         ports.entries.where((element) => this.inputs.containsKey(element.key)));
     final outputs = Map.fromEntries(ports.entries
         .where((element) => this.outputs.containsKey(element.key)));
+    final blockLabel =
+        label == null ? '' : ' : ${Sanitizer.sanitizeSV(label!)}';
 
     var verilog = '';
     verilog += '//  $instanceName\n';
-    verilog += '${alwaysVerilogStatement(inputs)} begin\n';
+    verilog += '${alwaysVerilogStatement(inputs)} begin$blockLabel\n';
     verilog += _alwaysContents(inputs, outputs, assignOperator());
-    verilog += 'end\n';
+    verilog += 'end$blockLabel\n';
     return verilog;
   }
 }
--- a/lib/src/modules/conditionals/case.dart
+++ b/lib/src/modules/conditionals/case.dart
@@ -11,6 +11,7 @@ import 'package:collection/collection.dart';
 import 'package:meta/meta.dart';
 import 'package:rohd/rohd.dart';
 import 'package:rohd/src/modules/conditionals/ssa.dart';
+import 'package:rohd/src/utilities/sanitizer.dart';
 
 /// Represents a single case within a [Case] block.
 class CaseItem {
@@ -20,8 +21,11 @@ class CaseItem {
   /// A [List] of [Conditional]s to execute when [value] is matched.
   final List<Conditional> then;
 
+  /// An optional label for this case body.
+  final String? label;
+
   /// Executes [then] when [value] matches.
-  CaseItem(this.value, this.then);
+  CaseItem(this.value, this.then, {this.label});
 
   @override
   String toString() => '$value : $then';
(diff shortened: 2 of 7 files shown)


lucasphillips:
Once we are confident that we have the right approach to implement this feature, we will need to add some general documentation explaining what it does. Right now the only documentation is the descriptions of the new parameters.

lib/src/modules/conditionals/case.dart
```diff
@@ -119,6 +127,9 @@ class Case extends Conditional {
   List<Conditional>? get defaultItem => _defaultItem;
   List<Conditional>? _defaultItem;
 
+  /// Optional label for the default case block.
+  final String? defaultLabel;
```
mkorbel1, line 131: idea: what if we accepted a `name` instead, and used it to construct the default label, something like `default_$name`?  not sure if that's better or worse.  thoughts?

lib/src/modules/conditionals/flop.dart
```diff
@@ -146,7 +152,9 @@ class FlipFlop extends Module with SystemVerilog {
     var contents = [q < _d];
 
     if (_en != null) {
-      contents = [If(_en!, then: contents)];
+      contents = [
+        If(_en!, then: contents, ifLabel: label == null ? null : '${label}_en')
```
mkorbel1, line 156: since `flop` uses custom `instantiationVerilog`, i think this label addition here actually has no effect.  there's no begin/end blocks at all for this one-liner in generated SV.

lib/src/modules/conditionals/if.dart
```diff
@@ -82,10 +88,14 @@ class If extends Conditional {
 
   /// If [condition] is high, then [then] executes, otherwise [orElse] is
   /// executed.
-  If(Logic condition, {List<Conditional>? then, List<Conditional>? orElse})
+  If(Logic condition,
+      {List<Conditional>? then,
+      List<Conditional>? orElse,
+      String? ifLabel,
```
mkorbel1, line 94: similar to prior comments: what if we just accepted one `label`/`name` and named the blocks `if_$name` and `else_$name` or something automatically?  or do you think there's enough demand likely for people to customize the if and else block names separately?  and if so, maybe it's ok to expect them to use the more verbose version of `If.block`?

test/block_labels_test.dart
```diff
@@ -0,0 +1,393 @@
...
+class IncrModule extends Module {
+  Logic get result => output('result');
+  IncrModule(Logic toIncr) : super(name: 'incr') {
+    toIncr = addInput('toIncr', toIncr, width: toIncr.width);
+    addOutput('result', width: toIncr.width);
+    result <= toIncr + 1;
+  }
+}
+
+void main() {
```
mkorbel1, line 204: we should add some tests that actually generate SV and confirm some substring is inside, e.g.

```dart
final sv = mod.generateSynth();
expect(sv, contains('if(thing) : my_label');
```

test/block_labels_test.dart
```diff
@@ -0,0 +1,393 @@
...
+class LabeledMultiBlockModule extends Module {
+  Logic a;
+  Logic b;
+  Logic c;
+  Logic d;
+
+  Logic get out_0 => output('out_0');
+  Logic get out_1 => output('out_1');
+
+  String firstBlockLabel;
```
mkorbel1, line 145: we should test that when *no* label is provided, the output SV looks right.  could make these `String?` (nullable) and have some tests which set them to `null`, then confirm compilation and correct verilog generation (you could even loop through and run all your tests with both a string and null provided, if you want)

lib/src/modules/conditionals/always.dart
```diff
@@ -176,12 +184,14 @@ abstract class Always extends Module with SystemVerilog {
         ports.entries.where((element) => this.inputs.containsKey(element.key)));
     final outputs = Map.fromEntries(ports.entries
         .where((element) => this.outputs.containsKey(element.key)));
+    final blockLabel =
+        label == null ? '' : ' : ${Sanitizer.sanitizeSV(label!)}';
 
     var verilog = '';
     verilog += '//  $instanceName\n';
-    verilog += '${alwaysVerilogStatement(inputs)} begin\n';
+    verilog += '${alwaysVerilogStatement(inputs)} begin$blockLabel\n';
```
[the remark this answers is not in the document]
mkorbel1: thanks for finding those sections -- my read of this is that they *do* need to be unique?  That would, unfortunately, mean we need to probably uniquify them during SV generation in the `Synthesizer` stack.  We'll have to think about whether that introduces a non-backwards-compatible change or if we can avoid it, as well.  We could also experiment with violating the rule in multiple simulators and linters and seeing if our interpretation is correct or not.

mkorbel1:
Hi @lucasphillips @curtisan-intel @capp21 there's still interest in this feature, and you have made a great start here!  Are you still interested in taking this forward to a merge?  Otherwise, we might take this as a starting point and build it out from here

lib/src/modules/conditionals/always.dart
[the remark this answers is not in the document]
mkorbel1: so reading into it and thinking a little more, since each block scope is a new scope, the naming only needs to be uniquified locally within that scope!  that means we can use the `name` for top-level `always` blocks, since that's already unique, then we can handle uniquification at `Conditional` levels each step through the `verilogContents` API, which is much less disruptive for uniquification!  I think that works

capp21:
> Hi @lucasphillips @curtisan-intel @capp21 there's still interest in this feature, and you have made a great start here! Are you still interested in taking this forward to a merge? Otherwise, we might take this as a starting point and build it out from here

I'm still interested! I need to dedicate some time to complete this.

mkorbel1:
Great!  Let me know on what timeline you think we can close this.  Happy to collaborate!

I'd like to use it for #593, but depending on timeline may push that one first with this one following.

curtisan-intel:
Hi @mkorbel1, myself and @sweeksBigBlue and @jcfarwe are going to look into this over the next week or two and try to get it finished up.

lib/src/modules/conditionals/always.dart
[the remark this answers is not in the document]
sweeksBigBlue: @mkorbel1 we can look into this, but maybe this is better as a different issue we can address? I'm not seeing a clean way of addressing this that doesn't make this an unwieldly PR. Not exactly familiar enough with how ROHD is generating blocks to say I have a good recursive solution for this. We could do a list and lookup for current scope but I think that is a substantial change to the code and would maybe make code review less reasonable (I'm also not a dart expert and still learning so I'm trying to keep my lack of current knowledge in mind).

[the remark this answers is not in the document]
mkorbel1: maybe we can meet to sketch out what it looks like? i think once we figure out the pattern it will end up being relatively simple and not too disruptive to apply it

Issue #552, opened 2025-01-21
Omit redundant parentheses in generated SystemVerilog


mkorbel1:
### Motivation

There are a handful of cases where generated SystemVerilog has too many parentheses.  Currently, they are added conservatively for safety, to ensure order of operations and evaluation matches the generated intent.  With some knowledge about the context in which an expression is placed, it should be possible to omit (at least some) unnecessary parentheses.

A couple examples:

```SystemVerilog
// module instantation
my_mod inst1(
  .a_0((a[0]))
);
```

```SystemVerilog
// chained operators
assign a = (b & (c & (d & e)));
```

It's important to maintain parentheses sometimes, especially when it helps disambiguate order of operations to the user.  Maybe, for example, keep `&` and `|` terms separated  by parentheses?  This is a bit subjective -- perhaps there exist some other recommendations in other languages/contexts.

### Desired solution

Omit "unnecessary" parentheses in generated SystemVerilog

gus-bonilla-86:
Hi, we would like to work on this issue [ROHD_Team2] [Intel Open Source Software Hackathon 2025].

mkorbel1:
Nice!  Assigned!

PR #586, opened 2025-03-08
Adding mechanisms to remove redundant parentheses.

gus-bonilla-86:
## Description & Motivation

There are cases where generated SystemVerilog code has too many parentheses to increase safety, ensure order of operations and matching the generated intent. This changes add mechanisms to identify redundancies with parenthesis and remove them to simplify the generated code.

## Related Issue(s)

#552 

## Testing

A unit testing script was created to validate the [RedundancyHandler] functionality.
rohd/test/redundancy_handler_test.dart

## Backwards-compatibility

No.

## Documentation

The [RedundancyHandler] class was added with its code documentation, no need to add more.
rohd/lib/src/utilities/redundancy_handler.dart

### DIFF
--- a/lib/src/module.dart
+++ b/lib/src/module.dart
@@ -11,11 +11,11 @@ import 'dart:async';
 import 'dart:collection';
 
 import 'package:meta/meta.dart';
-
 import 'package:rohd/rohd.dart';
 import 'package:rohd/src/collections/traverseable_collection.dart';
 import 'package:rohd/src/diagnostics/inspector_service.dart';
 import 'package:rohd/src/utilities/config.dart';
+import 'package:rohd/src/utilities/redundancy_handler.dart';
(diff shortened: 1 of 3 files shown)


mkorbel1 (commented):
Similar comments as #584, will review in more detail soon when I can

mkorbel1 (changes requested):
See notes in https://github.com/intel/rohd/pull/584#pullrequestreview-2671906722, which applies to this PR as well

mkorbel1:
I think this PR is not on a trend where it will be merged into ROHD, so I'm closing it.  Thank you for the time you've spent on this, and the original issue is still worth addressing, but this approach isn't quite right.  If you'd like to continue on this development with an adjusted approach, we can either re-open this PR or open a new one.

PR #606, opened 2025-07-11
bug fix for zero value toRadixString()

### LINKED ISSUE #605: toRadixString() fails on LogicValue.zero
### Describe the bug

By trimming 0s from the output string, toRadixString() forgets to add back a single '0' when the value is actually zero.

### To Reproduce

 LogicValue.zero.zeroExtend(32).toRadixString() fails with a bad array access.

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

desmonddak:
## Description & Motivation

Make sure to check if computed radixString is of zero length and add back a single leading zero.

## Related Issue(s)

Fixes #605 

## Testing

Added a test for converting zero with and without leading zeros flag using different radices.

## Backwards-compatibility

No.

## Documentation

None needed.

### DIFF
--- a/lib/src/values/logic_value.dart
+++ b/lib/src/values/logic_value.dart
@@ -729,9 +729,11 @@ abstract class LogicValue implements Comparable<LogicValue> {
             (match) => '${match.group(0)}$sepChar')
         .replaceAll('$sepChar<', '<'));
 
-    final fullString = spaceString[0] == sepChar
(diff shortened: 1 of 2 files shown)


mkorbel1 (approved):
Looks good!

PR #614, opened 2025-07-24
Cloning and matching

mkorbel1:
## Description & Motivation

This PR introduces a couple ideas:
- Requiring a `clone` to be included on `Logic`s and `Interface`s
- Adding APIs to `Module` that take advantage of those capabilities

The `addMatched*` functions on `Module` are parameterized and return the same datatype as was received.  Additionally, `Module` now supports `LogicStructure`s of any type as ports, though the SV still packs them on the interface.  These functions are also useful for creating ports when you want the widths/dimensions to match the original argument.

There's also some `connect*` functions for `Interfaces`, which basically just wrap the more common way to create interfaces by calling `clone` and `connectIO`.  There's one for arbitrary `Interface`s and one for `PairInterface`s.

## Related Issue(s)

N/A

## Testing

Added new testing for new functionality.

## Backwards-compatibility

No, though there's new expectations for any implemented `Interface`s and `LogicStructure`s if they wish to properly handle these APIs.

## Documentation

Yes, API docs included.  User guide updates WIP.

### DIFF
--- a/lib/src/module.dart
+++ b/lib/src/module.dart
@@ -684,14 +684,15 @@ abstract class Module {
     // TODO: confirm that no matched input/output structure has ANY nets in it
     //  or we will violate all kinds of assumptions
     if (source.isNet || (source is LogicStructure && source.hasNets)) {
-      //TODO
-      throw Exception('Matched inputs cannot have nets in them.');
+      throw PortTypeException(
+          source, 'Matched inputs cannot have nets in them.');
     }
 
     final inPort = (source.clone(name: name) as LogicType)..gets(source);
 
     if (inPort.name != name) {
-      throw Exception('clone name failed'); // TODO
+      throw PortTypeException.forIntendedName(name,
+          'The `clone` method for $source failed to update the signal name.');
     }
 
     if (inPort is LogicStructure) {
@@ -768,7 +769,7 @@ abstract class Module {
     _checkForSafePortName(name);
 
     if (!source.isNet) {
-      throw Exception('Matched inOuts must be nets.');
+      throw PortTypeException(source, 'Matched inOuts must be nets.');
     }
 
     _inOutDrivers.add(source);
@@ -793,7 +794,8 @@ abstract class Module {
     final inOutPort = (source.clone(name: name) as LogicType)..gets(source);
 
     if (inOutPort.name != name) {
-      throw Exception('clone name failed'); // TODO
+      throw PortTypeException.forIntendedName(name,
+          'The `clone` method for $source failed to update the signal name.');
     }
 
     if (inOutPort is LogicStructure) {
@@ -865,15 +867,16 @@ abstract class Module {
     _checkForSafePortName(name);
 
     if (toMatch.isNet || (toMatch is LogicStructure && toMatch.hasNets)) {
-      //TODO
-      throw Exception('Matched outputs cannot have nets in them.');
+      throw PortTypeException(
+          toMatch, 'Matched outputs cannot have nets in them.');
     }
 
     // must make a new clone of it, to avoid people using ports of other modules
     final outPort = toMatch.clone(name: name) as LogicType;
 
     if (outPort.name != name) {
-      throw Exception('clone name failed'); // TODO
+      throw PortTypeException.forIntendedName(name,
+          'The `clone` method for $toMatch failed to update the signal name.');
     }
 
     if (outPort is LogicStructure) {
--- a/lib/src/module.dart
+++ b/lib/src/module.dart
@@ -55,6 +55,13 @@ abstract class Module {
   /// An internal mapping of inOut names to their sources to this [Module].
   late final Map<String, Logic> _inOutSources = {};
 
+  // late final Map<String, LogicStructure> _inputStructures = {};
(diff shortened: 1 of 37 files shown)


doc/user_guide/_docs/A08-modules.md
```diff
@@ -2,14 +2,14 @@
...
-2. Outputs (from `output` or `addOutput`, or their array equivalents) are the only way logic outside of a `Module` can consume signals from that `Module`.  There are no internal vs. external versions of `output`s, so they may be consumed inside of `Module`s as well.
+1. Inputs and inOuts (from `input`, `addInput`, `inOut` or `addInOut` methods, or their array and match equivalents) return *internal* copies of ports that should be used inside a `Module`.  Signals should not be consumed directly from outside a `Module`.  Internal module logic should consume the internal versions.  Logic outside the `Module` can drive to (or receive from, in the case of inOut) that `Module` only through the external copies, i.e. arguments passed to `addInput`, `addInOut`, etc.
+2. Outputs (from `output` or `addOutput`, or their array and match equivalents) are the only way logic outside of a `Module` can consume signals from that `Module`.  There are no internal vs. external versions of `output`s, so they may be consumed inside of `Module`s as well.
```
desmonddak, line 12: i think you should spell out the match equivalents here.

doc/user_guide/_docs/A08-modules.md
```diff
@@ -50,9 +50,11 @@ class MyModule extends Module {
 
 All gates or functionality apart from assign statements in ROHD are implemented using Modules.
 
-#### Inputs, outputs, widths, and getters
+#### Ports, widths, and getters
 
-The default width of an input and output is 1.  You can control the width of ports using the `width` argument of `addInput()` and `addOutput()`.  You may choose to set them to a static number, based on some other variable, or even dynamically based on the width of input parameters.  These functions also return the input/output signal.
+The default width of a port is 1.  You can control the width of ports using the `width` argument of `addInput()` and `addOutput()`.  You may choose to set them to a static number, based on some other variable, or even dynamically based on the width of input parameters.  These functions also return the input/output signal.
+
+There are also similar functions called `addMatchedInput` and `addMatchedOutput` which will create a port with matching widths and types.  This is especially useful for creating `LogicStructure` ports.
```
desmonddak, line 57: What about Arrays?  Should these all be listed in one place so people know the variety and can itemize them?

lib/src/module.dart
```diff
@@ -741,6 +767,64 @@ abstract class Module {
     return inOutPort;
   }
 
+  /// Registers a signal as an [inOut] to this [Module] and returns an inOut
```
desmonddak, line 770: nit: second inOut -- do you want to bracket it, or define it as input/output

lib/src/modules/bus.dart
```diff
@@ -23,9 +23,11 @@ class BusSubset extends Module with InlineSystemVerilog {
   late final String _subsetName;
 
   /// The input to get a subset of.
-  late final Logic _original;
+  ///
+  /// This should not be used outside of this module.
+  late final Logic original;
```
desmonddak, line 28: Then why is it not private if it should not be used outside?

desmonddak (commented):
Looks good and I tried it on FPConverter which dramatically simplified the IO handling of a LogicStructure.  This will apply to FPAdder  FPMultiplier, the two square root components, and the Fixed2Float, Float2Fixed.

doc/tutorials/chapter_8/01_interface.md
```diff
@@ -21,7 +21,9 @@ The `Logic.port` constructor makes interface port definitions a little cleaner b
```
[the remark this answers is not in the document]
mkorbel1: good catch, thank you!

lib/src/module.dart
mkorbel1: thanks fixed here and in multiple places

lib/src/modules/bus.dart
mkorbel1: I updated the message here, but the idea is that it's useful to know the input and/or input name for these modules, even though you shouldn't connect to them.

lib/src/module.dart
```diff
@@ -874,6 +991,29 @@ abstract class Module {
     return inOutArr;
   }
 
+  /// Connects the [source] to this [Module] using [Interface.connectIO] and
```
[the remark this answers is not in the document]
desmonddak: I just want to know as a user when I should be calling connectIO versus connectInterface.

mkorbel1:
Thank you for reviewing, @desmonddak !

desmonddak:
This looks like a major cleanup.  I even see a LogicStructure naming change that will help SV output (I noticed with Fixed2Float, I think  I can verify it and remove a named in the component).

desmonddak:
So I notice a change from `addMatchedOutput` to `addTypedOutput` in that the latter is unhappy with the passing in of a FixedPoint as second argument.

desmonddak:
Ahh.  This is where a clone function is passed.

desmonddak:
New changes look good.

PR #639, opened 2025-12-18
SchematicSynthesizer

### LINKED ISSUE #162: Offline schematic generation
### Motivation

We have a mechanism for creating HTML-based schematics for our components once they are in our web page environment.

What would be useful during component development is to have a schematic generator service like our Waveform generator that can generate schematics.

All the tooling for doing this is in our tools area. What may be tricky is how to load up a web page viewer to point to the schematics as the area would need the d3schematic collateral there as well. This is currently installed by our tools/ infrastructure to be served as part of documentation, this enhancement would allow for a person developing a component to create a schematic of their component locally and use it for visualization.

It may be possible to use the schematic generator for larger designs, though at some point we would need to turn to more robust solutions for schematic generation, but for component development, the d3schematic service could be quite serviceable.

### Desired solution

Create a new class, say SchematicGenerator, which takes a Module as argument and builds a schematic html + d3 collateral.  The user would simply open a web page at this filesystem area and be able to visualize their design.

The tooling to do this is in the tools/ area as we use d3schematic generation for our online documentation.

### Alternatives considered

_No response_

### Additional details

_No response_

desmonddak:
## Description & Motivation

This PR adds a `Synthesizer` for D3 schematics that can be loaded using an ELK-JSON loader.
This produces the Yosys JSON format.

## Related Issue(s)

This fixes https://github.com/intel/rohd-hcl/issues/162, but moves it into the rohd repository as a general purpose capability for ROHD.

## Testing

We added a basic yosys JSON loader borrowed and documented from d3-hwschematics (<https://github.com/Nic30/d3-hwschematic>.  We ran a few complex examples with multiple instantiation.

## Backwards-compatibility

No.  We added a prepare() routine for `SynthBuilder` which is optional and helpful for this and future `Synthesizer`s.

## Documentation

Just a simple one-liner to parallel the `SystemVerilogSynthesizer`.  We probably need to make a concerted effort to document more of the utilities in ROHD.


mkorbel1 (commented):
Still WIP reviewing this, but posting current checkpoint

lib/src/synthesizers/schematic/constant_handler.dart
```diff
@@ -0,0 +1,420 @@
...
+  return true;
+}
+
+/// Information about a per-input constant driver.
+class PerInputConstInfo {
+  /// The list of internal net IDs driving the input.
+  final List<int> _ids;
+
+  /// The constant value being driven.
+  final BigInt _value;
```
mkorbel1: why is this `BigInt` instead of `LogicValue`?

desmonddak:
Replaced with synthesizers/netlist

Issue #270, opened 2023-02-13
Add automation for code coverage in PRs and pushes


mkorbel1:
### Motivation

Many repos have code coverage metrics provided in two places automatically:
- On the README as a badge, indicating coverage for the repo by the test suite.
- In PRs as an indication of how much coverage there is including the changes of the PR.
  - This can also sometimes include a change in coverage relative to the mainline.

Adding automation around these can have some nice benefits, for example:
- Demonstrate to users how well tested the repository is
- Identify testing gaps in new PRs more easily

It would be nice if ROHD (and related packages) had some automation around this.

### Desired solution

An automated coverage run that provides information about coverage in PRs and on the mainline.  It would be nice to generate the whole HTML report in a way that is easily inspectable.

### Alternatives considered

Some tools (even free ones) exist to do a lot of this, but using another software product integrated into the repository requires jumping through some additional hurdles.

Ideally, we are able to implement this without dependence on a 3rd party service.

chykon:
Can I take it? Are there any deadlines?

mkorbel1:
> Can I take it?

Sure, it's all yours, thanks!

> Are there any deadlines?

Nope, no firm deadline :)

mkorbel1:
Assigned, thank you @maifeeulasad !

maifeeulasad:
<img width="1244" height="704" alt="Image" src="https://github.com/user-attachments/assets/21bf6549-cafc-4f24-910a-d42733012981" />

Seems like the runner is stuck, maybe this kind of runner is available only to this repo. Or some other issues. It would be really wonderful, if we can resolve this in multiple PRs.

For the first one I am sending to render `main` branches coverage. Then we can do that for all PR, on every push make a comment from bot for current and main branch. Does this work? It will be more tested. What's your opinion on this?

mkorbel1:
@maifeeulasad  Oh yeah, that runner is a special self-hosted runner by Intel.  For local testing, you could change the runner to just `ubuntu-latest` and the free GitHub runners should pick it up I think.  Will review your PR, sorry for the delay!  Multiple PRs seems fine if they are well divided.  Main branch coverage first seems ok, then per-PR comments, yeah that sounds good!

maifeeulasad:
I can see that CI has failed due to some dependency issues. Raising a quick fix for that.

maifeeulasad:
Here is the update:

> The last PR for generating report for each push event is now working.

[...]

PR #646, opened 2026-01-20
Add automation for code coverage in pushes

maifeeulasad:
## Description & Motivation

partial solution of #270 

## Related Issue(s)
#270

## Testing

 - Ran the script locally
 - Couldnt run in my repo, due to some runner limitation I guess. It was stuck

## Backwards-compatibility
Completly, as it is new feature

Non breaking change, as it is a new feature, working on a completely different branch

## Documentation

Only one created script is documented inside, nothing more than that.


tool/gh_actions/generate_coverage.sh
```diff
@@ -0,0 +1,99 @@
...
+# 2026 January 20
+# Author: Maifee Ul Asad<maifeeulasad@gmail.com>
+
+set -euo pipefail
+
+# Remove old coverage data
+rm -rf coverage
+
+# Run tests with coverage
+dart test --coverage=coverage || true
```
mkorbel1: this runs coverage collection differently than the other script (https://github.com/intel/rohd/blob/main/tool/generate_coverage.sh), do you know if this approach is a newer/better way to collect coverage?  should we update the other script?  should these scripts be merged into one or one calls another instead of independent?

tool/gh_actions/generate_coverage.sh
```diff
@@ -0,0 +1,99 @@
...
+fi
+
+# Extract coverage percentage
+if command -v lcov &> /dev/null; then
+    SUMMARY=$(lcov --summary coverage/lcov.info 2>&1 | grep -E "lines\.*:")
+    PERCENT=$(echo "$SUMMARY" | grep -oP '\d+\.\d+' | head -1)
+    echo "Coverage: ${PERCENT}%"
+else
+    PERCENT="0.0"
+    echo "Coverage: 0.0%"
```
mkorbel1: would an error message be better there instead of just 0?

tool/gh_actions/generate_coverage.sh
```diff
@@ -0,0 +1,99 @@
...
+TEXT="${PERCENT}%"
+TEXT_LEN=${#TEXT}
+TEXT_WIDTH=$((TEXT_LEN * 63))
+LABEL_WIDTH=63
+VALUE_WIDTH=$((TEXT_WIDTH / 10 + 10))
+TOTAL_WIDTH=$((LABEL_WIDTH + VALUE_WIDTH))
+
+# Generate SVG badge
+# Style credit goes to shields.io
+cat > /tmp/coverage-badge.svg << EOFSVG
```
mkorbel1: question: why make our own SVG instead of using something like https://shields.io/ like we use for the README?

mkorbel1 (commented):
This looks great!  Do you want to try running this on your fork (update the runner so it uses the GitHub-hosted ones so it runs for you) to demo that it works?

tool/gh_actions/generate_coverage.sh
maifeeulasad: - it will push our github metrics up, ref: https://github.com/intel/rohd/graphs/traffic
 - i like keeping 3rd party as much out as possible, that's why i kept it out. But I don't don't have anything against shields or badge io
 - and if we use shields io, we need to keep push it to that branch, and write a railgruard to disable infinite action glitch
 
 
 but if you want i can make adjustment for that

tool/gh_actions/generate_coverage.sh
maifeeulasad: Care to suggest one please? I will update the PR accordingly

maifeeulasad:
@mkorbel1 I pushed to a different branch for testing and yes it is working as it should

 - this is the branch I tested with, https://github.com/maifeeulasad/rohd/tree/ci-repo-stat-maifee-test 
 - this is where the image got written, take a look at the filename please, it contains details regarding the branch name, https://github.com/maifeeulasad/rohd/blob/badges/coverage/ci-repo-stat-maifee-test.svg
 - you can find more details on the action we ran: https://github.com/maifeeulasad/rohd/actions/runs/21357565364/job/61468509731


So basically the strategy is to generate report for each PR for main and push to that `coverage` branch. And the main readme file will point to that svg somewhere and we will get a beautiful badge there. This is the strategy for main branch.

tool/gh_actions/generate_coverage.sh
maifeeulasad: - I forgot to generate branch coverage, fixing that right away
 - `dart pub global run` is deprecated, i think we should just use `dart run`, for our case `dart test` is just fine

I am willing to make modifications I just didn't know what that file was being used for that's why created this script completly bypassing that and generating svg along with it. Yeah, we can pass in some additional argument to tweak it's behavior. Wating to hear back from you!

ref: https://github.com/dart-lang/pub/issues/3066

tool/gh_actions/generate_coverage.sh
mkorbel1: Maybe "Error: coverage collection failed!" ?

tool/gh_actions/generate_coverage.sh
mkorbel1: Sure, if you're willing let's merge the scripts in a way that seems reasonable to you!  The other script was just for running coverage locally and viewing a report, so as long as the final solution (with either 1 or 2 scripts) supports both use cases (1. run locally and review in browser, 2. run as a github action for automated coverage collection), that should be good.  thanks for checking on deprecations!

tool/gh_actions/generate_coverage.sh
mkorbel1: ok, i was just curious, a custom SVG seems fine to me, thanks for the explanation!

mkorbel1 (commented):
It looks like you might need to waive the dead link error for the SVG since it's not generated yet -- when we do the next steps where its per-PR we can reenable it maybe

Thank you for the iterations!

tool/gh_actions/generate_coverage.sh
maifeeulasad: Done, both scripts are now merged, so all we need to do is pass `--generate-svg` in the old script and we will get the svg. Otherwise output are as identical as possible

tool/gh_actions/generate_coverage.sh
maifeeulasad: You got it. Now we have this custom error message!

maifeeulasad:
All done!

 - removed the link from svg from readme, as it is not generated yet
 - merged both scripts into one
 - added custom failure message

maifeeulasad:
bump!!

open for change requests

cc: @mkorbel1

mkorbel1:
Sorry for the delay on this!  Will review again shortly

desmonddak (approved):
I really want to try this out on newer code; this looks good.

mkorbel1 (approved):
Looks good, thank you very much for the contribution!

maifeeulasad:
Thanks @desmonddak @mkorbel1 :heart:

PR #650, opened 2026-04-16
Improvement in naming to match a future converged solution

### LINKED ISSUE #648: SV naming bug: `_findName` checks `Logic.name` instead of `preferredSynthName` for unpreferred classification
### Describe the bug

When the synthesizer picks a name for a merged signal group, it classifies each candidate as "preferred" or "unpreferred" by checking whether the name starts with _. The check incorrectly uses `Logic.name` (the raw element name) instead of `Logic.preferredSynthName` (the sanitized struct-hierarchy path). This causes struct leaf elements with auto-generated raw names like _s to lose their descriptive struct-path names.

### To Reproduce

```dart

class MyStruct extends LogicStructure {
  final Logic a;
  final Logic b;

  MyStruct({required this.a, required this.b, super.name = 'st'})
      : super([a, b]);

  @override
  MyStruct clone({String? name}) => MyStruct(
        a: a.clone(),
        b: b.clone(), // key: element.clone() → Naming.mergeable
        name: name ?? this.name,
      );
}

class Top extends Module {
  Top() {
    final inp = addTypedInput(
      'inp',
      MyStruct(
        a: Logic(name: 'a', width: 4),
        b: Logic(width: 4), // unnamed → auto '_s'
      ),
    );
    addOutput('out', width: 4) <= inp.b ^ inp.a;
  }
}

Future<void> main() async {
  final mod = Top();
  await mod.build();
  print(mod.generateSynth());
}
```

### Expected behavior

name should be 'inp__s'.

### Actual behavior

name is '_in0__s',

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

### LINKED ISSUE #649: SV Naming bug:  `tryMerge` blocks constant merging when `constNameDisallowed` is set
### Describe the bug

The `tryMerge` method in `SynthLogic` had a guard that refused to merge a constant signal with any signal whose `constNameDisallowed` flag was set. This flag is set on inputs to submodules that list them as `expressionlessInputs` (e.g., `BusSubset.original`). The guard was overly conservative — it prevented merging identical constants, producing dozens of redundant signal declarations for the same value.

### To Reproduce

```dart
class ExpressionlessSub extends Module with SystemVerilog {
  @override
  List<String> get expressionlessInputs => ['a', 'b'];

  ExpressionlessSub(Logic a, Logic b) {
    addInput('a', a, width: a.width);
    addInput('b', b, width: b.width);
    addOutput('out', width: a.width);
  }

  @override
  String? instantiationVerilog(
      String instanceType, String instanceName, Map<String, String> ports) =>
    '$instanceType $instanceName(.a(${ports['a']}), .b(${ports['b']}), '
    '.out(${ports['out']}));';
}

class BugModule extends Module {
  BugModule() {
    final c = Const(0, width: 8);
    final sub = ExpressionlessSub(c, c);
    addOutput('result', width: 8) <= sub.output('out');
  }
}

final mod = BugModule();
await mod.build();
print(mod.generateSynth());
```

### Expected behavior

one declaration:  `logic a; assign a = 8'h0;` with `.b(a)` reusing it

### Actual behavior

two declarations: `logic a; logic b; assign a = 8'h0;` `assign b = 8'h0;`

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

desmonddak:
## Description & Motivation

Discovered two anomalies in SystemVerilog synthesized names that need fixing to match a new converged solution.
It would be difficult to be compatible with these behaviors and this reduces noise in the differences in naming.

1) how unnamed structure leaf signals are named -- we were not picking up a field name as preferred.
2) we were replicating const assignment names dozens of times when we could use just one name.

## Related Issue(s)

Fix #648
Fix #649

## Testing

We replicated simple examples that demonstrate the two bugs, verified they expose the bug and the fixes work.

## Backwards-compatibility

No.  SV output will change slightly.

## Documentation

No.

### DIFF
--- a/lib/src/synthesizers/utilities/synth_logic.dart
+++ b/lib/src/synthesizers/utilities/synth_logic.dart
@@ -252,19 +252,6 @@ class SynthLogic {
     }
 
     // pick a preferred, available, mergeable name, if one exists
-    //
-    // Bug fix: the unpreferred check must use `preferredSynthName` (the
-    // sanitized struct-path name), not the raw `Logic.name`.
-    //
-    // A struct field created with `Logic()` (no name) auto-generates the
-    // raw name `_s`.  When the struct is cloned via `addTypedInput`,
-    // `chooseCloneNaming` promotes the field to `Naming.mergeable`, so it
-    // enters `_mergeableLogics`.  The raw name `_s` starts with `_` and
-    // IS unpreferred, but its `preferredSynthName` is the full struct
-    // path — e.g. `inp_myStruct__s` — which starts with `i` and is NOT
-    // unpreferred.  Checking `.name` instead of `.preferredSynthName`
-    // caused these fields to be wrongly deprioritized, losing their
-    // descriptive struct-path names in favor of anonymous `_in0` names.
     final unpreferredMergeableLogics = <Logic>[];
     final uniquifiableMergeableLogics = <Logic>[];
     for (final mergeableLogic in _mergeableLogics) {
@@ -335,18 +322,6 @@ class SynthLogic {
       a.adopt(b);
       return (removed: b, kept: a);
     }
-
-    // Bug fix: a previous guard here blocked merging when one side was a
-    // constant and the other had `constNameDisallowed`.  This was overly
-    // conservative.  `constNameDisallowed` only means "don't use the
-    // constant literal (e.g. `8'h0`) as the signal name" — it does NOT
-    // mean the signal cannot be merged.  `adopt()` already propagates
-    // the flag via `_constNameDisallowed |=`, so the merged result still
-    // gets a proper named declaration with an `assign` statement.
-    //
-    // Without this fix, when multiple `expressionlessInput` ports (e.g.
-    // from `BusSubset`) were driven by the same constant, each produced
-    // its own redundant declaration instead of sharing one signal.
 
     if (!a.mergeable && !b.mergeable) {
       return null;
--- a/lib/src/synthesizers/utilities/synth_logic.dart
+++ b/lib/src/synthesizers/utilities/synth_logic.dart
@@ -47,11 +47,13 @@ class SynthLogic {
   /// structure is a port of that [module].
   bool isStructPortElement([Module? module]) =>
       (this is! SynthLogicArrayElement) &&
-      logics.any((e) =>
-          e.isPort &&
-          e.parentStructure != null &&
-          e.parentStructure!.isPort &&
-          (module == null || e.parentStructure!.parentModule == module));
+      logics.any(
+        (e) =>
+            e.isPort &&
+            e.parentStructure != null &&
+            e.parentStructure!.isPort &&
+            (module == null || e.parentStructure!.parentModule == module),
+      );
 
   /// Indicates if this signal is a port, optionally for a specific [module].
   bool isPort([Module? module]) =>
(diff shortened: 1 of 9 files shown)


lib/src/synthesizers/utilities/synth_logic.dart
```diff
@@ -213,92 +229,124 @@ class SynthLogic {
...
     if (_renameableLogic != null) {
       return uniquifier.getUniqueName(
-          initialName: _renameableLogic!.preferredSynthName);
+        initialName: _renameableLogic!.preferredSynthName,
+      );
     }
 
     // pick a preferred, available, mergeable name, if one exists
+    //
+    // Bug fix: the unpreferred check must use `preferredSynthName` (the
```
mkorbel1: i think this bug fix description doesn't need to live in the codebase, it could be added as a PR note or something instead.  in the future people probably will not care that there used to be a bug here, and the usage of `preferredSynthName` seems obvious enough that it doesn't need any explanation why it shouldn't have been something else.  do you agree?

test/logic_structure_test.dart
```diff
@@ -320,4 +377,30 @@ void main() {
...
+      final mod = _StructLeafNamingModule();
+      await mod.build();
+      final sv = mod.generateSynth();
+
+      expect(sv, isNot(contains('_in0')),
+          reason: 'Struct leaf from unnamed Logic() should use its '
+              'struct-path name, not fall back to anonymous _in names.\n$sv');
+    });
+
+    test('const merge not blocked by constNameDisallowed', () async {
```
mkorbel1: this test doesnt seem like it should live in "logic_structure_test.dart"?

lib/src/synthesizers/utilities/synth_logic.dart
```diff
@@ -213,92 +229,124 @@ class SynthLogic {
...
       return (removed: b, kept: a);
     }
 
-    if ((a.isConstant && b.constNameDisallowed) ||
-        (b.isConstant && a.constNameDisallowed)) {
-      // we cannot merge a constant if the other disallows the constant name
-      // since we can lose a constant assignment
-      return null;
-    }
+    // Bug fix: a previous guard here blocked merging when one side was a
```
mkorbel1: similarly, this bug fix comment doesn't seem necessary in the code, just in the PR (though it is indeed a helpful comment for the PR).

mkorbel1 (changes requested):
Great, thank you!

test/logic_structure_test.dart
desmonddak: synth_logic.dart then?

lib/src/synthesizers/utilities/synth_logic.dart
desmonddak: Yes.  It is detritus.  I have to track this location in my future PRs so I found this useful to make sure I didn't erase the fix later, in a later PR.

desmonddak:
Bug fix #648: the unpreferred check must use `preferredSynthName` (the sanitized struct-path name), not the raw `Logic.name`.    
 A struct field created with `Logic()` (no name) auto-generates the raw name `_s`.  When the struct is cloned via `addTypedInput`, `chooseCloneNaming` promotes the field to `Naming.mergeable`, so it enters `_mergeableLogics`.  The raw name `_s` starts with `_` and  IS unpreferred, but its `preferredSynthName` is the full struct path — e.g. `inp_myStruct__s` — which starts with `i` and is NOT unpreferred.  Checking `.name` instead of `.preferredSynthName` caused these fields to be wrongly deprioritized, losing their descriptive struct-path names in favor of anonymous `_in0` names.
 
Bug fix #649: a previous guard here blocked merging when one side was a constant, and the other had `constNameDisallowed`.  This was overly conservative.  `constNameDisallowed` only means "don't use the constant literal (e.g. `8'h0`) as the signal name" — it does NOT mean the signal cannot be merged.  `adopt()` already propagatesthe flag via `_constNameDisallowed |=`, so the merged result still gets a proper named declaration with an `assign` statement.

 Without this fix, when multiple `expressionlessInput` ports (e.g., from `BusSubset`) were driven by the same constant, each produced its own redundant declaration instead of sharing one signal.

mkorbel1 (approved):
Looks great!

PR #652, opened 2026-04-17
Central synthesizer naming

desmonddak:
## Description & Motivation

Migrate synthesis naming to a central namer so that all synthesizers (and even WaveDumper) can agree on signal naming.
This is needed for a new synthesizer to be added (netlist).

## Related Issue(s)

None.

## Testing

I ran existing examples in ROHD and compared SV.

I added extensive test matrix `test/naming_cases_test.dart` for the combinations of naming options (reserved, mergeable, etc) for which I generated tests that pass before and after the central naming.

Two bugs were filed: bug #648 and bug #649 which are fixed in [PR # 650](https://github.com/intel/rohd/pull/650) for the traditional SV code as well as in this PR for central naming).

I ran a very, very, very large design example and compared SV before and after.

## Backwards-compatibility

Naming is slightly different due to ordering, so when we do have collisions, we will see _0, _1 on different signals. Plus the two bug fixes involving unnamed substructures and repeated const names.

## Documentation

No. None needed.


lib/src/synthesizers/synthesizer.dart
```diff
@@ -1,23 +1,39 @@
...
   /// by [getInstanceTypeOfModule].
+  ///
+  /// Optionally a [lookupExistingResult] callback may be supplied which
+  /// allows the synthesizer to query already-generated `SynthesisResult`s
+  /// for child modules (useful when building parent output that needs
+  /// information from children).
   SynthesisResult synthesize(
-      Module module, String Function(Module module) getInstanceTypeOfModule);
+      Module module, String Function(Module module) getInstanceTypeOfModule,
+      {SynthesisResult? Function(Module module)? lookupExistingResult,
```
mkorbel1: is this function and map actually used anywhere?  I don't see it?

lib/src/synthesizers/utilities/synth_sub_module_instantiation.dart
```diff
@@ -25,13 +24,16 @@ class SynthSubModuleInstantiation {
...
+  ///
+  /// Names are allocated from [parentModule]'s shared namespace via
+  /// [Module.allocateSignalName], ensuring no collision with signal names or
+  /// other submodule instances — even across multiple synthesizers.
+  void pickName(Module parentModule) {
     assert(_name == null, 'Should only pick a name once.');
 
-    _name = uniquifier.getUniqueName(
-      initialName: module.uniqueInstanceName,
+    _name = parentModule.allocateSignalName(
```
mkorbel1: why is this called `allocateSignalName` instead of `allocateInstanceName` or something?  A module is not a signal, confusing name?

lib/src/synthesizers/synthesizer.dart
```diff
@@ -1,23 +1,39 @@
...
 abstract class Synthesizer {
+  /// Called by [SynthBuilder] before synthesis begins, with the top-level
+  /// module(s) being synthesized.
+  ///
+  /// Override this method to perform any initialization that requires
+  /// knowledge of the top module, such as resolving port names to [Logic]
+  /// objects, or computing global signal sets.
+  ///
+  /// The default implementation does nothing.
+  void prepare(List<Module> tops) {}
```
mkorbel1: does this `prepare` function get used anywhere?

mkorbel1 (changes requested):
I haven't finished reading it all yet, but first pass a few questions

lib/src/synthesizers/synthesizer.dart
desmonddak: It is to be overridden as the netlist synthesizer (PR #654) needs to process top modules (no parents) whereas SV doesn't care.  I could use a better name and perhaps decorate it to be overrridden.

In `NetlistSynthesizer`:
```dart
  @override
  void prepare(List<Module> tops) {
    _topModules = Set<Module>.from(tops);
  }
  ```

File: lib/src/synthesizers/synthesizer.dart (hunk 1 above)
Comment (desmonddak) [reply lib/src/synthesizers/synthesizer.dart]: It was used for debug in netlist but is now removed.

File: lib/src/synthesizers/utilities/synth_sub_module_instantiation.dart (hunk 2 above)
Comment (desmonddak) [reply lib/src/synthesizers/utilities/synth_sub_module_instantiation.dart]: It should be -- in fact we have an issue that we currently don't allow an instance and a signal to have the same name which is now fixed in this version.

File: lib/src/synthesizers/synthesizer.dart (hunk 3 above)
Comment (desmonddak) [reply lib/src/synthesizers/synthesizer.dart]: Removed.  It is used by the upcoming netlist synthesizer, but we can use a simpler check there.

File: lib/src/synthesizers/utilities/synth_sub_module_instantiation.dart (hunk 2 above)
Comment (mkorbel1) [reply lib/src/synthesizers/utilities/synth_sub_module_instantiation.dart]: I thought signals and instances shared a namespace in verilog? Or am I wrong?

Comment (desmonddak) [reply lib/src/synthesizers/utilities/synth_sub_module_instantiation.dart]: Well my search indicated they do not share a namespace.  But looking at the LRM:
```text
The module name space is introduced by the module, interface, package, program, checker,
and primitive constructs. It unifies the definition of modules, interfaces, programs, checkers,
functions, tasks, named blocks, instance names, parameters, named events, net declarations, variable
declarations, and user-defined types within the enclosing construct.
```
So I guess all of these need to be uniquely named wrt each other.  Most CAD systems do allow netlists to have the same wire and instance names.
For example EDIF:
```text
In EDIF (Electronic Design Interchange Format), having the same name for an instance and a
 signal (net) is generally permissible because they exist in separate name spaces within the
syntax, allowing a net and an instance to share the same identifier.
```
But I guess because SystemVerilog is a 'programming' language and not a netlist, they choose to treat them as having the same namespace (e.g. a module is a namespace and *everything* needs to be named uniquely.  Which is dumb.  I'd never create a map with all objects in it.

I still think we should use `allocateSignalName` and `allocateInstanceName` because we may cache them (and again, I would always use typed maps for performance and safety).

desmonddak: I should say, I come from the Physical Design world where every format I have worked with (including my own for ChPPR) had separate namespaces. The only tricky case was ports versus wires where many systems disagree, especially when you deal with feedthrus and external connections.

desmonddak: I created a central namespace under Module to be consistent (SV naming is LCD).

lib/src/synthesizers/utilities/synth_logic.dart
```diff
@@ -11,12 +11,25 @@ import 'package:collection/collection.dart';
...
+  ///
+  /// When `true` (the default), identical constants are collapsed to one
+  /// declaration — desirable for simulation-oriented output such as
+  /// SystemVerilog, where a single `assign wire = VALUE;` feeds all
+  /// downstream consumers.
+  ///
+  /// When `false`, each constant input keeps its own declaration.  This is
+  /// useful for netlist/visualization outputs where seeing every individual
+  /// constant connection is more informative than an optimized fan-out net.
+  static bool mergeConstantInputs = true;
```
mkorbel1: This seems to be the first "global flag" in ROHD that changes behavior of everything (all modules, from any source).  It seems like it could introduce many corner case bugs, especially since it could change during execution (e.g. what if SV generation of a custom module modifies it?) and doubles the validation space of testing with and without this set.  Do we really need it?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,297 @@
...
+  factory Namer.forModule({
+    required Map<String, Logic> inputs,
+    required Map<String, Logic> outputs,
+    required Map<String, Logic> inOuts,
+  }) {
+    final portRenames = <Logic, String>{};
+    final portLogics = <Logic>{};
+    final portNames = <String>[];
+
+    void collectPort(String rawName, Logic logic) {
```
mkorbel1: port `Logic`s are guaranteed to have the same `name` as the name of the port, so I think the separate argument for `rawName` is unnecessary

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,297 @@
...
+      final sanitized = Sanitizer.sanitizeSV(rawName);
```
mkorbel1: port names must be sanitary always, so it should never be the case that `sanitized != logic.name`

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,297 @@
...
+        _portLogics = portLogics {
+    _signalNames.addAll(portRenames);
+  }
+
+  /// Creates a [Namer] for the given module ports.
+  ///
+  /// Sanitized port names are reserved in the signal namespace.  Ports
+  /// whose sanitized name differs from [Logic.name] are cached immediately.
+  factory Namer.forModule({
+    required Map<String, Logic> inputs,
```
mkorbel1: If this is `forModule`, why not pass a `Module` itself as one argument instead of the port maps?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,297 @@
...
+        throw UnavailableReservedNameException(baseName);
+      }
+
+      _uniquifier.getUniqueName(initialName: baseName, reserved: true);
+      return baseName;
+    }
+
+    var candidate = baseName;
+    var suffix = 0;
+    while (!_isAvailable(candidate)) {
```
mkorbel1: The `Uniquifier` already does this same operation in `getUniqueName`, why reimplement here?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,297 @@
...
+      candidate = '${baseName}_$suffix';
+      suffix++;
+    }
+
+    _uniquifier.getUniqueName(initialName: candidate);
+    return candidate;
+  }
+
+  /// Returns `true` if [name] has not yet been claimed in the namespace.
+  bool isNameAvailable(String name) => _isAvailable(name);
```
mkorbel1: This function and the one for instance don't appear to be used anywhere?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,297 @@
...
+
+  /// Allocates a collision-free name in the signal namespace.
+  ///
+  /// When [reserved] is `true`, the exact [baseName] (after sanitization)
+  /// is claimed without modification; an exception is thrown if it collides.
+  String allocateSignalName(String baseName, {bool reserved = false}) =>
```
mkorbel1: this seems redundant with `signalNameOf` and is unused?  or is this just exposed for testing purposes?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,297 @@
...
+/// are assigned lazily on the first [signalNameOf] call.  Instance names
+/// are allocated explicitly via [allocateInstanceName].
+@internal
+class Namer {
+  // ─── Shared namespace ───────────────────────────────────────────
+
+  final Uniquifier _uniquifier;
+
+  /// Sparse cache: only entries where the canonical name has been resolved.
+  /// Ports whose sanitized name == logic.name may be absent (fast-path
```
mkorbel1: I think this `_signalNames` never has any port names in it?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,297 @@
...
+
+    if (reserved != null) {
+      return _nameAndCacheAll(reserved, candidates);
+    }
+
+    if (renameable != null) {
+      return _nameAndCacheAll(renameable, candidates);
+    }
+
+    for (final logic in preferredMergeable) {
```
mkorbel1: nit: why does this one not follow the same `firstWhereOrNull` pattern as below?

test/name_test.dart
```diff
@@ -136,6 +136,11 @@ void main() {
         final nameTypes = [nameType1, nameType2];
 
         // skip ones that actually *should* cause a failure
+        //
+        // Note: SystemVerilog allows using the same identifier for a signal
```
mkorbel1: I think this note is incorrect after prior comment thread?

test/signal_registry_test.dart
```diff
@@ -0,0 +1,142 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// signal_registry_test.dart
+// Tests for Module canonical naming (SynthesisNameRegistry).
```
mkorbel1: what is `SynthesisNameRegistry`?

mkorbel1 (changes requested):
This is looking pretty good, thank you for refactoring it this way, I can see how it helps your other upcoming changes!

lib/src/synthesizers/utilities/synth_logic.dart
desmonddak: I need a solution that is different in SV versus netlist.  In SV, we want to optimize constants to have just one per value, but in netlist visualization this would result in very ugly schematics and we would rather see an individual constant wire at the point of that input.

I could instead push this into netlist only modification so on production of the netlist, individual constants are produced.

desmonddak: We didn't end up depending on this as the netlister actually splits them apart when it needs to.

lib/src/utilities/namer.dart
mkorbel1: Same question stands? dont see a change in latest commit

lib/src/utilities/namer.dart
```diff
@@ -40,108 +39,48 @@ class Namer {
...
-  bool isInstanceNameAvailable(String name) => _isAvailable(name);
-
-  /// Allocates a collision-free instance name.
-  ///
-  /// When [reserved] is `true`, the exact [baseName] (after sanitization)
-  /// is claimed without modification; an exception is thrown if it collides.
-  String allocateInstanceName(String baseName, {bool reserved = false}) =>
-      _allocateUniqueName(
-        Sanitizer.sanitizeSV(baseName),
+  String allocateRawName(String baseName, {bool reserved = false}) =>
```
mkorbel1: question: why is it a "raw" name and passed a "base" name?  Trying to understand the terminology and differences

lib/src/synthesizers/utilities/synth_module_definition.dart
```diff
@@ -760,8 +760,8 @@ class SynthModuleDefinition {
   ///
   /// Signal names are read from `Namer.signalNameOf` (for user-created
```
mkorbel1: These references to functions use ticks instead of `[]` which bypasses some lints, e.g. the `Namer.signalNameOf` no longer exists

test/name_test.dart
mkorbel1: unresolved, still looks like the issue is there?

test/signal_registry_test.dart
mkorbel1, line 1: file name is still called "signal registry"?

mkorbel1 (changes requested):
Looking close now to mergeable!

test/signal_registry_test.dart
desmonddak: //
// signal_registry_test.dart
// Tests for Module canonical naming (Namer).

I added more coverage tests.

lib/src/synthesizers/utilities/synth_sub_module_instantiation.dart
```diff
@@ -25,13 +25,15 @@ class SynthSubModuleInstantiation {
...
-  void pickName(Uniquifier uniquifier) {
+  ///
+  /// Names are allocated from [parentModule]'s [Namer]'s shared namespace
+  /// via [Namer.allocateName].
+  void pickName(Module parentModule) {
     assert(_name == null, 'Should only pick a name once.');
 
-    _name = uniquifier.getUniqueName(
-      initialName: module.uniqueInstanceName,
+    _name = parentModule.namer.allocateName(
```
mkorbel1: question: does this mean the name gets selected multiple times if you run the synthbuilder multiple times, but keeps the same `namer` being used?  doesn't this mean all the good names would already be taken and you'd end up with totally different naming if you synthesized again?

desmonddak: Indeed this was an issue and it signaled something critical: we want canonical names produced in `Module.build()` rather than just by the SV synthesizer to enable cross-probing between future synthesized formats.

A fix is in for this now.

lib/src/synthesizers/utilities/synth_module_definition.dart
```diff
@@ -14,11 +14,18 @@ import 'package:meta/meta.dart';
...
+import 'package:rohd/src/utilities/namer.dart';
 
 /// A version of [BusSubset] that can be used for slicing on [LogicStructure]
 /// ports.
 class _BusSubsetForStructSlice extends BusSubset {
+  /// The stable destination [Logic] this slice drives.
+  ///
+  /// Used as the [instanceNameKey] so that, although a fresh
+  /// [_BusSubsetForStructSlice] is created on every synthesis pass, its
+  /// canonical instance name is memoized against the persistent destination
```
mkorbel1, line 26: This comment looks like it used to apply to some field but is now just floating around?  Should this just be deleted?

lib/src/synthesizers/utilities/synth_module_definition.dart
```diff
@@ -767,49 +765,59 @@ class SynthModuleDefinition {
   }
 
   /// Picks names of signals and sub-modules.
+  ///
+  /// Signal names are read from [Namer.signalNameOf] for user-created
+  /// [Logic] objects) or kept as literal constants and are allocated from
+  /// [Namer.signalNameOf].  Submodule instance names are allocated
+  /// from [Namer.allocateName].  All names share a single
```
mkorbel1: it actually seems to be using `instanceNameOf` now?  In general, though, I don't think lower-level implementation details need to be enumerated in this comment anyways?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,257 @@
...
+  // ─── Name availability / allocation ─────────────────────────────
+
+  /// Returns `true` if [name] has not yet been claimed in the namespace.
+  bool isAvailable(String name) => _uniquifier.isAvailable(name);
+
+  /// Allocates a collision-free name in the shared namespace.
+  ///
+  /// When [reserved] is `true`, the exact [baseName] (after sanitization)
+  /// is claimed without modification; an exception is thrown if it collides.
+  String allocateName(String baseName, {bool reserved = false}) =>
```
mkorbel1: why is this wrapper around getUniqueName present?  why not just call the uniquifier's function as we did for signal names?

lib/src/synthesizers/utilities/synth_logic.dart
```diff
@@ -212,92 +212,25 @@ class SynthLogic {
...
-                  (element) =>
-                      uniquifier.isAvailable(element.preferredSynthName),
-                )
-                ?.preferredSynthName ??
-            unpreferredMergeableLogics.first.preferredSynthName,
+  ///
+  /// Delegates to signal namer which handles constant value naming, priority
+  /// selection, and uniquification via the module's shared namespace.
+  String _findName() =>
+      parentSynthModuleDefinition.module.namer.signalNameOfBest(
```
mkorbel1, line 229: Important question: In general for name selection for instances and signals, the order of selection seems very important.  For this PR, everything seems clean to me since the order of name selection will be consistent between SystemVerilog generation (as proven in your unit tests).  However, I'm curious about the long-term utility of this naming approach when *other* things start also needing to select names.  For example, if you attach debug infrastructure to query what names instances/signals have prior to running the SynthBuilder stack, wouldn't that affect name selection and thus give you different SystemVerilog depending on whether you debugged beforehand or not?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  // ─── Signal naming (Logic → String) ─────────────────────────────
+
+  /// Returns the canonical name for [logic].
+  ///
+  /// The first call for a given [logic] allocates a collision-free name
+  /// via the underlying [Uniquifier].  Subsequent calls return the cached
+  /// result in O(1).
+  String signalNameOf(Logic logic) {
```
mkorbel1: this function doesn't seem to be used outside of this class anymore except for in tests and comments? generally the API is replaced by `signalNameOfBest`?  so maybe this should be `@visibleForTesting` so we don't accidentally use this function elsewhere in the codebase and bypass the correct API?

lib/src/module.dart
```diff
@@ -202,6 +214,18 @@ abstract class Module {
...
+  /// across repeated synthesis passes (e.g. netlist then SystemVerilog).
+  ///
+  /// Defaults to the [Module] itself, which is correct for modules that are
+  /// part of the built hierarchy and therefore persist across passes.
+  /// Synthesis-time throwaway modules that are *recreated* on every pass (and
+  /// thus have a fresh [Module] identity each time) must override this to
+  /// return a stable identity — typically the [Logic] they drive — so their
+  /// instance name does not drift run-to-run.
+  @internal
+  Object get instanceNameKey => this;
```
mkorbel1, line 227: Is this thing still used?

mkorbel1 (changes requested):

lib/src/synthesizers/utilities/synth_logic.dart
desmonddak: The debug infrastructure has to run the netlist synthesizer to retrieve names, so by necessity, the `Namer` has run and assigned canonical names.

desmonddak: The latest changes create a weak-naming mark for `Logic`s that will be eliminated by SV heuristics.  We moved this marking into the base synthesizer utilities so it is common to all so that if other synthesizers run first, they will create the same naming bias that SV would have done.

The guarantee is that all synthesizers will share the same names off of the same build.  But this new feature creates the same bias on collapsible structures -- to name them after other `Logic` so that they get the weaker names upon collision.

lib/src/module.dart
mkorbel1: what's the resolution?  unresolved

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,244 @@
...
+/// ensuring no name collisions in the generated SystemVerilog.
+///
+/// Port names are reserved at construction time.  Internal signal names
+/// are assigned lazily on the first [signalNameOfBest] call.  Instance names
+/// are assigned lazily on the first [instanceNameOf] call.
+@internal
+class Namer {
+  // ─── Shared namespace ───────────────────────────────────────────
+
+  final Uniquifier _uniquifier;
```
mkorbel1, line 28: nit: doc comment?

lib/src/synthesizers/utilities/synth_module_definition.dart
```diff
@@ -746,71 +887,139 @@ class SynthModuleDefinition {
...
+    // had first chance at the shortest basenames.
+    for (final submodule in weakSubmodules) {
+      submodule.pickName(module);
+    }
+    for (final signal in _signalsInModuleOrder(weakSignals)) {
+      signal.pickName();
+    }
+  }
+
+  List<SynthLogic> _signalsInModuleOrder(Iterable<SynthLogic> signals) {
```
mkorbel1: what's the point of this function?  order is somewhat arbitrary, no?

lib/src/synthesizers/systemverilog/systemverilog_synth_module_definition.dart
```diff
@@ -30,36 +30,46 @@ class SystemVerilogSynthModuleDefinition extends SynthModuleDefinition {
...
             as SystemVerilogSynthSubModuleInstantiation)
-
           // map inouts to the appropriate `_SynthLogic`s
           ..setInOutMapping(_NetConnect.n0Name, dst)
           ..setInOutMapping(_NetConnect.n1Name, src);
 
     // notify the `SynthBuilder` that it needs declaration
     supportingModules.add(netConnect);
 
+    if (pickName) {
```
mkorbel1: what's the point of this pickName flag?  isn't it always true in the implementation?

lib/src/utilities/namer.dart
```diff
@@ -0,0 +1,244 @@
...
+      uniquifier.getUniqueName(initialName: logic.name, reserved: true);
+    }
+
+    return Namer._(uniquifier: uniquifier, portLogics: portLogics);
+  }
+
+  // ─── Name availability / allocation ─────────────────────────────
+
+  /// Returns `true` if [name] has not yet been claimed in the namespace.
+  bool isAvailable(String name) => _uniquifier.isAvailable(name);
```
mkorbel1, line 73: this should maybe just be `@visibleForTesting` since nothing else uses it?

lib/src/synthesizers/utilities/synth_module_definition.dart
desmonddak: Logic's declared first get priority.  Maybe this should say 'source order'.  Once you are in SynthLogic, you have unordered sets, so the naming is truly arbitrary unless we use this routine.

lib/src/synthesizers/systemverilog/systemverilog_synth_module_definition.dart
desmonddak: I was using it when we first worked on refactoring for two-pass (mark weak Logics).  We do not need this control, so I removed.

lib/src/module.dart
desmonddak: Overridden by `busSubsetForStructSlice`, a synthesized module.  All other modules just point to themselves.

lib/src/utilities/namer.dart
desmonddak: I put in a real comment.

lib/src/utilities/namer.dart
desmonddak: I set it to use that pragma.

desmonddak:
I use Resolve if I think I have fixed.  I accidentally thought I had resolved one I had not in the last round.  But otherwise, you may not know I have read and attempted a fix.

lib/src/synthesizers/utilities/synth_module_definition.dart
mkorbel1: Actually, surprisingly, the default `Set` implementation in Dart is a `LinkedHashSet` which iterates in the same order as things were added.  I think we can rely on that for consistent ordering of identical hardware?  It may be ordered based on something opaque (e.g. connectivity and construction order) but at least it is consistent?

desmonddak: The most recent checkin backs out this 'module order' feature and uses the stable order provided by the synthesis traverssal captured in this `Set` implementation to keep our code simpler.

lib/src/utilities/namer.dart
```diff
@@ -126,6 +126,12 @@ class Namer {
     return name;
   }
 
+  /// Returns the synthesis-level signal name for [logic].
+  ///
+  /// Equivalent to the internal [_signalNameOf] allocation but exposed for
+  /// use in wave-dumping and tests.
+  String signalNameOf(Logic logic) => _signalNameOf(logic);
```
mkorbel1, line 134: Should this be `@visibleForTesting` also?

lib/src/synthesizers/systemverilog/systemverilog_synth_module_definition.dart
```diff
@@ -30,34 +30,41 @@ class SystemVerilogSynthModuleDefinition extends SynthModuleDefinition {
...
 
+    netConnectSynthSubModInst.pickName(module);
+
     return netConnectSynthSubModInst;
   }
 
-  /// Replace all [assignments] between two [LogicNet]s with a [_NetConnect].
-  void _replaceNetConnections() {
+  /// Builds [_NetConnect] instances for [LogicNet] assignments.
+  void _buildNetConnectsForNaming() {
```
mkorbel1: why is this "forNaming"?

lib/src/synthesizers/systemverilog/systemverilog_synth_module_definition.dart
mkorbel1, line 1: update copyright year on this file

tool/gh_codespaces/install_dart.sh
mkorbel1, line 1: update copyright year on this file

lib/src/synthesizers/utilities/synth_module_definition.dart
```diff
@@ -513,9 +539,126 @@ class SynthModuleDefinition {
...
+    // process/collapse the marked objects.
+    prepareForNaming();
     _pickNames();
+    process();
+  }
+
+  /// Performs base-owned preparation before names are picked.
+  ///
+  /// Synthesizers must not override this method.
+  @nonVirtual
```
mkorbel1: why not just make it private instead of `@nonVirtual`?

mkorbel1 (changes requested):

lib/src/utilities/namer.dart
```diff
@@ -130,8 +140,14 @@ class Namer {
   ///
   /// Equivalent to the internal [_signalNameOf] allocation but exposed for
   /// use in wave-dumping and tests.
+  @visibleForTesting
   String signalNameOf(Logic logic) => _signalNameOf(logic);
 
+  /// Returns the [Logic] the namer chose as the source of signal [name], or
+  /// `null` if [name] has no single source [Logic] (e.g. a constant net) or
+  /// has not been named yet.
+  Logic? sourceLogicOf(String name) => _nameSources[name];
```
mkorbel1: why did you add this and `_nameSources`? nothing is using it?

mkorbel1 (changes requested):
some feedback not addressed, and also new weird thing added?

lib/src/utilities/namer.dart
desmonddak: It is used in a later PR.  I can move it there.  It is the reverse lookup:  we have a signal_15 and we need to get the `Logic` corresponding to it. This is necessary during FLC production to allow for cross-probing: we need to get the original signal name so we don't have to replicate symbol tables.

mkorbel1 (approved):
Looks great!

PR #653, opened 2026-04-17
ROHD Hierarchy: an API for compact, generic traversal of a remote design

desmonddak:
…pact, canonical addresses

## Description & Motivation

It would be nice to have a hierarchy api that allows us to remotely traverse a design hierarchy using compact addressing to support a devtools-based debugger.  For example, if we receive a dictionary of the design, we can communicate requests for more data from the design by using a compressed addressing scheme like (5th instance, 4th instance below that, 15th signal) to receive a set of data on that signal.  If we need 1000s of simulation values, for example, we do not need to transmit all names, providing an order of magnitude savings in transmission bandwidth.

This capability allows us to traverse an occurrence view over multiple data requests incrementally.  For now, we primarily have two views: a 'slim' view, which displays the full hierarchy with names, and a full view, which shows all connectivity. This can be extended to allow for a more incremental dictionary load to allow a remote agent to use limited memory to transit a very large design.

## Related Issue(s)

None.

## Testing

Tests are provided with a small design fixture using a new netlist format in JSON, equivalent to Yosys JSON.  A future netlist synthesizer will provide design data nd this `rohd_hierarchy adapter` will allow us to communicate between remote agents.

## Backwards-compatibility

No

## Documentation

A `README.md` is included to describe the motivation, architecture, and key features of the api.


packages/rohd_hierarchy/analysis_options.yaml
mkorbel1, line 1: should we include (either copy/paste or via some sort of `include` if that is supported sufficiently) the rest of the analyzer rules from base ROHD?

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
```diff
@@ -0,0 +1,576 @@
...
+    address = startAddr;
+    for (final (i, s) in signals.indexed) {
+      s.address = startAddr.signal(i);
+    }
+    for (final (i, c) in children.indexed) {
+      c.buildAddresses(startAddr.child(i));
+    }
+  }
+}
+
```
mkorbel1: Split classes into individual files?

mkorbel1 (commented):
Will be reviewing this next, couple quick comments until I have a moment to continue.  This looks really great so far!

mkorbel1:
Question: how do you imagine this package being used/depended on?  Would we release it as a separate package on pub.dev?

packages/rohd_hierarchy/lib/rohd_hierarchy.dart
```diff
@@ -0,0 +1,52 @@
...
+/// - [BaseHierarchyAdapter] - Base class with shared adapter implementation
+/// - [NetlistHierarchyAdapter] - Adapter for netlist format (Yosys JSON / ROHD)
+///
+/// This package has no dependencies and can be used standalone by any
+/// application that needs to navigate hardware hierarchies.
+///
+/// ## Quick Start
+/// ```dart
+/// // 1. Create hierarchy
+/// final root = HierarchyNode(id: 'top', name: 'top',
```
mkorbel1: nit: multi-line code comment not formatted as nicely as it could have been

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
```diff
@@ -0,0 +1,576 @@
...
+// 2026 January
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:collection/collection.dart';
+import 'package:meta/meta.dart';
+
+/// The kind of node in the hardware hierarchy.
+enum HierarchyKind {
+  /// A module definition in the hierarchy.
+  module,
```
mkorbel1: should this be called "definition" instead to contrast with "instance"?  and actually, more generally, what is a definition in a hierarchy?

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
```diff
@@ -0,0 +1,576 @@
...
+  ///
+  /// Tool-specific primitive types (e.g. ROHD's `FlipFlop`) should be
+  /// handled by the producer: the synthesizer should map them to
+  /// `$`-prefixed cell types in the JSON output, or the adapter should
+  /// set [isPrimitive] on the node at construction time.
+  ///
+  /// Use this before a [HierarchyNode] exists (e.g. when deciding whether
+  /// to recurse into a Yosys cell definition).  For an existing node, use
+  /// the instance getter [isPrimitiveCell] instead.
+  static bool isPrimitiveType(String cellType) => cellType.startsWith(r'$');
```
mkorbel1: question: is this yosys conventions leaking into a generic abstraction or is it fine to keep them coupled like this?

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
```diff
@@ -0,0 +1,576 @@
...
+  /// Empty for instances.
+  final List<Signal> signals;
+
+  /// Child modules/instances. Populated from sub-modules in the hierarchy.
+  final List<HierarchyNode> children;
+
+  /// Hierarchical address for this node.
+  /// Assigned by [buildAddresses] to enable efficient navigation.
+  /// Format: [child0, child1, ..., childN] for nested modules.
+  HierarchyAddress? address;
```
mkorbel1: something feels off about this not being immutable (`final`) and requires a function later to update it.  is there a way to make this safer/more automatic?

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
```diff
@@ -0,0 +1,576 @@
...
+    super.isComputed,
+  }) : super(direction: direction);
+
+  /// Creates a Port with minimal parameters.
+  factory Port.simple({
+    required String name,
+    required String direction,
+    int width = 1,
+    String? id,
+    String type = 'wire',
```
mkorbel1: "wire" is an SV type -- do we want that to be the default here?  in ROHD everything is `Logic` and ports are by default `logic` as well in generated SV, wheras `LogicNet` inOut ports would map to `wire`

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
```diff
@@ -0,0 +1,576 @@
...
+  /// Always set in production code; may be null in test fixtures.
+  final String? fullPath;
+
+  /// ID of the scope (module) containing this signal.
+  final String? scopeId;
+
+  /// Direction of the signal if it's a port.
+  /// Null for internal signals (wires, registers).
+  /// "input", "output", or "inout" for ports.
+  final String? direction;
```
mkorbel1: why do signals have a direction if they are only used for ports, which have their own class that extends signal?

packages/rohd_hierarchy/lib/src/hierarchy_search_controller.dart
```diff
@@ -0,0 +1,219 @@
...
+  // ─────────────── Scroll helper ───────────────
+
+  /// Compute the scroll offset needed to reveal the selected item in a
+  /// fixed-height list.
+  ///
+  /// Returns `null` if the item is already visible.  The caller should
+  /// call `scrollController.jumpTo(offset)` with the returned value.
+  ///
+  /// This is a pure calculation with no Flutter dependency.
+  static double? scrollOffsetToReveal({
```
mkorbel1, line 201: does this make sense to keep in this package?  i dont know where it is used, though

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
```diff
@@ -0,0 +1,865 @@
...
+// hierarchy_service.dart
+// Abstract interface for source-agnostic hardware hierarchy navigation.
+//
+// 2026 January
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:rohd_hierarchy/src/hierarchy_models.dart';
+
+/// Default path separator used when constructing paths from the tree.
+const String _hierarchySeparator = '/';
```
mkorbel1: multiple files assume `/`, should const be centralized and reused?

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
```diff
@@ -0,0 +1,865 @@
...
+
+/// A source-agnostic interface for navigating hardware hierarchy.
+///
+/// All search and navigation is driven by walking the [HierarchyNode] tree.
+/// Nodes hold their [HierarchyNode.name], [HierarchyNode.children], and
+/// [HierarchyNode.signals].  Full paths are constructed on the fly by
+/// joining names with [_hierarchySeparator] — no pre-baked path strings
```
mkorbel1: reference to private var in public doc comment

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
```diff
@@ -0,0 +1,865 @@
...
+      final sigIdx = indices.last;
+      return (sigIdx >= 0 && sigIdx < walked.node.signals.length)
+          ? [...walked.parts, walked.node.signals[sigIdx].name]
+              .join(_hierarchySeparator)
+          : null;
+    }
+    return walked.parts.join(_hierarchySeparator);
+  }
+
+  /// Resolve a waveform-style ID (dot-separated, e.g. `"dut.adder.clk"`)
```
mkorbel1: first time in the code i've seen a reference to "waveform-style" -- is this terminology a remnant of a refactor?

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
```diff
@@ -0,0 +1,865 @@
...
+    final parts = _splitPath(query);
+    final results = <String>[];
+    _searchSignalsRecursive(
+        root, [root.name], parts, 0, results, effectiveLimit);
+    return results;
+  }
+
+  /// Whether [query] contains glob or regex metacharacters that should
+  /// trigger the regex search engine instead of the plain substring search.
+  static bool hasRegexChars(String query) =>
```
mkorbel1, line 157: is this a safe way to decide what kind of search to use generally?

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
```diff
@@ -0,0 +1,865 @@
...
+      query.contains('?') ||
+      query.contains('[') ||
+      query.contains('(') ||
+      query.contains('|') ||
+      query.contains('+');
+
+  /// Check if a [node] or any of its descendants match [searchTerm].
+  ///
+  /// The search term is split on `/` or `.` into hierarchical segments.
+  /// Each segment is matched case-insensitively via substring containment
```
mkorbel1: question: why case insensitive?

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
```diff
@@ -0,0 +1,865 @@
...
+    if (query.trim().isEmpty) {
+      return const [];
+    }
+    final parts = _splitPath(query);
+    final results = <String>[];
+    _searchSignalsRecursive(
+        root, [root.name], parts, 0, results, effectiveLimit);
+    return results;
+  }
+
```
mkorbel1, line 154: general question: have you considered fuzzy searches?

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
mkorbel1, line 1: question: will this package eventually support tracing connectivity? e.g. connection extractor in rohd bridge?

packages/rohd_hierarchy/test/coverage_gaps_test.dart
mkorbel1, line 1: why not included in appropriate locations?

packages/rohd_hierarchy/test/hierarchy_search_controller_test.dart
```diff
@@ -0,0 +1,587 @@
...
+        viewportHeight: 250,
+        currentOffset: 0,
+      );
+      // item at 200..250, viewport 0..250 → exactly visible
+      expect(offset, isNull);
+    });
+  });
+
+  // ------------------------------------------------------------------
+  // VCD-style dot-separated paths
```
mkorbel1, line 267: "vcd style"? another new term?

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
desmonddak: We may want to change this but for new having primitive start with '$' is a good base check.

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
desmonddak: We removed this enum.  We introduce the concepts of `occurrence` for a level of hierarchy and `occurrence signal`, versus a module `definition` or a `Logic`.

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
desmonddak: Created a getter.

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
desmonddak: I created a more formal regex api to allow us to extend.

packages/rohd_hierarchy/test/coverage_gaps_test.dart
desmonddak: Merged into other tests.

packages/rohd_hierarchy/test/hierarchy_search_controller_test.dart
desmonddak: For now this makes sense. We prefer '/' separated hierarchy, but VCD uses '.'

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
desmonddak: Yes, this package has hooks in it to support loading and eventually tracing connectivity.

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
desmonddak: This is an open bug -- VCD is capitalizing the top module and we need to push the fix there.  
We have some generalization now on hierarchy separator to make this package immune to that.

packages/rohd_hierarchy/lib/src/hierarchy_models.dart
desmonddak: See comments above.  It would be good to isolate 'master' kinds of definitions (definition/port/instance/net/Logic/etc.) from occurrence (unfolded) references.
our use of `module` blurs that a bit, serving both as definition and occurrence.  We need a third concept.

desmonddak: I removed SVisms, and reduce the usage of terms to try to achieve a canonical subset.

desmonddak:
It is unclear if this really should be a separate package or integrated into ROHD.  The waveform viewer depends on it and that is somewhat independent of ROHD.  So I created it as a package because other capabilities will come that are in ROHD proper that use it as well.

packages/rohd_hierarchy/analysis_options.yaml
desmonddak: We now include the `../../analysis_options.yaml`.

packages/rohd_hierarchy/README.md
```diff
@@ -0,0 +1,243 @@
...
+## Package overview
+
+`rohd_hierarchy` is a source-agnostic Dart package that implements this
+dictionary model. It provides data models, search utilities, and adapter
+interfaces that work independently of any particular HDL toolchain or
+transport layer.
+
+### Data models
+
+- **`HierarchyOccurrence`** — An occurrence of a module definition in the
```
mkorbel1: occurrence of a module instance, not definition, right?

mkorbel1 (commented):
(note also that there are remaining unresolved comments from last review)

packages/rohd_hierarchy/test/coverage_gaps_test.dart
mkorbel1: i still see the file here?

mkorbel1 (commented):
looking pretty good!

packages/rohd_hierarchy/test/coverage_gaps_test.dart
desmonddak: I had migrated tests to other files but did not delete so it is entirely redundant.

packages/rohd_hierarchy/lib/src/hierarchy_search_controller.dart
desmonddak: This is used by the widgets to find occurrences by regex.

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
desmonddak: Migrated to a constants file.

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
desmonddak: added to constants file, made public, renamed to hierarchyPathSeparator.

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
desmonddak: This is safe:  if it has no special characters, it will perform a direct string search, so this is an optimization.

packages/rohd_hierarchy/lib/src/hierarchy_service.dart
desmonddak: Bug has been fixed so we do case-sensitive searching now.

desmonddak: All comments to the contrary fixed.

packages/rohd_hierarchy/test/hierarchy_search_controller_test.dart
desmonddak: I've tried to clarify elsewhere that hierarchies that come from loading waveform files like VCD or FST will have '.' as a separator and so we need to transform to '/' for everything to work -- or at least that is the policy we are using rather than trying to use multiple separators.

This is a common problem in CAD tools:  a lot of backend tools use '.' to represent hierarchy that has been flattened versus '/' to represent hierarchy that is still retained in layout.  So I see this a lot: what hierarchy separator to use and in what situation.

packages/rohd_hierarchy/README.md
desmonddak: I've updated the README to be more clear on definitions.

packages/rohd_hierarchy/test/coverage_gaps_test.dart
desmonddak: Deleted in latest PR

desmonddak:
> (note also that there are remaining unresolved comments from last review)

I went through comments I 'resolved' and answered them more carefully with a fix and a comment in many cases.

mkorbel1 (approved):
Looks good to me now I think!

PR #654, opened 2026-04-17
Netlist synthesizer

desmonddak:
## Description & Motivation

This is a netlist synthesizer that uses central naming to match the names of our SystemVerilog synthesizer but outputs in the Yosys JSON format for external use.

Also added are examples, pushed into lib/src so that we can access them with tests.

One exhaustive example is filter_bank which has all elements of ROHD included.

## Related Issue(s)

This PR includes all changes from PR #652, so we should merge that first to be clean.

## Testing

Netlist tests are included using the examples.

## Backwards-compatibility

No.

## Documentation

Yes, it is a new capability.  Basic documentation is in `architecture.md` and in one of the tutorials.


desmonddak:
Note that I moved examples into lib/src to enable them to be used as part of tests.  The example directory has the outer main and tests for these examples.

PR #657, opened 2026-04-18
devtools improvements

desmonddak:
## Description & Motivation

Our devtools implementation has significant gaps, even as a shell, and this PR is to prepare it to support more functionality by making at least the shell work properly.

Currently, it would not run using `Open DevTools in Browser` because of a service worker problem (waiting for one that never gets started).  It also would only connect to the first isolate running at top level (say in an example) and not be able to connect to an isolate in a test function inside a main().

So now the shell can at least connect to the first available debugger session using `Open DevTools in Browser`.

I added some basic widget utilities that will be used in upcoming enhancements to devtools:  a Help widget and a Snapshot widget (allowing for screenshots of widget panes).

So, the shell is now workable; it can connect to a ROHD debug session, load its design into the module tree widget, and show the details pane with other functionality stubbed out but plumbed.

## Testing

Ran the build_web.sh
Ran a debug session in the test area on a simulation.
Ran a devtools session and connected to the debugger, displayed the details pane, took snapshots.

## Backwards-compatibility

Some ambiguity on the service worker connection (we don't seem to need it after extensive testing in a more complex devtools implementation)

## Documentation

No update required.


desmonddak:
We have integrated a better base layer for devtools.

PR #660, opened 2026-05-03
trial run for Node v24

desmonddak:
## Description & Motivation

Node 20 will be deprecated and we are using advanced node features in future PRs.

## Related Issue(s)

None.

## Testing

We have been using Node 24 in local testing.  This is a workflow CI test.

## Backwards-compatibility

No.

## Documentation

None needed.

### DIFF
--- a/.github/workflows/general.yml
+++ b/.github/workflows/general.yml
@@ -38,7 +38,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v4
         with:
-          node-version: 20
+          node-version: 24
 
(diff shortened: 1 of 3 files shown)


mkorbel1 (approved):
Looks good to me, any reason not to just merge this in?

desmonddak:
I don't see a reason not to merge in.  I have an equivalent PR for rohd-hcl and it passed CI.
Using node24 locally.

mkorbel1 (approved):
Looks good!

PR #661, opened 2026-05-04
Update coverage.yml

sfblackl-intel:
Resolving OpenSSF Token Permission issue.

## Backwards-compatibility

## Documentation


mkorbel1 (approved):
Thank you for the fix!

PR #662, opened 2026-05-08
SystemC_04

desmonddak:
## Description & Motivation

We should have a SystemC translation of our ROHD modules to simulate in a SystemC environment.

## Related Issue(s)

None.

## Testing

Leverage iverilog tests to drive SystemC testing, except for somethings like lack of 'X' propagation.

## Backwards-compatibility

No.

## Documentation

Very basic documentation added to parallel SystemVerilog.


lib/src/synthesizers/systemc/systemc_synth_sub_module_instantiation.dart
```diff
@@ -0,0 +1,113 @@
...
+    );
+
+    final inlineRepresentation =
+        _inlineSystemCExpression(portNameToValueMapping);
+
+    return '($inlineRepresentation)';
+  }
+
+  /// Generates the inline SystemC expression for the gate module.
+  String _inlineSystemCExpression(Map<String, String> inputs) {
```
mkorbel1, line 57: why do this here instead of how SV does it as a part of the module definition?  surely we can reuse the operator string

test/systemc_vector_test.dart
```diff
@@ -0,0 +1,1279 @@
...
+// Author: Desmond A. Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'dart:math';
+import 'package:rohd/rohd.dart';
+import 'package:rohd/src/utilities/simcompare.dart';
+import 'package:test/test.dart';
+
+// ===== Modules from flop_test.dart =====
+
+class FlopTestModule extends Module {
```
mkorbel1: why do modules need to be copied in here?  why can't it leverage them either via relative import or just add to the existing tests to run the same existing vectors on systemc versions as well?

lib/src/synthesizers/systemc/systemc_synthesis_result.dart
```diff
@@ -0,0 +1,1750 @@
...
+    // Promote internal clock signals (from SimpleClockGenerator) to ports
+    for (final clkName in _promotedClockSignals) {
+      final n = _scName(clkName);
+      lines.add('  ${systemCInType(1)} $n{"$n"};');
+    }
+    for (final sig in _synthModuleDefinition.outputs) {
+      final n = _scName(sig.name);
+      lines.add('  ${systemCOutType(sig.width)} $n{"$n"};');
+    }
+    return lines.join('\n');
```
mkorbel1, line 245: are inouts not supported?

lib/src/synthesizers/systemc/systemc_synthesis_result.dart
```diff
@@ -0,0 +1,1750 @@
...
+      if (m is InlineSystemVerilog) {
+        final resultSynthLogic = ssmi.inlineResultLogic;
+        if (resultSynthLogic == null) {
+          continue;
+        }
+        final expr = _gateExpression(m, inputExprs);
+        final dst = _scName(resultSynthLogic.name);
+        destinations.add(dst);
+        bodyLines.add('    $dst = $expr;');
+      } else if (m is Add) {
```
mkorbel1, line 538: here also, can't this special case be implemented in Add itself?

lib/src/utilities/simcompare.dart
```diff
@@ -436,4 +427,757 @@ abstract class SimCompare {
...
+  /// Path to the precompiled header, built lazily on first compilation.
+  static String? _pchPath;
+
+  /// Builds the precompiled header for systemc.h if not already done.
+  /// Returns the directory containing systemc.h.gch, or null on failure.
+  ///
+  /// In CI, the PCH is pre-built by `tool/gh_actions/setup_systemc_pch.sh`
+  /// before tests run, so this just finds it on disk. Locally it builds
+  /// on first use (safe because local runs are typically sequential).
+  static String? _ensurePch(String scHome, String cxxStd) {
```
mkorbel1: should we split out system-c specific (and sv specific?) things from simcompare into separate classes/utilities?

lib/src/utilities/simcompare.dart
```diff
@@ -436,4 +427,757 @@ abstract class SimCompare {
...
+      return false;
+    }
+    if (buildOnly) {
+      return true;
+    }
+    return runSystemCVectors(exe, vectors);
+  }
+
+  // ══════════════════════════════════════════════════════════════════════
+  // Trace-based SystemC co-simulation
```
mkorbel1: this is cool, but why only systemc? can we make it more generic up through sv also?

lib/src/utilities/systemc_cosim_ffi.dart
```diff
@@ -0,0 +1,961 @@
...
+      ..writeln('    static char _buf[512];');
+
+    _generateOutputDispatch(sb, narrow: false);
+
+    sb
+      ..writeln("    _buf[0] = '0'; _buf[1] = 0;")
+      ..writeln('    return _buf;')
+      ..writeln('}')
+      ..writeln()
+      // ──── sc_cosim_advance ────
```
mkorbel1: multi-line strings instead?

lib/src/utilities/systemc_cosim_ffi.dart
mkorbel1, line 1: should this live in rohd-cosim instead and also support arbitrary systemc modules instead of only ROHD-generated ones?

lib/src/module.dart
```diff
@@ -1155,6 +1156,27 @@ abstract class Module {
             .getSynthFileContents()
             .join('\n\n////////////////////\n\n');
   }
+
+  /// Returns a synthesized SystemC version of this [Module].
+  ///
+  /// Generates SystemC code that is equivalent to the hardware described by
+  /// this module, using the same naming strategy as [generateSynth].
+  String generateSystemC() {
```
mkorbel1, line 1170: similar comment as before -- shall we avoid adding functions like this in favor of heading towards deprecation of generateSynth or making it more generic to accept different `Synthesizer`s?

lib/src/synthesizers/systemc/systemc_synth_sub_module_instantiation.dart
desmonddak: I'm not sure:  this is an architectural question -- as you add languages, do you expand `Module` for inlining and could different `Module`s be 'leaf' for different languages.  In the case of netlist, the leaf modules are just empty boxes.

SystemC does some type conversions as it assigns
``` systemc 
_aExtended_add_bExtended_carry = static_cast<bool>(sc_uint<6>(sc_uint<6>(aExtended.read()) + sc_uint<6>(bExtended.read()))[5]);

   bExtended = (sc_uint<1>(bool(0)), b.read());
```

But SystemVerilog can be different:
```verilog
assign {_aExtended_add_bExtended_carry, sum} = ({
```

On the top-side we agreed that we don't want `Module.generateSynth` even as a simplification to `.generateSystemVerilog` and `.generateSystemC`, but on the leaf side, we need some kind of `LeafService` which provides these things rather than fattening `Module` for each language.

Really, the 'inline' concept is fairly tied to SV output, and while we could transmogrify the string, it is probably cleaner to have an API that says 'am I leaf in this context, if so, let me pass you to the `LeafService` with this context and module.

For example, the MLIR compiled version may need something else at leaf level.

test/systemc_vector_test.dart
desmonddak: I should fix.

We DO run the same vectors on SystemC as SystemVerilog (see simcompare.dart).

lib/src/synthesizers/systemc/systemc_synthesis_result.dart
desmonddak: They are -- this is my mistake.  `sc_inout`.

lib/src/utilities/simcompare.dart
desmonddak: Yes.  I'll look into a cleaner partitioning.

lib/src/synthesizers/systemc/systemc_synth_sub_module_instantiation.dart
mkorbel1: i agree I think, should we consider what the API might look like for SV first and then map systemc onto it? or do it together?  there's a lot of context within the modules that is relevant for generating outputs of them, some of it private, but maybe it need not be private and can be defined and registered separately?  then there's backwards compatibility to consider.

lib/src/utilities/systemc_cosim_ffi.dart
desmonddak: Here is the dilemma:  for SystemVerilog we have the iVerilog simulator for doing regressions.  For SystemC we can use the SystemC compiler (really just g++)  But in order to drive it, I  used this ffi technique.

But we cannot depend on rohd-cosim inside rohd to validate SystemC gate representations which is what we want to do: for every iVerilog simulation of a ROHD primitive, we would want to add a quick sim for SystemC to do a miter test with the same vectors (simcompare).  

So should we add this to rohd-cosim -- sure, but rohd-cosim already points to rohd, so perhaps it can import them, or we can build a more general form.

If we move it entirely there, then we cannot validate the SystemC synthesizer;  as we add more capabilities to ROHD (perhaps another leaf gate), it would be good to validate in parallel with Verilog.

lib/src/module.dart
desmonddak: I'll remove.

lib/src/utilities/systemc_cosim_ffi.dart
mkorbel1: Ah ok, gotcha.  Options are to use the FFI like you have i guess, or to do something like we do in the iverilog case where we generate the vectored testbench natively in the target language instead of expecting cross-language communication.  either way seems ok to me, but now i understand the motivation for this in here, thanks!

test/systemc_vector_test.dart
desmonddak: I merged testing for these into the main unit tests for components.

lib/src/synthesizers/systemc/systemc_synthesis_result.dart
desmonddak: I added some support.

lib/src/utilities/systemc_cosim_ffi.dart
desmonddak: Fixed

lib/src/synthesizers/systemc/systemc_synth_sub_module_instantiation.dart
desmonddak: This is a large architectural change.  I have a PR for it, but I would like to isolate from this PR so that we can have a basic SystemC capability and then we can study how to make this change.  Basically I looked at how SV has a lot of leaf-level output code embeded in the modules themselves as well as there being a mixin.  Then the question is how do we have the same type of functionality but be output language agnostic and make sure we use this across all synthesizers.  So I'm hoping we can resolve this for now and defer to the other PR (actually two of them: one modifies SystemC for the new paradigm, keeping the existing SV output and the second transitions SV to the new paradigm).

lib/src/module.dart
desmonddak: I need to leave this in temporarily to allow for tests and removing depends on integrating module_services_api first.
If we make that decision, I can remove.

lib/src/utilities/simcompare.dart
desmonddak: We could do it for Verilator but not iVerilog.  But this is introducing yet more complexity that maybe we will migrate to cosim.   For iVerilog we would end up with an IPC pipe, similar to rohd-cosim.

We could consider migrating the FFI calls to cosim as systemc could just run with vectors.

lib/src/synthesizers/systemc/systemc_synthesis_result.dart
desmonddak: I will introduce a new PR that migrates this functionality to a cleaner, more shared architecture.

lib/src/utilities/simcompare.dart
desmonddak: Done.

PR #663, opened 2026-06-03
Element, subset, and swizzle collapsing in generated SV

mkorbel1:
## Description & Motivation

This PR dramatically improves SV generation by reducing the amount of unnecessary assignments, `net_connect`s, swizzles, subsets, and intermediate signals for `Logic`s and `LogicArray`s (both net and non-net) in generated SystemVerilog.  This makes the SV much more readable, debuggable, and tool friendly.

## Related Issue(s)

N/A

## Testing

Added extensive new testing for collapsing capabilities, existing suite helps protect.

## Backwards-compatibility

No

## Documentation

No


lib/src/synthesizers/systemverilog/systemverilog_synth_module_definition.dart
```diff
@@ -80,6 +92,840 @@ class SystemVerilogSynthModuleDefinition extends SynthModuleDefinition {
...
+
+          // Preferred path: the element is connected by exactly one assignment
+          // and used nowhere else.
+          final connectingAssignments = elementAssignments[element];
+          if (connectingAssignments != null &&
+              connectingAssignments.length == 1 &&
+              (elementUseCount[element] ?? 0) == 1) {
+            final assignment = connectingAssignments.single;
+            final source =
+                assignment.src == element ? assignment.dst : assignment.src;
```
desmonddak: Should this be a resolved comparison?  What if src or dst was unresolved and then looked up in the array? IS there a way for this to happen?

lib/src/synthesizers/systemverilog/systemverilog_synth_module_definition.dart
```diff
@@ -16,8 +16,20 @@ class SystemVerilogSynthModuleDefinition extends SynthModuleDefinition {
...
+  ///
+  /// Populated by aggregate-collapse passes, then distributed alongside
+  /// [_collapseMarkedChainableModules] so inline rendering (including
+  /// recursive, multi-level chains) can resolve them.
+  final Map<SynthLogic, SystemVerilogSynthSubModuleInstantiation>
+      _inlineableSubmoduleMap = {};
+
   @override
   void process() {
+    _collapseAggregateConnections();
```
desmonddak, line 30: Since these were designed before the central_naming idea, they will break from it and collapse connections potentially with good names, leaving weak names for other nets.

There may be a need for three levels:   mark the stuff that could go away,  an area for common optimizations, an area for language-specific.  The marking could be a superset of all known optimizations that could happen (give those Logics the lowest priority in naming).

lib/src/synthesizers/systemverilog/systemverilog_synth_module_definition.dart
mkorbel1: Good point. I don’t think this can currently fail because SynthAssignment.src/dst resolve replacements in their getters, and both elementAssignments and elementLogics are keyed with .resolved. So an unresolved endpoint should be normalized when it is read. But using .resolved here matches the rest of this pass and makes the invariant explicit, so I’ll update it.

lib/src/synthesizers/systemverilog/systemverilog_synth_module_definition.dart
mkorbel1: That makes sense. These passes currently run in the SV `process()` phase after names are picked, so they can clear an object that already claimed a good basename. The output stays correct, but surviving objects may get uglier suffixes.

I think the clean long-term fix is probably a pre-naming weak-claim hook for language-specific optimizations: base would still own the naming order, while SV could conservatively mark aggregate/net-bus candidates that may be removed later. The actual collapse realization would remain in the SV stack.

I’d prefer not to move these full passes into base in this PR, since the legality and realization are SV-specific and the exact candidate search is fairly stateful/fixed-point. For now I think the naming tradeoff is acceptable, and we can follow up with a generic “mark possible language-specific removals before naming” hook if the suffix churn becomes noticeable.

test/array_collapsing_test.dart
```diff
@@ -182,4 +951,854 @@ void main() {
...
+                reason: 'no per-bit net_connect should remain: $reason');
+          }
+        }
+
+        // --- functional correctness (always) ---
+        await SimCompare.checkFunctionalVector(mod, vectors);
+        SimCompare.checkIverilogVector(mod, vectors);
+      });
+    }
+  });
```
desmonddak, line 1896: I didn't see any naming checks to see if you end up retaining only basename nets or accidentally toss that net and keep one with _0.  I think this code is not enforcing the namig order, but it would be good to have a test for it and know how you are going to enforce better names (with a current test that you waive).

test/array_collapsing_test.dart
```diff
@@ -182,4 +951,854 @@ void main() {
...
+            expect(topBody, isNot(contains('net_connect')),
+                reason: 'no per-bit net_connect should remain: $reason');
+          }
+        }
+
+        // --- functional correctness (always) ---
+        await SimCompare.checkFunctionalVector(mod, vectors);
+        SimCompare.checkIverilogVector(mod, vectors);
+      });
+    }
```
desmonddak, line 1895: I test is needed where a SynthLogic has been merged before a specific aggregate collapse happens.
element <= intermediate;
intermediate <= realSource;
You want to be sure you don't see the intermediate kept.

test/array_collapsing_test.dart
```diff
@@ -182,4 +951,854 @@ void main() {
...
+          if (config.isNet) {
+            expect(topBody, isNot(contains('net_connect')),
+                reason: 'no per-bit net_connect should remain: $reason');
+          }
+        }
+
+        // --- functional correctness (always) ---
+        await SimCompare.checkFunctionalVector(mod, vectors);
+        SimCompare.checkIverilogVector(mod, vectors);
+      });
```
desmonddak, line 1894: Need a test showing a partial assignement is not collapsed.

desmonddak:
I think it makes sense to commit this PR for cleaning up the major cases.  Two main arch points to prepare for is the naming problem, which we tried to solve in central_naming by marking first in the synth base classes, and then another is to think about what will be common optimizations versus SV specific.

Unfortunately a netlist branch is not yet in and that has a really general approach to deal with nested assignments and series of pack/unpack/pack as well as subset/concat/subset which can occur in a nested fashion (a very wide interface split into subsets and then recombined in a couple levels possibly with some splits and recombines and finally end up at the very same interface.  Then mixed array and struct assignments, reversals, etc. all get very tricky, especially with partial assignments and subset assignments.  Worth thinking about a general optimization approach along with the naming issue.  Nested Interface assignments containing busses is the source of these.

test/array_collapsing_test.dart
mkorbel1: you're right, i added a test with a skip and explanation.

test/array_collapsing_test.dart
mkorbel1: nice test, added!

test/array_collapsing_test.dart
mkorbel1: had a test that covered this partially, improved it, thanks!

mkorbel1:
Thank you for the great feedback @desmonddak !  I think it's all addressed now so I'll merge once CI passes again.

I agree there's room for improvement in terms of generality and applicability to future synth stacks, so we should keep that in mind as we merge more things in.  But at least this is a big improvement for now and has good regression tests to help protect future architectural reuse and improvements.

PR #665, opened 2026-06-06
Devtools

desmonddak:
## Description & Motivation

A baseline for a ROHD design debugger: this version has a strong ability to connect to DTD and multiple VMs, and has reconnection and state restoration if we restart the debugger.

It currently will display the latest values in the Details pane, demonstrating a rudimentary connection to the simulator via DTD/VM.

## Related Issue(s)

Replaces PR #657

## Testing

Tested web build version as well as Devtools in Browser versions.

## Backwards-compatibility

No

## Documentation

Yes.  Help popups are enabled to describe features currently available.


desmonddak:
This is semi-usable for debug, allowing for hierarchy traversal, port listings, and live value display.

You can build the full devtools version, which installs in `extension/` locally.
```sh
./tool/gh_actions/devtools/install_devtool.sh
```
Then you can use the VSCode command `Run Devtools in Browser` and debug a live ROHD test being debugged in the current VSCode session because the details pane will maintain live port values for whatever level of hierarchy you want to observe.

If you want to debug a remote VSCode session, you can use the tasks that are built into the rohd_devtool_extension area or you can build it yourself.  Either way you need to be in the `rohd_devtool_extension/` subdirectory:
```sh
make web
```
and then use Flutter to start up the application (see the `.vscode/tasks.json` for the exact command).

Once started, you paste in the DTD address (or the port-forwarded version) and it will discover the VM debug sessions in that VSCode session, link, and retrieve the design hierarchy and current values (updated live).

This is a baseline framework to which we will add schematic viewing and live waveforms.

tool/gh_actions/devtool/install_devtools.sh
```diff
@@ -0,0 +1,95 @@
...
+#   bash tool/gh_actions/devtool/install_devtools.sh
+#
+# 2024 January 03
+# Author: Yao Jing Quek <yao.jing.quek@intel.com>
+
+set -euo pipefail
+
+DEST="../extension/devtools"
+
+# ── Helper: strip service-worker / PWA artifacts from a deployed build ──
```
mkorbel1: This looked weird and like a lot to me.  Here's what copilot thinks:

I’m not convinced we should handle this with this much post-build rewriting.

The service-worker concern makes sense for a DevTools extension, since stale Flutter web caches can cause confusing behavior. But this implementation feels fragile: it rewrites generated `flutter_bootstrap.js` with a regex, injects runtime unregister scripts into generated `index.html`, deletes cache storage broadly from the service worker, and forces clients to navigate. Those are fairly invasive behaviors for a build/deploy script, and they may break or behave differently when Flutter changes its generated bootstrap format.

Could we first remove the PWA bits at the source/build level instead? For example, remove the manifest link from `rohd_devtools_extension/web/index.html`, avoid shipping `manifest.json`, and use Flutter/devtools extension build options to disable service-worker generation if available. If we need migration cleanup for users who already have a stale service worker cached, I’d prefer a much smaller, documented compatibility step rather than ongoing regex edits against generated Flutter internals.

doc/tutorials/chapter_9/rohd_vf_example/lib/counter.dart
mkorbel1, line 1: it looks like this was a bad revert to an old version, rm these changes?

rohd_devtools_extension/lib/main_standalone.dart
mkorbel1, line 1: This is really cool to have a standalone version!  It might be worth adding to the rohd_devtools_extension/README.md about it (and in general updating the README since its very out of date).

rohd_devtools_extension/assets/icons/rohd_logo.png
mkorbel1, line 1: Why not use the ROHD "R" logo?

rohd_devtools_extension/lib/rohd_devtools/cubit/details_tab_cubit.dart
```diff
@@ -0,0 +1,31 @@
+// Copyright (C) 2025 Intel Corporation
```
mkorbel1: check copyright years

rohd_devtools_extension/lib/rohd_devtools/services/vm_service_signal_value_source.dart
```diff
@@ -0,0 +1,376 @@
...
+    if (widthValue is int) {
+      return widthValue;
+    }
+    if (widthValue is String) {
+      return int.tryParse(widthValue) ?? 1;
+    }
+    return 1;
+  }
+
+  void _rememberTime(int timePs) {
```
mkorbel1: rohd time is not really "picoseconds" so much as unitless time measurement -- should we be maybe be careful not to make unit assumptions?

rohd_devtools_extension/lib/rohd_devtools/ui/signal_details_card.dart
```diff
@@ -7,123 +7,149 @@
...
-                mainAxisSize: MainAxisSize.min,
-                children: <Widget>[
+    unawaited(showDialog(
+        context: context,
+        builder: (context) => StatefulBuilder(
+            builder: (context, setState) => AlertDialog(
+                title: const Text('Filter Signals'),
+                content:
+                    Column(mainAxisSize: MainAxisSize.min, children: <Widget>[
                   CheckboxListTile(
```
mkorbel1, line 88: while we're at it, we need to add inout in here?  (comment applies to all the signal stuff)

rohd_devtools_extension/pubspec.yaml
```diff
@@ -10,27 +10,36 @@ environment:
...
+    sdk: flutter
   flutter_bloc: ^9.0.0
+  flutter_simple_treeview: ^3.0.2
+  logging: ^1.2.0
+  rohd:
+    path: ../../rohd
+  rohd_devtools_widgets:
+    path: packages/rohd_devtools_widgets
+  vm_service: ^14.3.1
+  web: ^1.0.0
```
mkorbel1, line 28: reminder to ourselves: add these to list of new dependencies

rohd_devtools_extension/packages/rohd_devtools_widgets/pubspec.yaml
```diff
@@ -0,0 +1,13 @@
+name: rohd_devtools_widgets
+description: "Shared widgets for ROHD DevTools sub-packages: help button, PNG export, auto-hiding overlay."
```
mkorbel1, line 2: question: why is this a separate package?

if it's because this stuff is reused across other flutter views like the wave viewer and schematic viewer then I get it and it makes sense, but then we should probably at least add a README and stuff to the package right?

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/logic_type_utils.dart
```diff
@@ -0,0 +1,387 @@
...
+      result.write('x');
+    }
+  }
+  return result.toString();
+}
+
+/// Convert a hex value string (e.g. "0x1a3f" or "1a3f") to binary (MSB-first).
+///
+/// Returns null if the input can't be parsed.
+String? hexToBinary(String hexValue, int width) {
```
mkorbel1, line 296: feels like this exists in LogicValue already?

rohd_devtools_extension/analysis_options.yaml
mkorbel1, line 1: why delete?

rohd_devtools_extension/packages/rohd_devtools_widgets/pubspec.yaml
desmonddak: Yes, the wave and schematic viewers are reusing elements of this widgets package.  I'll look into this.  I have experimented with publishing these, and I think README.md and LICENSE were needed. I'll bring those in here.

doc/tutorials/chapter_9/rohd_vf_example/lib/counter.dart
desmonddak: Done.

rohd_devtools_extension/assets/icons/rohd_logo.png
desmonddak: Done

tool/gh_actions/devtool/install_devtools.sh
desmonddak: I removed this by removing the manifest.json and updating index.html and tested (the last paragraph suggestion).
It is working.

rohd_devtools_extension/lib/rohd_devtools/cubit/details_tab_cubit.dart
desmonddak: Done

rohd_devtools_extension/lib/rohd_devtools/ui/signal_details_card.dart
desmonddak: Done. I checked downstream changes as well (future branches)

rohd_devtools_extension/packages/rohd_devtools_widgets/pubspec.yaml
desmonddak: Added LICENSE and README.md.

rohd_devtools_extension/analysis_options.yaml
desmonddak: This analysis_options.yaml was not checking most of the rules that we used for ROHD.  I found a way to make it work across this divide, but it ended up making ROHD a flutter app.  So I reverted but added back all the rules that we use in ROHD so the code is of consistent quality.

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/logic_type_utils.dart
desmonddak: I call into LogicValue now.  The issue is that this is string level manipulation of values coming from the wave viewer (later PR).  So we reuse the complex functionality in LogicValue with wrappers.

rohd_devtools_extension/lib/main_standalone.dart
desmonddak: Updated the README.md.
The standalone version is a bit more useful as you can send it the URI from a separate VSCode session and pick the VM you want to debug.  Devtools is more fixed: you are required to use the DTD from the current VSCode and it picks the first VM it can find.

rohd_devtools_extension/lib/rohd_devtools/services/vm_service_signal_value_source.dart
desmonddak: I fixed this.  I think at some point we need to know the time measurement for the wave display, but for now it is unitless.

rohd_devtools_extension/pubspec.yaml
desmonddak: Many more are coming in later PRs.  We will resolve at that point.

rohd_devtools_extension/packages/rohd_devtools_widgets/pubspec.yaml
```diff
@@ -7,6 +7,8 @@ environment:
   flutter: '>=3.0.0'
 dependencies:
   flutter: {sdk: flutter}
+  rohd:
```
mkorbel1: question: do you think the path dependency should be in `dependency_overrides` instead so that when other packages depend on this `rohd_devtools_widgets` package they dont get a dependency error indicating that it has a path dependency?  i.e. it should just say the normal `rohd` dependency in the `dependencies`, since `dependency_overrides` doesn't get inherited by dependers

tool/gh_actions/devtool/build_web.sh
mkorbel1, line 1: i understand that this script is replaced by install_devtools.sh, but I still need the actual output generated somewhere that i can include in a release.  where do I get that from?  don't i still want a script like this one for the build_and_copy or is the install_devtools output reliable enough to release as part of the rohd package even though it's not using the standard building approach?

desmonddak: I may not fully understand your concern:   install_devtools.sh is running just like build_web.sh was -- it is installing the devtools extension in releasable form as part of the 'General  Build Devtools'  workflow, which is passing.  It is a better name for the same script, and it has been updated to include all the assets required by the latest devtools extension.

desmonddak: Since `Build Devtools` workflow runs, is there a way to test this release package the way you expect it to be used?

mkorbel1: hmm, well i know how to manually test it:
when we build it, it should show up in the proper build output area for the repo such that if another repo points at it, the devtools extension is automatically discoverable without additional configuration.

i don't know if there's a nice way to automatically test that?  maybe we dont need an additional repo and can just use the rohd repo itself?  maybe there's a CLI method to query available devtools extensions?

rohd_devtools_extension/packages/rohd_devtools_widgets/pubspec.yaml
desmonddak: It should probably point to ^0.6.9 so that people can use a compatible ROHD version?

rohd_devtools_extension/packages/rohd_devtools_widgets/pubspec.yaml
```diff
@@ -7,8 +7,7 @@ environment:
   flutter: '>=3.0.0'
 dependencies:
   flutter: {sdk: flutter}
-  rohd:
-    path: ../../..
+  rohd: ^0.6.9
```
mkorbel1, line 10: yes, but you should *also* add a path `dependency_overrides` for rohd probably below so that the tests operate on the actual rohd you're developing

tool/gh_actions/devtool/build_web.sh
desmonddak:
```shell
dart test_devtools_install.dart <path>  
```
where ```<path>``` can be any repo, or even the github artifacts branch.

This is under the rohd_devtools_extension/tools area and is invoked as part of the devtools installation workflow.

mkorbel1 (approved):
Looks great, i think ready!

tool/gh_actions/devtool/build_web.sh
mkorbel1: very cool!

PR #667, opened 2026-06-14
keyring-style dart installation rather than holding keys

### LINKED ISSUE #666: Codespaces dart installation fails due to old key mechanism
### Describe the bug

We see a failure in the CI for codespaces installation due to using an old dart key locally.

### To Reproduce

See recent CI logs.

### Expected behavior

Copilot suggests the following `~/tool/gh_codespaces/install_dart.sh` which is undergoing test in our central_naming PR #652 :
```
#!/bin/bash
set -euo pipefail

sudo apt-get update
sudo apt-get install -y wget gpg apt-transport-https

sudo mkdir -p /usr/share/keyrings
wget -qO- https://dl-ssl.google.com/linux/linux_signing_key.pub \
  | gpg --dearmor \
  | sudo tee /usr/share/keyrings/dart.gpg >/dev/null

echo "deb [signed-by=/usr/share/keyrings/dart.gpg] https://storage.googleapis.com/download.dartlang.org/linux/debian stable main" \
  | sudo tee /etc/apt/sources.list.d/dart_stable.list

sudo apt-get update
sudo apt-get install -y dart
```

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

```yaml

```

### Additional: Context

_No response_

desmonddak:
## Description & Motivation

The key we have is old and a solution is to use a keyring approach to install dart.

## Related Issue(s)

#666 

## Testing

I tested as part of PR #652 but cherry-picking this change for review and will be needed for all PRs to pass CI.

## Backwards-compatibility

No

## Documentation

None needed.  This is an infrastructure scripting change.


mkorbel1 (commented):
In https://github.com/intel/rohd/pull/256 it was indicated that Dart team recommended saving the key locally.  From some quick search it seems like this can help mitigate man-in-the-middle attacks and reduce reliance on more servers as part of CI.  Practically speaking, the occasional work this causes seems maybe to outweigh the risks here.  I.e. I'm inclined to  accept this solution.  Thoughts?

desmonddak:
Respecting https://github.com/intel/rohd/pull/256, I put back the local key cache with fallback to wget upon failure when the key is stale with a loud warning in the logs.  That way, we do not have a failing CI when the key changes.

mkorbel1 (approved):

PR #668, opened 2026-06-14
new dart analyzer failure with bad override

desmonddak:
## Description & Motivation

New PRs cannot run with new dart version because of old override that is a failure in lint

## Related Issue(s)

None

## Testing

Ran analyze under new dart

## Backwards-compatibility

No

## Documentation

This is a single line change in the tutorial section which uses an override spuriously.


desmonddak:
This has been merged into PR #667 because both fixes are needed to get CI to run again.

mkorbel1 (approved):

desmonddak:
Subsumed by #667

PR #669, opened 2026-06-22
Module services

desmonddak:
## Description & Motivation

This is a new API for dealing with services after a module has been built.  Rather than adding more methods to `Module`, we create services that can operate on a built `Module`, like writing out waveforms, or synthesizing outputs like SystemVerilog.

So instead of `'module.generateSynth(outputPath: 'build/${module.definitionName}.sv')`  for outputing SystemVerilog and `WaveDumper(module, outputPath: 'build/${module.definitionName}.vcd')` for outputing Waveforms, we use:

```dart
  final sv = SvService(module, outputPath: 'build/${module.definitionName}');
  final waves = WaveformService(module, outputPath: 'build/${module.definitionName}.vcd');
  ```
  
  As we add more services, this is a more consistent API with fewer surprises.
   

## Related Issue(s)

None

## Testing

Tests are converted to the new Service apis and run through.

## Backwards-compatibility

No.  But eventually, we would like to deprecate existing methods in favor of a services model.

## Documentation

Small changes in the README.md and throughout the tutorial and tests.

### CODE REVIEW

File: lib/src/diagnostics/module_service.dart [hunk 1]
```diff
@@ -0,0 +1,82 @@
...
+
+  /// A JSON-serialisable summary of this service.
+  Map<String, Object?> toJson();
+}
+
+/// A [ModuleService] that emits output to one or more files.
+///
+/// Establishes the common output convention shared by synthesis, netlist,
+/// trace, and waveform services:
+///  - [outputPath] — the default file or directory written by [write].
```
mkorbel1, line 36: i like this API because it's easy to understand, but should we be using some sort of URI object or something from Dart instead that supports referencing files, directories, and other weird stuff instead?  im not sure the answer, genuine question

lib/src/diagnostics/module_service.dart
```diff
@@ -0,0 +1,82 @@
...
+      writeOutputTextFile(path, contents);
+}
+
+/// An [OutputService] that generates source-code text, keyed per module
+/// definition.
+///
+/// Shared by the language code-generation services (e.g. SystemVerilog and
+/// SystemC), which all produce a combined single-file [output] as well as
+/// per-definition contents.
+abstract class CodegenService extends OutputService {
```
mkorbel1, line 70: nit: should it be `CodeGenService`?

lib/src/diagnostics/module_service.dart
```diff
@@ -0,0 +1,82 @@
...
+  /// The combined single-file generated output (including any header).
+  String get output;
+
+  /// The generated output keyed by module definition name
+  /// ([Module.definitionName]).
```
mkorbel1, line 75: what if one definition generates multiple files?  SynthBuilder supports that now

lib/src/synthesizers/systemverilog/sv_service.dart
```diff
@@ -0,0 +1,194 @@
...
+/// await dut.build();
+/// final sv = SvService(dut);
+///
+/// // Write individual .sv files:
+/// sv.writeFiles('build/');
+///
+/// // Or get the concatenated output (like generateSynth):
+/// print(sv.allContents);
+/// ```
+class SvService extends CodegenService {
```
mkorbel1, line 35: question: what do you think about `SystemVerilogService` instead?  more consistent with other stuff?

lib/src/diagnostics/module_services.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  /// The singleton instance.
+  static final ModuleServices instance = ModuleServices._();
+
+  // ─── Hierarchy (auto-registered by Module.build) ──────────────
+
+  /// The most recently built top-level [Module].
+  ///
+  /// Set automatically at the end of [Module.build].
+  Module? rootModule;
```
mkorbel1, line 37: should this be `@internal` or something?

lib/src/wave_dumper.dart
mkorbel1, line 1: it feels like this has a lot of duplicate code with the existing waveform_service?  should this just use the service under the hood and be deprecated then?

mkorbel1:
(those were pending comments, figured i'd submit them for now, but i'm also adding relevant ones to https://github.com/intel/rohd/pull/673 as well)

desmonddak:
Obviated by a different split amongst PRs for source_debug and module_services_api branches.

PR #671, opened 2026-06-22
Fst Writer_03

desmonddak:
## Description & Motivation

It would be nice to have a `WaveDumper` that can output the FST format because it is much more dense and can be incrementally accessed.

## Related Issue(s)

None

## Testing

We ran the output through a few tools, notably Wellen and GTKwave.

## Backwards-compatibility

No.

## Documentation

No.  We will need to add.  But at the user level, it becomes a choice between VCD and FST>


lib/src/fst/fst_types.dart
```diff
@@ -0,0 +1,236 @@
+// Copyright (C) 2021-2026 Intel Corporation
```
mkorbel1: nit: new file not copyrighted since 2021?

lib/src/wave_dumper.dart
```diff
@@ -15,27 +17,52 @@ import 'package:rohd/src/utilities/sanitizer.dart';
...
-  final File _outputFile;
+  /// The waveform output format (VCD or FST).
+  final WaveFormat format;
+
+  /// The FST writer configuration (only used when [format] is
+  /// [WaveFormat.fst]).
+  final FstWriterConfig? fstConfig;
+
+  /// The file to write dumped output waveform to (VCD only).
+  File? _outputFile;
```
mkorbel1: why no longer final?

lib/src/wave_dumper.dart
```diff
@@ -93,7 +133,81 @@ class WaveDumper {
   /// write contents to the output file.
   static const _fileBufferLimit = 100000;
 
-  /// Buffers [contents] to be written to the output file.
+  // ─────────────── VCD initialization ───────────────
+
+  /// Initializes VCD output.
+  void _initVcd() {
```
mkorbel1: it feels a little odd that all the VCD-specific stuff lives in wave_dumper, but all the FST-specific stuff is in a separate FSTWriter.  Should we extract a common API of some sort for "writers" rather than have such a difference in implementation?

lib/rohd.dart
```diff
@@ -4,6 +4,8 @@
 export 'src/exceptions/exceptions.dart';
 export 'src/external.dart';
 export 'src/finite_state_machine.dart';
+export 'src/fst/fst_types.dart';
```
mkorbel1, line 8: update copyright year

test/fst_writer_test.dart
mkorbel1, line 1: it might be nice to have our vcd tests and fst tests somehow both parse the same simulation and confirm that both saw certain events at the same times

lib/src/wave_dumper.dart
desmonddak: I am applying this comment to fst-writer which will have the `WaveformService` form of API.  This branch was designed to be mergeable independent of module_services_api as I wasn't sure when we would merge.  FST is interesting for exploring waveforms with structure (array/struct).

desmonddak: Actually it just seems easier to merge this after module_services_api.  So I will presume this from now on to make it easier to review.

lib/src/wave_dumper.dart
desmonddak: No longer relevant with refactoring to redirect to `WaveformService`.

lib/src/wave_dumper.dart
desmonddak: No longer here.

lib/rohd.dart
desmonddak: Done

test/fst_writer_test.dart
desmonddak: in waveform_service_test.dart:111:  VCD and FST contain matching value-change events'

test/wave_dumper_test.dart
mkorbel1, line 1: should we make tests that dump FST, convert to VCD, then confirm the right values at the right time like we already do for VCD?  and/or some other way to make sure that actual simulations make the right values at the right time?  i think some of these VCD checking tests were catching actual tricky spots.

mkorbel1 (commented):
i reviewed some of the FST-specific pieces, generally looks good, but I'll defer full review until the common pieces in prior PRs are merged first

PR #673, opened 2026-06-25
Module services api_02

desmonddak:
## Description & Motivation

This is a subset of the PR #669 that introduces the new API for Services attached to a module (outputters like wave dumpers, netlisters, translators, tracing services, etc).

## Related Issue(s)

None
## Testing

This runs a full dart test with the new API in place in the tests.

## Backwards-compatibility

No, we retain the old interfaces but they should eventually be deprecated.  
WaveDumper
generateSynth
...

## Documentation

Some examples have been modified to match.


lib/src/utilities/simcompare.dart
mkorbel1, line 1: this file has the comma formatting thing, can't tell if there's actually anything new in here?

lib/src/diagnostics/module_service.dart
```diff
@@ -0,0 +1,71 @@
...
+}
+
+/// A [ModuleService] that emits output to one or more files.
+///
+/// Establishes the common output convention shared by synthesis, netlist,
+/// trace, and waveform services:
+///  - [outputPath] — the default file or directory written by [write].
+///  - [multiFile] — whether [write] emits one file per module definition
+///    (a directory) or a single combined file.
+///  - [write] — performs the write, honouring [multiFile].
```
mkorbel1, line 36: i like this API because it's easy to understand, but should we be using some sort of URI object or something from Dart instead that supports referencing files, directories, and other weird stuff instead? im not sure the answer, genuine question

lib/src/diagnostics/module_service.dart
```diff
@@ -0,0 +1,71 @@
...
+  /// omitted.
+  void write([String? path]);
+}
+
+/// An [OutputService] that generates source-code text, keyed per module
+/// definition.
+///
+/// Shared by language code-generation services, which all produce a combined
+/// single-file [output] as well as per-definition contents.
+abstract class CodegenService extends OutputService {
```
mkorbel1: nit: should it be `CodeGenService`?

lib/src/diagnostics/module_service.dart
```diff
@@ -0,0 +1,71 @@
...
+  /// The combined single-file generated output (including any header).
+  String get output;
+
+  /// The generated output keyed by module definition name
+  /// ([Module.definitionName]).
+  Map<String, String> get contentsByDefinitionName;
+
+  /// The generated output for a single module [definitionName], or `null` when
+  /// that definition was not generated.
+  String? moduleOutput(String definitionName) =>
```
mkorbel1: what if one definition generates multiple files? SynthBuilder supports that now

lib/src/synthesizers/systemverilog/sv_service.dart
```diff
@@ -0,0 +1,201 @@
...
+/// await dut.build();
+/// final sv = SvService(dut);
+///
+/// // Write individual .sv files:
+/// sv.writeFiles('build/');
+///
+/// // Or get the concatenated output (like generateSynth):
+/// print(sv.allContents);
+/// ```
+class SvService extends CodegenService {
```
mkorbel1: question: what do you think about `SystemVerilogService` instead? more consistent with other stuff?

lib/src/diagnostics/module_services.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  /// The singleton instance.
+  static final ModuleServices instance = ModuleServices._();
+
+  // ─── Hierarchy (auto-registered by Module.build) ──────────────
+
+  /// The most recently built top-level [Module].
+  ///
+  /// Set automatically at the end of [Module.build].
+  Module? rootModule;
```
mkorbel1: should this be `@internal` or something?

lib/src/diagnostics/module_services.dart
```diff
@@ -0,0 +1,73 @@
...
+
+  /// Returns the module hierarchy as a JSON string.
+  ///
+  /// DevTools evaluates this via `EvalOnDartLibrary` to display the module
+  /// hierarchy.  Richer design views (e.g. a slim netlist) are composed by the
+  /// DevTools client from the relevant registered service.
+  String get hierarchyJSON {
```
mkorbel1: this seems like an odd implementation, is this temporary?  like why does calling the service on something change which module is the top?

lib/src/diagnostics/waveform_service.dart
```diff
@@ -0,0 +1,435 @@
...
+      for (final subm in m.subModules) {
+        if (subm is InlineSystemVerilog) {
+          continue;
+        }
+        modulesToParse.add(subm);
+      }
+    }
+  }
+
+  // ─── VCD output helpers ───────────────────────────────────────
```
mkorbel1, line 301: it feels like this has a lot of duplicate code with the existing `WaveDumper`? should it just use this service under the hood and be deprecated then?

lib/src/modules/conditionals/flop.dart
```diff
@@ -88,6 +88,11 @@ class FlipFlop extends Module with SystemVerilog {
   /// Only initialized if a constant value is provided.
   late LogicValue _resetValueConst;
 
+  /// Returns the constant reset value if one was provided, or null if the
+  /// reset value is a port or no reset exists.
+  LogicValue? get constantResetValue =>
```
mkorbel1: ok, but why add?

lib/src/modules/conditionals/flop.dart
mkorbel1, line 1: update copyright

lib/src/modules/conditionals/sequential.dart
```diff
@@ -135,6 +135,14 @@ class Sequential extends Always {
   /// The input edge triggers used in this block.
   final List<_SequentialTrigger> _triggers = [];
 
+  /// Returns the edge polarity for each trigger input port.
+  ///
+  /// Each entry pairs the trigger input port name with whether the trigger
+  /// fires on a positive edge (`true`) or negative edge (`false`).
+  List<({String portName, bool isPosedge})> get triggerEdges => _triggers
```
mkorbel1: why?

lib/src/synthesizers/systemverilog/system_verilog_service.dart
```diff
@@ -0,0 +1,201 @@
...
+        result[defName] = fc.contents;
+      }
+    }
+    return result;
+  }
+
+  /// Writes each module's SV to a separate file in [directory].
+  ///
+  /// Files are named `<definitionName>.sv`.
+  void writeFiles(String directory) {
```
mkorbel1, line 163: rohd-bridge also generates a filelist.f, maybe worth considering? see:
https://github.com/intel/rohd-bridge/blob/5cba1f5c9df2562c54ce55279a73ee7d71b48cc6/lib/src/bridge_module.dart#L991

lib/rohd.dart
```diff
@@ -1,6 +1,9 @@
 // Copyright (C) 2021-2023 Intel Corporation
```
mkorbel1: copyright years (and in general)

lib/rohd.dart
```diff
@@ -1,6 +1,9 @@
...
 // SPDX-License-Identifier: BSD-3-Clause
 
+export 'src/diagnostics/module_service.dart';
```
mkorbel1: barrel file for the diagnostics?

tool/gh_actions/check_tmp_test.sh
```diff
@@ -13,9 +13,13 @@ set -euo pipefail
 
 declare -r folder_name='tmp_test'
 
-# The "tmp_test" folder after performing the tests should be empty.
+# The "tmp_test" folder after performing the tests should be empty,
+# except for the precompiled-header cache (pch/) which is intentionally
```
mkorbel1: if it's intentionally persistent why is it in tmp_test?  should we have a separate gitignored build/ dir?

tool/gh_actions/run_tests.sh
```diff
@@ -11,8 +11,8 @@
 
 set -euo pipefail
 
-dart test
+dart test $(find test -name '*_test.dart' | sort)
```
mkorbel1: is this the right way to do this or should we be running `dart test` in all the packages one at a time?

dart_test.yaml
mkorbel1, line 1: which files/tests need ffi?

README.md
```diff
@@ -9,7 +9,6 @@
 [![Chat](https://img.shields.io/discord/1001179329411166267?label=Chat)](https://discord.gg/jubxF84yGw)
 [![License](https://img.shields.io/badge/License-BSD--3-blue)](https://github.com/intel/rohd/blob/main/LICENSE)
 [![Contributor Covenant](https://img.shields.io/badge/Contributor%20Covenant-2.1-4baaaa.svg)](https://github.com/intel/rohd/blob/main/CODE_OF_CONDUCT.md)
-[![Coverage](https://raw.githubusercontent.com/intel/rohd/refs/heads/badges/coverage/main.svg)](https://github.com/intel/rohd/blob/main/.github/workflows/coverage.yml)
```
mkorbel1, line 12: why remove coverage?

lib/src/diagnostics/module_service.dart
desmonddak: Ack.  What does this mean?  I am concerned that I need to look up a symbol from a module, and not sure how to locate the file it is in.  

Is this header/source or something more than that split?

lib/src/synthesizers/systemverilog/sv_service.dart
desmonddak: That is probably better.  I have ScService, I hope SystemCService is ok.

tool/gh_actions/check_tmp_test.sh
desmonddak: Just to span multiple tests -- needs to be deleted after tests are run.  I'll confirm the flow is efficient and cleanup is more clear.

tool/gh_actions/run_tests.sh
desmonddak: It is the wrong way to do things.  I will fix.

dart_test.yaml
desmonddak: SystemC tests.

lib/src/diagnostics/module_service.dart
mkorbel1: I think this was preparing for things like IPXACT or UPF, I forget what exactly.  It isn't always necessarily an HDL equivalent, it might be metadata or different aspects of description.  I don't know the solution really for what you're building to support multiple files, but it's worth considering in case there is a future output which is multiple files just to make sure our APIs may support it in case we ever want to?

dart_test.yaml
mkorbel1: should we keep that for the systemc PR then?

desmonddak: Yup.  Several things here should be pushed there.  I had lifted module_services_api to precede systemc and some artifacts were left here (like PCH handling).  I'm pushing them to systemc and taking the comments along with them to fix them there.

README.md
desmonddak: I think I experimented with local coverity and them removed and accidentally removed this badge.

lib/src/modules/conditionals/flop.dart
desmonddak: Moved to SystemC branch where this makes more sense.

lib/src/modules/conditionals/flop.dart
desmonddak: Not necessary in this repo now as we are not making a change.

lib/src/modules/conditionals/sequential.dart
desmonddak: Moving to systemc branch for consideration there.

lib/src/utilities/simcompare.dart
desmonddak: Aligned formatting back to original

lib/src/diagnostics/module_service.dart
desmonddak: Yup -- updated

lib/src/synthesizers/systemverilog/sv_service.dart
desmonddak: Changed services to longer names.

lib/src/diagnostics/module_services.dart
desmonddak: Changed to setter/getter for potential side-effects

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: Filed issue #683

lib/rohd.dart
desmonddak: Done

tool/gh_actions/check_tmp_test.sh
desmonddak: Migrated PCH testing and documentation to systemc branch

lib/src/diagnostics/module_services.dart
desmonddak: It has been fixed to just return the JSON (a downstream fix was pending).

lib/src/diagnostics/waveform_service.dart
desmonddak: Done.  I wasn't sure if we were yet happy with this API so I left WaveDumper as is.  But now we can deprecate with the presumption we are moving to this Services model of API.

lib/src/diagnostics/module_service.dart
desmonddak: I propose to defer this and create a more complex abstract base class for more complex generators.

lib/src/diagnostics/module_service.dart
desmonddak: I filed this as Issue #684

lib/src/diagnostics/module_services.dart
mkorbel1: more generally what i meant is that should it be an open API for setting `rootModule` or an `@internal` one?

lib/src/module.dart
```diff
@@ -4,7 +4,7 @@
 // module.dart
 // Definition for abstract module class.
 //
-// 2021 May 7
+// 2026 July
```
mkorbel1: don't change authorship year? (applies to multiple files)

lib/src/synthesizers/systemverilog/system_verilog_service.dart
```diff
@@ -0,0 +1,201 @@
...
+/// await dut.build();
+/// final sv = SystemVerilogService(dut);
+///
+/// // Write individual .sv files:
+/// sv.writeFiles('build/');
+///
+/// // Or get the concatenated output (like generateSynth):
+/// print(sv.allContents);
+/// ```
+class SystemVerilogService extends CodeGenService {
```
mkorbel1, line 34: Something doesn't feel quite right to me in this API.  Some thoughts:
- If I want to generate some SV (pretty common), why am I constructing a "service"?  Though the abstraction levels as a service make sense to me.
- Why does the act of constructing said service immediately produce the output?
- In what cases do I want to run the SynthBuilder with a selected synthesizer instead of this service?
- When there are multiple synthesizers (e.g. SystemC), will I have multiple services or one service that takes multiple synthesizers?

The two APIs currently are basically:
- Turn this module into SV quickly (generateSynth function on module)
- Make a thing that builds an output based on this Synthesizer and give me outputs I can do more work with (SynthBuilder with Synthesizer arg)

This is deprecating the first, but keeping the second, and there's still some API overlap.

I think this is close to right, but something is a little off.  Thoughts?

lib/src/synthesizers/systemverilog/system_verilog_service.dart
```diff
@@ -0,0 +1,201 @@
...
+
+  /// The combined single-file generated output (alias for [synthOutput]).
+  @override
+  String get output => synthOutput;
+
+  /// Returns a map from module definition name to its SV file contents.
+  ///
+  /// Keys are [SynthesisResult.instanceTypeName] (the uniquified definition
+  /// name used in the generated SV).
+  Map<String, String> get contentsByName => {
```
mkorbel1, line 156: why do we need this when a list of fileContents already has the info needed?

lib/src/synthesizers/systemverilog/system_verilog_service.dart
```diff
@@ -0,0 +1,201 @@
...
+  /// Returns the concatenated SystemVerilog module definitions as a single
+  /// string, without the generation header.
+  ///
+  /// For the full output with header (matching `Module.generateSynth()`),
+  /// use [synthOutput].
+  String get allContents =>
+      fileContents.map((fc) => fc.contents).join(moduleSeparator);
+
+  /// The ROHD generation header prepended to single-file output.
+  String get synthHeader => '''
```
mkorbel1, line 114: since we're revisiting all this -- why does the header only come when it's all in 1 file?  what if you don't want the header?  what if you do want the header even across many files?

lib/src/synthesizers/systemverilog/system_verilog_service.dart
```diff
@@ -0,0 +1,201 @@
...
+        for (final fc in fileContents) fc.name: fc.contents,
+      };
+
+  /// Returns a map from module definition name
+  /// ([Module.definitionName]) to its SV file contents.
+  ///
+  /// This uses the original definition name (not uniquified), matching
+  /// the keys used by FLC trace data.
+  @override
+  Map<String, String> get contentsByDefinitionName {
```
mkorbel1: this isn't really reliable since multiple things could have the same original definitionName, right?

lib/src/synthesizers/systemverilog/system_verilog_service.dart
```diff
@@ -0,0 +1,201 @@
...
+
+  /// Creates a [SystemVerilogService] for [module].
+  ///
+  /// [module] must already be built.
+  ///
+  /// If [outputPath] is provided, output is written immediately: a directory
+  /// of per-module files when [multiFile] is `true`, otherwise the
+  /// concatenated SV output (with header) to that single file.
+  SystemVerilogService(this.module,
+      {bool register = true,
```
mkorbel1: undocumented arg, what does register do?  what's the implication of registering it or not?

benchmark/many_submodules_benchmark.dart
mkorbel1, line 1: check copyright years on all files

doc/tutorials/chapter_2/helper.dart
```diff
@@ -13,7 +13,8 @@ import 'package:rohd/rohd.dart';
 
 Future<void> displaySystemVerilog(Module mod) async {
   await mod.build();
-  print('\nYour System Verilog Equivalent Code: \n ${mod.generateSynth()}');
+  print('\nYour System Verilog Equivalent Code: \n '
+      '${SystemVerilogService(mod, register: false).synthOutput}');
```
mkorbel1: this is example code for people learning, why register:false?

lib/src/diagnostics/module_services.dart
```diff
@@ -0,0 +1,77 @@
...
+  Module? _rootModule;
+
+  /// The most recently built top-level [Module].
+  ///
+  /// Set automatically at the end of [Module.build].
+  Module? get rootModule => _rootModule;
+
+  set rootModule(Module? value) {
+    _rootModule = value;
+    ModuleTree.rootModuleInstance = value;
```
mkorbel1: is there a reason for a separate `ModuleTree` still?

lib/src/module.dart
```diff
@@ -1147,19 +1149,9 @@ abstract class Module {
...
- */
-
-''';
-    return synthHeader +
-        SynthBuilder(
-          this,
-          SystemVerilogSynthesizer(configuration: configuration),
-        ).getSynthFileContents().join('\n\n////////////////////\n\n');
+    return SystemVerilogService(this,
+            register: false, configuration: configuration)
```
mkorbel1: why register:false?

test/fsm_test.dart
```diff
@@ -183,7 +183,7 @@ void main() {
     final mod = TestModule(Logic(), Logic(), Logic());
     await mod.build();
 
-    final sv = mod.generateSynth();
```
mkorbel1, line 186: at least one test should still call generateSynth to preserve the functionality (we are deprecating not deleting)

lib/src/synthesizers/systemverilog/system_verilog_service.dart
```diff
@@ -0,0 +1,201 @@
...
+ */
+
+''';
+
+  /// Returns the full single-file SystemVerilog output with header,
+  /// identical to `Module.generateSynth()`.
+  ///
+  /// Computed once and cached so the timestamped header is stable for the
+  /// lifetime of this service.
+  late final String synthOutput = synthHeader + allContents;
```
mkorbel1: why `synthOutput`?  What's wrong with just the equivalent `output` that matches consistent API?

test/waveform_service_test.dart
mkorbel1, line 1: why not update this file to use the new API (and maintain one check that deprecated API lives)

tool/gh_actions/run_tests.sh
```diff
@@ -1,18 +1,27 @@
...
 
+# Run main package tests (auto-discovers test/)
 dart test
 
-# run tests in JS (increase heap size also)
+# Run tests in sub-packages
+for pkg in packages/*/; do
+  if [ -d "${pkg}test" ]; then
+    echo "Running tests in ${pkg}..."
+    (cd "$pkg" && dart pub get && dart test)
```
mkorbel1: why is pub get in here?

lib/src/module.dart
desmonddak: Well, that was easy for AI.  Fixed.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: Let me ponder this.  

I have `NetlistService`, `SystemVerilogService`, `SystemCService`, `TraceService`, and `WaveformService`.   So some of this is trying to go for least surprise even though they are not fully symmetric.  For example, the outputing services need to be setup after the build. `WaveformService` tracks with the simulator running, so it is truly a live 'service', but the output services execute  (though we could use an api that says 'generate' for those.  

I never thought of `SynthBuilder` as a user API.

Here is a live usage on HCL:
```dart
 final adder = FloatingPointAdderSinglePath(clk: clk, fp1, fp2);
      await adder.build();
      NetlistService(adder,
          outputPath: '$buildDir/${adder.definitionName}.json');
      final sv =
          SystemVerilogService(adder, outputPath: '$buildDir/${adder.definitionName}.sv'
              // embedSourceTraceComments: false,
              );
      final sc = SystemCService(adder,
          outputPath: '$buildDir/${adder.definitionName}.sc', multiFile: false);
      TraceService(adder, svService: sv, scService: sc).write('x.flc.json');
      WaveformService(adder,
          format: WaveOutputFormat.fst,
          outputPath: '$buildDir/${adder.definitionName}.fst');
```

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: This is an interesting point.  I should probably make the header optional and add it to whatever the output format is.
But:   if I do this now, I will likely break cross-probing which is downstream.

tool/gh_actions/run_tests.sh
desmonddak: I'm struggling with VSCode and how it handles packages.  The issue is that packages have their own pubspec.yaml.  There is confusion between what VSCode sees and what we see in the shell (on disk).   

I can revert and try over to see what is going on.

test/waveform_service_test.dart
desmonddak: Then this should be waveform_service_test.dart, right?

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: This is not a good routine.  I need to replace with a method that retrieves.  But with so much live code, I will need to first deprecate so things still run and then I can migrate.  Same with the other method that returns a Map.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: Let me check where this is called.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: Now unused. I removed.

benchmark/many_submodules_benchmark.dart
desmonddak: Done.

doc/tutorials/chapter_2/helper.dart
desmonddak: Removed registration

lib/src/module.dart
desmonddak: Removed registration

test/fsm_test.dart
desmonddak: Done.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: Changed all codegen style to use 'output'

test/waveform_service_test.dart
desmonddak: Renamed, left one test to drive `WaveDumper` despite deprecation.

lib/src/diagnostics/module_services.dart
desmonddak: Looks like it was set during module construction, so we have an `@internal` called `registerRootModule`.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: We added methods to call and deprecated these to allow me to merge and fix downstream by deprecation markers.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: Removing all calls that occur by definitionName, we must use the uniquified instance name.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: I fixed this in the downstream branch and pulled the header controls into this repo.
So we now have header output control orthogonal to multi-file control

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: One additional thing:  the synthesizers must run to store FLC info.  Then the `TraceService` can access to generate FLC.

lib/src/diagnostics/module_services.dart
desmonddak: `ModuleTree` is the way that devtools access the design hierarchy.  Some changes downstream will make this more clean.  `ModuleTree` has all the JSON communication using the rohd_hierarchy package.

`ModuleServices` is the holder of all the user-facing services we have.

desmonddak: Changes slightly to `ModuleTree.instance.rootModule = module`.  We can revisit once we have devtools integrated as this may be easier to refactor there if we need more clarity or modularity.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: I think of these services as constructing an in-memory view that can be further manipulated (written to file, written to string, perhaps searched, etc).  The only challenge I face is that if we change any line numbers then trace gets broken.

desmonddak: A added simpler `Module.dumpSystemVerilog` and `Module.dumpWaveforms` which don't require forming a Service.
I discovered some usefuless to the Service model:  it can stay alive and provide an API, so for example, the `NetlistService` can provide connectivity queries beyond just netlisting as it has created a model of the design. We could do that with Module after `build`, but there are other examples like the `WaveformService` enabling incremental waveform queries.  We can also make these more highly optioned so we don't keep extending all new capabilities as part of `Module`.

lib/src/diagnostics/module_service.dart
mkorbel1: if we don't fix before next release, will this not get us stuck with the API?

tool/gh_actions/run_tests.sh
```diff
@@ -11,8 +11,17 @@
 
 set -euo pipefail
 
+# Run main package tests (auto-discovers test/)
 dart test
 
-# run tests in JS (increase heap size also)
+# Run tests in sub-packages
+for pkg in packages/*/; do
```
mkorbel1: what about hierarchical packages (like rohd_devtools_widgets)?

lib/src/synthesizers/systemverilog/system_verilog_service.dart
mkorbel1: i dont see these in the current PR?  also, does `dumpWaves` work better?

lib/src/wave_dumper.dart
```diff
@@ -1,233 +1,61 @@
...
+/// ```
+///
+/// With:
+/// ```dart
+/// var service = WaveformService(module, outputPath: 'output.vcd');
+/// ```
+///
+/// For more control over filtering, timescale, and recording windows, see
+/// [WaveformService] constructor parameters.
+@Deprecated('Use WaveformService instead')
```
mkorbel1: or the `dumpWaves` on `Module`?

mkorbel1: Update documentation (`doc/` directory) as well.

benchmark/many_submodules_benchmark.dart
```diff
@@ -33,7 +33,7 @@ class ManySubmodulesBenchmark extends AsyncBenchmarkBase {
   Future<void> run() async {
     final dut = ManySubmodulesModule(Logic(), numSubModules: 10000);
     await dut.build();
-    dut.generateSynth();
+    SystemVerilogService(dut).output;
```
mkorbel1: should use `dumpSystemVerilog` instead in most cases, including tutorials, benchmarks, simple tests, etc?

benchmark/wave_dump_benchmark.dart
```diff
@@ -56,7 +56,7 @@ class WaveDumpBenchmark extends AsyncBenchmarkBase {
     _mod = _ModuleToDump(Logic(), _clk);
     await _mod.build();
 
-    WaveDumper(_mod, outputPath: _vcdTemporaryPath);
+    WaveformService(_mod, outputPath: _vcdTemporaryPath);
```
mkorbel1: should use `dumpWaves` in most scenarios including simple tests, benchmarks, tutorials, etc?

lib/src/diagnostics/module_services.dart
```diff
@@ -0,0 +1,79 @@
...
+  set rootModule(Module? module) {
+    ModuleTree.instance.rootModule = module;
+  }
+
+  /// Returns the module hierarchy as a JSON string.
+  ///
+  /// DevTools evaluates this via `EvalOnDartLibrary` to display the module
+  /// hierarchy.  Richer design views (e.g. a slim netlist) are composed by the
+  /// DevTools client from the relevant registered service.
+  String get hierarchyJSON => ModuleTree.instance.hierarchyJSON;
```
mkorbel1: nit: use `Json` instead (https://dart.dev/effective-dart/style#do-capitalize-acronyms-and-abbreviations-longer-than-two-letters-like-words)

lib/src/diagnostics/waveform_service.dart
```diff
@@ -0,0 +1,435 @@
...
+
+  /// Creates a [WaveformService] for [module].
+  ///
+  /// [module] must be built before construction.
+  ///
+  /// Use the optional constructor parameters to configure format, path,
+  /// filtering, timescale, start/stop times, flush size, and overwrite policy.
+  WaveformService(
+    this.module, {
+    this.outputPath = 'waves.vcd',
```
mkorbel1: should this default to `null` and automatically pick the name based on `format` instead?  e.g. if someone picks `format: .fst` but doesnt specify the path, it would be weird for it to be `waves.vcd`

maybe if `format` is default to `null` also then you can infer the type from the file name too?  e.g. if someone says `waves.fst` it defaults format to fst?

then just need to document behavior if there is none specified for either?

lib/src/diagnostics/waveform_service.dart
```diff
@@ -0,0 +1,435 @@
...
+/// The output format for waveform capture.
+enum WaveOutputFormat {
+  /// Value Change Dump — the classic text-based waveform format.
+  vcd,
+
+  /// Fast Signal Trace — a compact binary format.
+  ///
+  /// Requires an FST writer to be available; see the DevTools subclass for
+  /// a fully FST-backed implementation.
+  fst,
```
mkorbel1: is FST supported in this PR?

tool/gh_actions/install_dependencies.sh
```diff
@@ -12,3 +12,11 @@
 set -euo pipefail
 
 dart pub get
+
+# Install dependencies for sub-packages
+for pkg in packages/*/; do
```
mkorbel1: what about sub-sub-packages?

lib/src/diagnostics/module_services.dart
desmonddak: Done for JSON.

tool/gh_actions/install_dependencies.sh
desmonddak: The suggestion is to use a pub workspace.  Since there is a mix of flutter and dart packages, this means flutter controls.  But that gives us full repo analysis and testing without scripting, but with a dart program to manage.

lib/src/diagnostics/waveform_service.dart
desmonddak: I reworked these substantially to consider the idea of having byte stream as well as file output, and defaults like module name and use the format as the extension, etc.  So it would be good to reanalyze the changes to the API.  Conceptually similar, but more scalable.

benchmark/wave_dump_benchmark.dart
desmonddak: Done.  Renamed to `dumpWaves`.  Using the two dump routines everywhere except the new services tests.

benchmark/many_submodules_benchmark.dart
desmonddak: Done.

lib/src/wave_dumper.dart
desmonddak: Done.

desmonddak: Yes, the deprecation mentions first using `Module.dumpWaves()` and then the `WaveformService`.

tool/gh_actions/run_tests.sh
desmonddak: See comment below.  pub workspace is the Dart/Flutter answer.

tool/gh_actions/install_dependencies.sh
desmonddak: https://dart.dev/tools/pub/workspaces

benchmark/wave_dump_benchmark.dart
desmonddak: Done

benchmark/many_submodules_benchmark.dart
desmonddak: Done

lib/src/wave_dumper.dart
desmonddak: Done

tool/gh_actions/install_dependencies.sh
desmonddak: https://dart.dev/tools/pub/workspaces

lib/src/diagnostics/waveform_service.dart
desmonddak: No, but we check here and throw.  We should add fst soon.

lib/src/diagnostics/module_service.dart
desmonddak: This has been rearchitected to be consistent about output directories, filenames, and extensions.
Not sure we should move in the URI direction as it adds a level of complexity for user.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
desmonddak: This should be resolved:  we have dumpWaves and dumpSystemVerilog support for legacy.

tool/gh_actions/run_tests.sh
desmonddak: Please review PR #700

tool/gh_actions/install_dependencies.sh
desmonddak: Please review PR #700

lib/src/diagnostics/waveform_service.dart
```diff
@@ -0,0 +1,490 @@
...
+
+  /// Whether waveform bytes are written to [outputFilePath].
+  ///
+  /// When `false`, the service retains its waveform bytes in memory and exposes
+  /// them through [artifacts].
+  final bool writeToFile;
+
+  /// Whether to enable DevTools streaming.
+  ///
+  /// The base [WaveformService] stores this flag but takes no action on it.
```
mkorbel1: question: why does the base store the flag then?

lib/src/modules/conditionals/flop.dart
mkorbel1: this is back again

lib/src/module.dart
```diff
@@ -1147,25 +1197,19 @@ abstract class Module {
       throw ModuleNotBuiltException(this);
```
mkorbel1: nit: this check shouldn't be needed anymore since service will catch it?

doc/tutorials/chapter_8/02_finite_state_machine.md
```diff
@@ -272,7 +272,7 @@ Let also attach a `WaveDumper` to preview what is the waveform and what happened
 
 ```dart
 if (!noPrint) {
-  WaveDumper(oven, outputPath: 'oven.vcd');
+  oven.dumpWaves(outputPath: 'oven.vcd');
```
mkorbel1, line 275: good but we still mention `WaveDumper` in a couple of the tutorials' text, as well as in rohd.json extension snippets

rohd_devtools_extension/lib/rohd_devtools/services/tree_service.dart
```diff
@@ -19,10 +19,10 @@ import 'package:vm_service/vm_service.dart';
 class TreeService {
   /// Primary expression for hierarchy JSON — available in all ROHD versions
   /// that ship inspector_service.dart (i.e. main and later).
-  static const _primaryInvokeFunc = 'ModuleTree.instance.hierarchyJSON';
+  static const _primaryInvokeFunc = 'ModuleTree.instance.hierarchyJson';
 
   /// Fallback kept for any pre-inspector ROHD target.
-  static const _legacyInvokeFunc = 'ModuleTree.instance.hierarchyJSON';
+  static const _legacyInvokeFunc = 'ModuleTree.instance.hierarchyJson';
```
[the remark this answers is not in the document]
mkorbel1: @copilot is this valid since the ROHD release will ship with a matching devtools extension always?

lib/src/diagnostics/inspector_service.dart
```diff
@@ -86,19 +87,35 @@ class ModuleTree {
...
-        'reason': 'module not yet build',
-      });
+  String get hierarchyJson {
+    final rootModule = _rootModule;
+    return rootModule?.buildModuleTreeJsonSchema(rootModule) ??
+        json.encode({
+          'status': 'fail',
+          'reason': 'module not yet build',
+        });
+  }
```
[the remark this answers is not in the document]
mkorbel1: @copilot is this valid since the ROHD release will ship with a matching devtools extension always?

pubspec.yaml
```diff
@@ -7,7 +7,7 @@ issue_tracker: https://github.com/intel/rohd/issues
 documentation: https://intel.github.io/rohd-website/docs/sample-example/
 
 environment:
-  sdk: '>=3.0.0 <4.0.0'
+  sdk: ^3.6.0
```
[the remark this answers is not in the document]
mkorbel1: i think this is fine, people are likely past 3.6 by now mostly and `dart pub` commands will properly prevent issues and explain why upgrades do not work

lib/src/module.dart
desmonddak: Yes. Removed.

desmonddak:
On <flop>.constantResetValue():

For example, the netlist mapper reads it when converting a `FlipFlop` with a constant asynchronous reset into a Yosys `$adff` cell:

```dart
if (hasReset && !hasDynamicResetValue) {
  parameters['ARST_VALUE'] =
      flipFlop.constantResetValue!.toString(includeWidth: false);
}
```

The actual call is in `netlist_cell_mapper.dart`.

Given:

```dart
final q = flop(
  clk,
  d,
  reset: reset,
  resetValue: 9,
  asyncReset: true,
);
```

`constantResetValue` returns the width-adjusted `LogicValue` for `9`. The mapper uses it to emit the equivalent netlist parameter:

```text
ARST_VALUE = 9
```

For a reset value supplied through a `Logic` port instead, `constantResetValue` returns `null`, and the mapper uses the dynamic reset-value input rather than a constant parameter.

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
```diff
@@ -0,0 +1,114 @@
...
+import 'package:rohd/rohd.dart';
+import 'package:rohd/src/synthesizers/netlist/netlist_cell_mapper.dart';
+export '../utilities/synth_module_stop_policy.dart';
+
+/// Configuration for netlist synthesis.
+///
+/// The netlist synthesizer serves two main consumer flows, both configured
+/// through this configuration:
+///
+/// **Flow 1 — Slim JSON** (`NetlistService.slimJson`):
```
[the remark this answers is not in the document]
desmonddak: Yes, this is coming in a future PR which establishes the NetlistService.  We will fix this documentation for now.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
```diff
@@ -0,0 +1,209 @@
...
+  late final List<SynthFileContents> fileContents;
+
+  /// Creates a [SystemVerilogService] for [module].
+  ///
+  /// [module] must already be built.
+  ///
+  /// [outputDirectory] defaults to the current directory and [outputBaseName]
+  /// defaults to [Module.definitionName]. [includeHeader] defaults to
+  /// `!multiFile` to preserve the historical layout.
+  SystemVerilogService(
```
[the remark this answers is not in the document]
desmonddak: Fixed.  It can now register.

CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
+## Next Release
+
+- Migrated active packages to a pub workspace. The root `workspace:` list is
+  the single package roster, and `tool/workspace.dart` now provides
+  repository-wide analysis, native and Node.js test commands, and generation
+  of the multi-root VSCode workspace.
```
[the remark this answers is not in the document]
desmonddak: I removed as this belongs in a separate PR for pub workspace.

CHANGELOG.md
```diff
@@ -1,3 +1,27 @@
...
+  requiring filesystem output.
+- Added `SystemVerilogService` for configured SystemVerilog synthesis,
+  in-memory source output, artifact inspection, and explicit directory writes.
+  Added `WaveformService` for in-memory waveform capture with optional file
+  writing through `writeToFile`.
+- Added legacy-compatible `Module.dumpSystemVerilog` and `Module.dumpWaves`
+  convenience methods. `dumpSystemVerilog()` provides simple in-memory output
+  or writes an optional `outputPath`; `dumpWaves()` replaces `dumpWaveforms`
+  for standard VCD capture. `WaveDumper` and `generateSynth` are deprecated in
+  favor of these `Module` methods.
```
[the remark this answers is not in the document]
desmonddak: We removed this comment.  This was the former name of dumpWaves.

lib/src/diagnostics/inspector_service.dart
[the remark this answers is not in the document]
desmonddak: We removed this fallback as we will match the devtools expectation with the inspector service

lib/src/module.dart
```diff
@@ -1133,12 +1130,65 @@ abstract class Module {
...
+    String? outputPath,
+    bool multiFile = false,
+    SystemVerilogSynthesizerConfiguration configuration =
+        const SystemVerilogSynthesizerConfiguration(),
+  }) {
+    final service = SystemVerilogService(
+      this,
+      outputDirectory: outputPath ?? '.',
+      multiFile: multiFile,
+      configuration: configuration,
```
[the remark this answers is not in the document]
desmonddak: We aligned this with the WaveformService way of doing things with directory output, filename (defaulting to module name), and handling of extensions.

lib/src/diagnostics/waveform_service.dart
desmonddak: Removed.  It is coming in a later PR.

lib/src/module.dart
```diff
@@ -1133,32 +1130,85 @@ abstract class Module {
...
+  /// file. This legacy convenience method writes to [outputPath] when provided.
+  /// With [multiFile] `true`, [outputPath] must be a directory and each module
+  /// definition is written there. Otherwise, it is the path to a single
+  /// concatenated SystemVerilog file.
   ///
-  /// The [configuration] controls options specific to SystemVerilog output.
-  String generateSynth({
+  /// For additional output controls and access to synthesis results, use
+  /// [SystemVerilogService] directly.
+  SystemVerilogService dumpSystemVerilog({
```
mkorbel1: this should return a string not a service!

desmonddak: Resolved: it returns a `String`.

lib/src/synthesizers/systemverilog/system_verilog_service.dart
mkorbel1: The string-returning Module shortcut and the richer service API are much clearer now. I would like to finish aligning the related service contracts before this PR lands, while we can still change them without adding another compatibility layer.

`NetlistService` accepts `outputDirectory`/`outputBaseName`, but also a separate `outputPath` that triggers `write()` during construction. Its artifact name still comes from the basename, regardless of that exact path. In contrast, `SystemVerilogService` constructs the generated view in memory and writes through an explicit `writeOutputs()` operation. Callers currently need different rules for naming, write timing, and the write operation itself for two generated-output services.

Please align these generated-output services around one convention: directory/basename configuration should not itself trigger filesystem output; artifact names and default written filenames should agree; and the explicit write operation should use consistent naming and semantics. Please identify any exact-path entry point that must remain for real compatibility and keep that adaptation separate from the primary convention. Waveform capture necessarily runs over time, so this does not require it to have the same execution lifecycle as static code generation.

I'm okay deferring URI support. To keep that deferral manageable, let's keep artifact identity and byte access separate from the local filesystem destination, and avoid multiplying path overloads that would each need migration later. We do not need a URI or remote-storage framework in this PR. Please add focused netlist-service tests for the agreed construction, default artifact/write naming, and explicit-write behavior alongside the corresponding SV coverage.

lib/src/diagnostics/waveform_service.dart
```diff
@@ -0,0 +1,488 @@
...
+        : signal.value.toString(includeWidth: false);
+    final marker = _signalToMarkerMap[signal];
+    _writeToBuffer('$updateValue$marker\n');
+  }
+
+  // ─── Buffered I/O ─────────────────────────────────────────────
+
+  void _writeToBuffer(String contents) {
+    _fileBuffer.write(contents);
+    _inMemoryOutput.write(contents);
```
mkorbel1: Please make whole-history retention an explicit choice and preserve a file-backed capture path whose retained output memory does not grow with the entire trace. Every write currently appends to `_inMemoryOutput`, even when `writeToFile` is true; flushing only clears `_fileBuffer`. This also changes the behavior of `Module.dumpWaves` and deprecated `WaveDumper`, whose previous implementation did not keep the full VCD in memory.

I can see full history being useful for later debugger or waveform-viewer queries, but ordinary file dumping must not pay that unbounded retention cost by default. Long simulations of large designs can exhaust memory even though the caller is writing to disk. At minimum we need a way to disable full-history retention, and the normal file-dumping path should use that bounded-retention behavior by default. In-memory-only full-history capture should be an explicit, documented choice.

Please keep artifacts usable in the file-backed mode without reconstructing the entire trace in RAM. Reading the completed file as a byte stream is one possible approach; the exact mechanism can follow the intended artifact lifecycle. Please cover file-only capture, explicitly retained history, and artifact reads after simulation finalization in the tests. Returning `Stream.value` around a fully materialized string does not bound the retained memory.

lib/src/diagnostics/waveform_service.dart
```diff
@@ -0,0 +1,488 @@
...
+    }
+    if (stopTime != null && timestamp > stopTime!) {
+      return false;
+    }
+    return true;
+  }
+
+  void _captureTimestamp(int timestamp) {
+    if (!_isInRecordingWindow(timestamp)) {
+      _changedThisTimestamp.clear();
```
mkorbel1, line 461: Please snapshot the current signal values when recording begins. `_writeScope` emits the construction-time values, but `_captureTimestamp` discards pre-window changes without establishing a new initial state for the window. A signal that changes before `startTime` and then stays stable is therefore recorded with the wrong value.

For example, initialize an input to 0, change it to 1 at time 5, and use `startTime: 10`. With simulation actions at 15 and 20 but no further signal change, the resulting VCD still reports 0 at time 15. The same setup without `startTime` correctly reports 1.

Please add this stable-at-window-entry case as a regression and ensure the window has a correct initial snapshot even when no signal toggles at the boundary. Recording should not require an extra post-window transition to discover the current value.

lib/src/diagnostics/module_services.dart
```diff
@@ -0,0 +1,79 @@
...
+  /// Returns the registered service of type [T], or `null` if none.
+  T? lookup<T extends ModuleService>() => _services[T] as T?;
+
+  /// Removes the registered service of type [T], if any.
+  void unregister<T extends ModuleService>() {
+    _services.remove(T);
+  }
+
+  /// Resets all services.  Intended for test teardown.
+  void reset() {
```
mkorbel1, line 89: I'm comfortable with global discovery here, given ROHD's existing global `Simulator` model. The issue is that the two advertised ways to discover a service can disagree about whether it is still registered.

Constructing a `SystemVerilogService` populates both `ModuleServices.instance.lookup<SystemVerilogService>()` and `SystemVerilogService.current`. After `ModuleServices.instance.reset()`, lookup is null but `current` still returns the old service. An inspector using `current` therefore sees a service after the documented registry reset has removed it.

Please make these advertised discovery routes share a coherent source of truth and reset/unregister behavior across the services. A `current` accessor backed by registry lookup is one possible approach that would keep the generic registry independent of concrete service types. Please cover registration, replacement, opt-out, unregister, and reset without requiring tests to manually clear each service's static field.

Please also state the lifetime boundary: whether simulator reset affects discovery, or whether callers reset the registry separately, and whether discovered service types are independent latest registrations rather than an atomic set for `rootModule`. Separate lifetimes can be reasonable because generated results may remain useful after a simulation ends. This is not a request for a per-module registry or for simulator reset to automatically discard all synthesis results.

doc/user_guide/_docs/A21-generation.md
```diff
@@ -16,17 +17,32 @@ void main() async {
...
+    outputPath: 'build/systemverilog',
+    multiFile: true,
+  );
+  ```
+
+  For generated text without writing a file, use `dumpSystemVerilog` without an
+  `outputPath`:
+
+  ```dart
+  final generatedSv = myModule.dumpSystemVerilog().output;
```
mkorbel1: Follow-up to [the earlier string-return discussion](https://github.com/intel/rohd/pull/673#discussion_r3858488349).

Thanks for making `dumpSystemVerilog` return the String directly. The implementation now matches the requested API, but this guide still describes the previous version: it passes `outputPath` and `multiFile` to the shortcut and accesses `.output` on its result. Those examples cannot compile against the current signature, which accepts only `configuration` and returns a String. The port-configuration example also still passes `outputPath`.

Please finish updating these examples: use the shortcut directly for in-memory text, and `SystemVerilogService` plus its explicit write operation for filesystem or multi-file output. The later registration paragraph also says both Module shortcuts use normal service registration defaults, but the SV shortcut deliberately uses `register: false`; `dumpWaves` registers by default. Please make that distinction accurate too, preserving the agreed one-shot behavior rather than restoring the old parameters.

mkorbel1 (changes requested):
The simple Module shortcuts, retained services, and named artifacts are a useful split, and much of the previous feedback is now incorporated. Before this lands, please address the waveform-window correctness issue and the unconditional full-history memory retention, finish aligning the generated-output service contracts, make discovery reset coherent, and bring the guide in line with the final shortcut API.

Global type-keyed discovery seems reasonable alongside the existing Simulator model; I am not asking for a per-module registry. Deferring URI support is also reasonable provided this PR keeps artifact identity/byte access separate from filesystem destinations and establishes consistent local output conventions.

lib/src/synthesizers/netlist/netlist_service.dart
```diff
@@ -0,0 +1,306 @@
...
+    synthesizer = NetlistSynthesizer(configuration: configuration);
+    this.packageRoot = effectiveRoot;
+    synthBuilder = SynthBuilder(module, synthesizer);
+    _fullJson = synthesizer.synthesizeToJson(
+      module,
+      packageRoot: effectiveRoot,
+    );
+
+    final decoded = jsonDecode(_fullJson) as Map<String, dynamic>;
+    _modulesMap =
```
mkorbel1, line 93: Please preserve the ability to fetch full module details after selecting slim output. With `NetlistService(..., configuration: NetlistSynthesizerConfiguration(slimMode: true))`, the constructor decodes the already-reduced JSON into `_modulesMap`. Later, `moduleJson()` reads that same reduced map, so its cells still have no `connections`.

A one-NOT-gate example reproduces this: the default-mode service returns input/output connections, while the configured-slim service returns a `$not` cell without them. The default service followed by `slimJson` and then `moduleJson(...)` works; it is selecting `slimMode: true` at construction that loses the service's full-detail path.

The intended minimal-hierarchy-then-detailed-query workflow makes sense, and the initial payload should stay small. Please derive those different views from retained complete synthesis results, rather than making the transmitted slim projection the only source for later queries. The underlying synthesizer already retains full connectivity. Add a focused service regression that selects slim mode and then requests a module's complete connections.

lib/src/diagnostics/waveform_service.dart
```diff
@@ -0,0 +1,561 @@
...
+    for (final sig in snapshot) {
+      _writeSignalValueUpdate(sig);
+      onValueChange(sig, timestamp);
+    }
+    _changedThisTimestamp.clear();
+
+    onTimestampCapture(timestamp, snapshot);
+  }
+
+  void _writeWindowSnapshotIfNeeded(int timestamp) {
```
mkorbel1, line 478: The new window snapshot fixes the VCD case from [the earlier thread](https://github.com/intel/rohd/pull/673#discussion_r4018720841), but please deliver equivalent initialization to subclass consumers as well. `_writeWindowSnapshotIfNeeded` writes values directly into the VCD and bypasses the hooks.

A minimal subclass that seeds history in `onSignalCollected` and records updates in `onValueChange` still sees the construction-time 0 when the signal changes to 1 at time 5, recording starts at 10, and no further transition occurs. With idle simulation actions at 15 and 20, the VCD correctly reports 1 but the subclass history still contains only `(0, 0)`. The same subclass without a window receives the update and agrees with the VCD.

Please define and implement how the full window-entry state reaches subclass consumers, including the notification needed to finish its initial batch before subsequent updates. It need not describe initial values as physical transitions or require a particular new hook, but a consumer following the documented capture hooks needs a coherent starting state. Add a regression comparing hook-fed history with the VCD for this stable-at-window-entry case.

lib/src/synthesizers/netlist/netlist_service.dart
```diff
@@ -0,0 +1,306 @@
...
+        final modData = _modulesMap[definitionName];
+        if (modData == null) {
+          return jsonEncode(<String, String>{
+            'status': 'not_found',
+            'reason': 'module "$definitionName" not in netlist',
+          });
+        }
+        return jsonEncode(<String, Object?>{
+          'creator': 'ROHD netlist synthesizer',
+          'version': formatVersion,
```
mkorbel1: Please use one authoritative netlist format version across the service's full, per-module, and slim outputs. The same service currently reports `0.0.1` through `json` and `version`, but `0.0.5` through both `moduleJson()` and `slimJson`: the projections substitute `NetlistService.formatVersion` for the version emitted by the synthesizer.

Both inconsistencies reproduce on a one-gate module with default configuration. Preserve the actual schema version when projecting the synthesized data and add a consistency check across the views. If the service's transport envelope intentionally has a separate version, distinguish that explicitly from the underlying netlist format rather than substituting it under the same field.

lib/src/synthesizers/netlist/netlist_service.dart
```diff
@@ -0,0 +1,306 @@
...
+
+  /// Returns the set of module definition names in the netlist.
+  Set<String> get moduleNames => _modulesMap.keys.toSet();
+
+  /// Read-only access to the parsed modules map.
+  ///
+  /// Each key is a definition name and each value is the Yosys-style
+  /// module descriptor containing `ports`, `cells`, and `netnames`.
+  Map<String, dynamic> get synthesizedModules =>
+      Map<String, dynamic>.unmodifiable(_modulesMap);
```
mkorbel1, line 208: Please clarify the ownership contract of `synthesizedModules`. `Map.unmodifiable` protects the outer map, but the nested module/cell maps and bit lists still reference the service's `_modulesMap`. The current "read-only access" wording does not make that boundary clear.

If a caller modifies nested data, an uncached `moduleJson()` or `slimJson` can reflect the change while the original `json` and already-cached projections do not. Are callers expected to treat the entire returned graph as immutable, or is this intended to be an isolated/read-only snapshot?

Document the nested no-mutation requirement if shared data is intentional. If stronger isolation is the intended guarantee, implement that consistently instead. I am not asking for unconditional deep copies; the important point is to establish an ownership contract that keeps the different service views coherent.

mkorbel1 (changes requested):
The earlier changes address the file-backed retention regression, the VCD window initialization, the output-service conventions, discovery/reset, and the guide migration. The distinction between configurable in-memory service capture and ordinary file dumping is reasonable. The capability API also makes sense in preparation for the staged FST/debugger work; I am not asking to redesign or defer it.

The deeper pass found three remaining issues to address before merging: full module connectivity is unavailable after constructing a slim-configured netlist service; waveform window initialization does not reach subclass consumers; and full versus projected netlists advertise different format versions. The comments include small reproductions and passing controls. I have also included a parsed-view ownership clarification and grouped nonblocking cleanup/documentation notes.

### Nonblocking Cleanup and Documentation

A few smaller cleanup and API-documentation items to handle together; these are nonblocking relative to the correctness findings:

1. Reuse the netlist builder already created by the service. The constructor assigns `synthBuilder = SynthBuilder(module, synthesizer)`, then calls `synthesizeToJson()`, which constructs another builder. Prefer the existing combined-JSON operation on the retained builder, preserving any required downstream trace integration. This also fits keeping complete results available for the slim/full projections.
2. Correct the JSON shape documentation. `moduleJson()` describes a bare definition-name map but returns a creator/version/modules envelope. `slimJson` says it has the same structure as `toJson`, but `toJson` is a summary and slim output wraps module payloads under `netlist`. Show the actual shapes, and describe lookup keys in terms of the generated names returned by `moduleNames`.
3. Align `isCompatibleVersion` documentation with the chosen rule: the text says matching the major version is sufficient, while the implementation compares major and minor. No broader versioning framework is needed.
4. Describe the actual callback granularity. `onValueChange` says every value-change event, but a set coalesces changes and the hook receives the final signal value once per captured timestamp. `onTimestampCapture` says only timestamps with changes, but finalization can deliver an empty batch. The coalescing is reasonable; make the promises match the intended behavior and the window-initialization contract.
5. Clarify waveform artifact read timing: when it is complete, what a read during capture can observe, and that a file-backed `openRead()` opens the current file rather than an immutable snapshot or live tail. Also remove the claim that in-memory artifacts require `retainInMemory` to be *explicitly* enabled, since the accepted memory-only default enables it automatically.
6. Describe registry keys as the registration type argument rather than necessarily the runtime concrete type. A waveform subclass is registered by the base constructor under `WaveformService`; it is not automatically registered under its subclass type. The current global type-keyed model itself is fine.
7. Prefer the established `ModuleNotBuiltException` in NetlistService and WaveformService for the same unbuilt-module precondition, consistent with SystemVerilogService and SynthBuilder, unless there is a compatibility reason to keep the generic exception.
8. Given the FST/debugger work staged behind this PR, the shape of `canSendWaveforms()` is reasonable. Please just document the debugger operation that `true` promises, distinguishing that capability from transferring artifact bytes. For example, file-only VCD already has a readable artifact while this method returns false. This is a small contract clarification, not a request to rename/remove the API, add more flags, or bring the FST implementation into this PR.

lib/src/synthesizers/netlist/netlist_service.dart
mkorbel1: This still reproduces at `5db746fa4`. Reusing the retained builder removes the redundant synthesis, but `generateCombinedJson(synthBuilder, module)` still defaults to `configuration.slimMode`. With `slimMode: true`, it strips cell connections before the constructor populates `_modulesMap`, so `moduleJson()` still has no full connectivity to return.

The same one-NOT-gate probe fails with `connections == null` when configured slim; the default-mode control passes. Please keep this issue open and derive full module responses from complete retained synthesis data, while preserving the small initial slim payload. Please also add the requested regression that constructs a slim-configured service and then fetches full module connections.

lib/src/module.dart
```diff
@@ -1133,32 +1130,63 @@ abstract class Module {
...
+        register: false,
+      ).output;
+
+  /// Attaches waveform dumping for this [Module] to a VCD at [outputPath].
+  ///
+  /// Set [retainInMemory] to `true` when a debugger needs to query the
+  /// complete waveform history while the simulation is running.
+  WaveformService dumpWaves({
+    String outputPath = 'waves.vcd',
+    bool retainInMemory = false,
```
mkorbel1, line 1156: Would it make more sense to keep `retainInMemory` on `WaveformService`, rather than also exposing it on `dumpWaves()`? I think the explicit service better communicates an ongoing capture for displaying/querying waveforms and retaining history, while `dumpWaves()` reads as the quick dump-to-file API.

My preference is to keep that shortcut focused on bounded-memory file output and use `WaveformService` directly for debugger capture, including when both a file and retained history are needed. That workflow is already expressible with `writeToFile: true, retainInMemory: true` on the service.

Is there a common debugger workflow that benefits enough from keeping the shortcut to justify this extra parameter? I can see the convenience of adding one flag to an existing dump call, but I'd like to understand why it belongs on the simple dumping API rather than the explicit service. This isn't an objection to the service's retention defaults or the staged FST capability.

mkorbel1 (changes requested):
The waveform window initialization, projected netlist versions, and parsed-view ownership clarification are addressed, along with most of the grouped cleanup. The focused checked-in tests pass.

One correctness issue remains: a service constructed with `slimMode: true` still cannot return full module connectivity. I've followed up on the existing thread with the current reproduction. That is the reason to keep changes requested.

I've also asked whether the new `retainInMemory` option belongs on `dumpWaves()`, or whether retained/debugger capture should remain an explicit `WaveformService` choice. That's an API-boundary discussion, separate from the remaining defect. The two documentation notes are nonblocking.

### Nonblocking Documentation

Two small documentation follow-ups:

1. Please clarify that the window-entry snapshot is a separate callback batch and a value-change batch can follow at the same timestamp. For a signal entering time 10 at 1 and changing to 0 during time 10, the hooks now deliver `(10, 1)` followed by `(10, 0)`. That makes sense for entry initialization plus captured changes, but the opening `onValueChange` wording still promises each signal only once with its final value at a timestamp. A small start-boundary regression would make the intended two-batch behavior explicit. The original missing-initialization issue is fixed.
2. The JSON views now agree on their version, which fixes D3. Please also clarify the role of `NetlistService.formatVersion`: its documentation still calls it the version produced by the service, but it remains `0.0.5` while the outputs carry the synthesizer's `0.0.1`. This is a documentation/constant-role clarification, not another observed payload mismatch or compatibility failure.

PR #674, opened 2026-07-01
Collapse adjacent bit selects in swizzles to ranges

mkorbel1:
## Description & Motivation

This improves generated SystemVerilog for swizzles by compacting descending contiguous single-bit selections into legal slice expressions.

Examples:
- `{a[7], a[6], a[5]}` can now emit as `a[7:5]`
- `{arr[0][1], arr[0][0]}` can now emit as `arr[0][1:0]`

The collapse is intentionally conservative. It only applies to descending contiguous bit selects from the same packed source, and leaves unsupported cases expanded, including ascending selections, adjacent pre-existing range operands, expression-based selects, nested swizzle boundaries, and unpacked array sources.

## Related Issue(s)

N/A

## Testing

Added new tests, updated some other tests

## Backwards-compatibility

No

## Documentation

No

### DIFF
--- a/lib/src/modules/bus.dart
+++ b/lib/src/modules/bus.dart
@@ -1,4 +1,4 @@
-// Copyright (C) 2021-2025 Intel Corporation
+// Copyright (C) 2021-2026 Intel Corporation
 // SPDX-License-Identifier: BSD-3-Clause
 //
 // bus.dart
@@ -183,6 +183,11 @@ class BusSubset extends Module with InlineSystemVerilog {
 class Swizzle extends Module with InlineSystemVerilog {
   final String _out = Naming.unpreferredName('swizzled');
 
+  /// A regular expression that will have matches if an expression is a single
+  /// bit select of a signal or packed array element.
+  static final RegExp _singleBitSelectRegex =
+      RegExp(r'^\(?([A-Za-z_][A-Za-z0-9_$]*(?:\[\d+\])*)\[(\d+)\]\)?$');
+
   /// The output port containing concatenated signals.
   late final Logic out;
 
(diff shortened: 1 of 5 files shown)


lib/src/modules/bus.dart
```diff
@@ -315,15 +319,108 @@ class Swizzle extends Module with InlineSystemVerilog {
...
   }
+
+  /// Rewrites runs of adjacent descending single-bit selects from the same
+  /// packed signal into wider SystemVerilog slices.
+  ///
+  /// For example, `a[7], a[6], a[5]` becomes `a[7:5]`, and
+  /// `a[0][1], a[0][0]` becomes `a[0][1:0]`. Ascending runs are intentionally
+  /// left expanded because SystemVerilog slices cannot reverse bit order with
+  /// `lower:upper` syntax.
+  List<({String expression, int width})> _collapseContiguousBitSelects(
```
desmonddak, line 343: Could this have been done on SynthNets instead of strings?
This looks like a final pass for SV, but the concept make sense upstream

lib/src/modules/bus.dart
```diff
@@ -315,15 +319,108 @@ class Swizzle extends Module with InlineSystemVerilog {
...
+
+        lowerIndex = nextSelectedBit.index;
+        endIndex++;
+      }
+
+      if (endIndex == index + 1) {
+        operands.add((expression: expression, width: input.width));
+      } else {
+        operands.add((
+          expression: '${selectedBit.source}[${selectedBit.index}:$lowerIndex]',
```
desmonddak, line 380: IT would be kinda fun to add the SV shorthands for when you are selecting up to the end of the array, or reverse-indexing.

desmonddak:
I think this PR makes sense right now to get SV much cleaner.  I could see pushing some of the optimizations into the base and leaving here some SV-specific conventions like [:7].  Otherwise, the collapse could be structural, and then the strings would come out from there unless we are avoiding structural change in the synth layer for this case.

lib/src/modules/bus.dart
mkorbel1: are those safe with all tools?  generally have been trying to keep things minimal fanciness to maximize tool compatibility

lib/src/modules/bus.dart
mkorbel1: I agree and considered both options.  It was kinda nice how localized this change was, but since you bring it up let me reevaluate again.

mkorbel1: I think it would really cause a lot of churn in the synth stack to yank this up to a generic non-sv-specific pass, but then would still require additional sv-specific considerations.  Replacing modules (swizzles and bussubsets) can introduce a lot of complexity here, maybe a lot of corner cases and risks.  I think when we get to another generator that needs to apply similar transforms we can more concretely identify what makes sense to pull up to a common place and reuse.

lib/src/modules/bus.dart
mkorbel1: i think lets keep things simpler for now, resolving

mkorbel1:
I agree, makes SV a lot cleaner and is pretty elegant and limited, we can look for additional opportunities for reuse when the situation actually presents itself

PR #675, opened 2026-07-02
Netlist 

desmonddak:
## Description & Motivation

This is a netlist synthesizer that produces a netlist for the generated design in an extension of the Yosys output netlist format.
It provides routines for emitting just the hierarchy and ports ("slim" mode) as well as fully expanded and has hooks for even more incremental expansion modes.

## Related Issue(s)

None.

## Testing

There is a suite of tests that compare the netlist and its names against the SystemVerilog output. This netlist depended on the last central_naming branch to assure that signals in both formats had identical names.

## Backwards-compatibility

No.

## Documentation

This is a minor API addition (`Module.generateNetlist()`) but we will add more documentation and examples of the format, etc.
It will have some options as well, such as multiFile, which should parallel the `generateSynth()` API.


lib/src/examples/filter_bank_modules.dart
mkorbel1, line 1: i think it makes more sense to keep example code in the example directory at the root of the repo.  if necessary you can make sub-folders in there.

lib/src/examples/filter_bank_modules.dart
```diff
@@ -0,0 +1,937 @@
...
+    // enabled; otherwise it outputs high-Z.  Joining out↔bus makes the
+    // two nets share the same wire.
+    TriStateBuffer(storedValue, enable: writeEnable, name: 'busDriver')
+        .out
+        .gets(bus);
+  }
+}
+
+/// The top-level polyphase FIR filter bank.
+class FilterBank extends Module {
```
mkorbel1: this is a pretty substantial example, split into multiple files (one per class mostly)?

lib/src/module.dart
mkorbel1, line 1: formatting issue appeared in here, can't tell if there's any real change in here?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,2104 @@
...
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd/src/synthesizers/utilities/utilities.dart';
+import 'package:rohd/src/utilities/sanitizer.dart';
+
+///
+/// Skips SystemVerilog-specific processing (chain collapsing, net connects,
+/// inOut inline replacement) since netlist represents all sub-modules as
+/// cells rather than inline assignment expressions.
+class _NetlistSynthModuleDefinition extends SynthModuleDefinition {
```
mkorbel1: let's split classes into different files for this synth stuff, this is a huge file.  consider marking things `@internal` if you want to make sure they're not used outside of ROHD

lib/src/synthesizers/netlist/netlist_synthesizer.dart
mkorbel1, line 1: general question: we call this "netlist", but some might interpret that as being like a verilog gate level netlist?  or do you think it's an appropriate name?  this generates some specific JSON representation, right?

lib/src/synthesizers/netlist/leaf_cell_mapper.dart
```diff
@@ -0,0 +1,478 @@
...
+      final outName = tsb.inOuts.keys.first; // inout output
+      final r = ctx.remap({inName: 'A', enName: 'EN', outName: 'Y'});
+      return (
+        cellType: r'$tribuf',
+        portDirs: r.portDirs,
+        connections: r.connections,
+        parameters: <String, Object?>{'WIDTH': ctx.width(inName)},
+      );
+    });
+
```
mkorbel1: question: what do you do about custom SystemVerilog modules that are not Combinational/Sequential?  I know they aren't present really in ROHD, but people may implement their own.

lib/src/synthesizers/netlist/netlist_options.dart
```diff
@@ -0,0 +1,100 @@
...
+  final bool slimMode;
+
+  /// When `true`, contiguous ascending runs of ≥3 integer bit IDs in
+  /// `bits` arrays and cell `connections` arrays are replaced with
+  /// `"start:end"` range strings (e.g. `[52, 53, 54, 55]` → `["52:55"]`).
+  ///
+  /// This is backward-compatible: Yosys-format arrays already mix
+  /// integers with constant strings `"0"` and `"1"`.  Parsers can
+  /// detect range strings by the presence of `:`.
+  final bool compressBitRanges;
```
mkorbel1, line 96: is this similar to what's going to be available in the new range/range collapsing PR?  i guess that should merge first before this one, or do you want to test that it meets your needs first?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,2104 @@
...
+
+  /// Creates a [NetlistSynthesizer].
+  ///
+  /// All synthesis parameters are bundled in [options]; see
+  /// [NetlistOptions] for documentation on each field.
+  NetlistSynthesizer({this.options = const NetlistOptions()});
+
+  @override
+  bool generatesDefinition(Module module) =>
+      // Only modules with sub-modules generate their own module definition.
```
mkorbel1: this can be a little dangerous of an assumption -- sometimes a module will be custom systemverilog but also contain submodules for the purpose of simulation behavior.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,2104 @@
...
+      module is! FlipFlop && module.subModules.isNotEmpty;
+
+  @override
+  SynthesisResult synthesize(
+    Module module,
+    String Function(Module module) getInstanceTypeOfModule, {
+    SynthesisResult? Function(Module module)? lookupExistingResult,
+    Map<Module, SynthesisResult>? existingResults,
+  }) {
+    final isTop = module.parent == null;
```
mkorbel1: why do you care if it's the top?  what if someone wants to synthesize starting at a sub-module of the overall hierarchy? i think for the SV synth stack it's legal.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,2104 @@
...
+  @override
+  bool generatesDefinition(Module module) =>
+      // Only modules with sub-modules generate their own module definition.
+      // Leaf modules (no children) become cells inside their parent.
+      // FlipFlop has internal Sequential sub-modules but should be emitted as
+      // a flat Yosys $dff primitive, not as a hierarchical module.
+      module is! FlipFlop && module.subModules.isNotEmpty;
+
+  @override
+  SynthesisResult synthesize(
```
mkorbel1, line 75: wow what a function! can we break it down a little?  also, conceptually, I noticed that this doesn't seem to follow the same pattern really as the SV synth stack where each module only is responsible for what's required to assemble itself with sub-modules, this seems like one mega pass across all the design all at once?  do you pay a perf penalty anywhere for this approach for large designs?  if its effectively the same approach, is there a reason you chose to go this way?

lib/src/synthesizers/netlist/netlist_utils.dart
```diff
@@ -0,0 +1,460 @@
...
+    return null;
+  }
+
+  /// Safely retrieve the name from a [SynthLogic], returning null if
+  /// retrieval fails (e.g. name not yet picked, or the SynthLogic has
+  /// been replaced).
+  static String? tryGetSynthLogicName(SynthLogic sl) {
+    try {
+      return sl.name;
+      // ignore: avoid_catches_without_on_clauses
```
mkorbel1: why ignore? what are you trying to catch here?

lib/src/synthesizers/netlist/netlist.dart
```diff
@@ -0,0 +1,15 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// netlist.dart
+// Barrel file for netlist synthesis library.
+//
+// 2026 February 11
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+export 'leaf_cell_mapper.dart';
```
mkorbel1: are you sure you want to export all of this as part of the ROHD package or only some?

test/netlist_test.dart
mkorbel1, line 1: idea/question: with more synth stacks coming including this one, do we want to make it so that almost every test that builds some hardware runs through all the synth stacks?  right now a common pattern is functional tests, then iverilog tests, both with the same vectors.  maybe we should make the pattern more generic or like also pass modules through all the synth stacks?

lib/src/synthesizers/netlist/netlist_options.dart
desmonddak: I think so.  There are a range of optimizations I added to make netlists 'pretty' and now we have some of those ideas being replicated in the Verilog synthesizer and the base synthesizer.  See netlist_passes.dart.  Hopefully, we can eliminate most of those passes.

These options are controllable by bools.  So I would like to get them into the source and then make sure that they are covered by the newer general optimizations before removing, as I can test with on / off switches, and I believe I have tests for each optimization.

With the tests and a viewer, it should be easy to confirm or else merge these optimizations into the base so all output formats see the same structure.

There are some Verilog optimizations that I think should also be moved to the base.

This is why central_naming was key: we have a way to predict optimizations and reorder naming now to bias good names toward 'keeper' wires.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: It is a netlist in its truest form, utilizing JSON as its format, and capturing a hierarchical hypergraph with attributes on cells and nets. It has no more and no less description than a netlist would:  cells, ports, instances, wires.  It is actually quite close to EDIF in structure.

A Verilog gate-level netlist uses the Verilog language as its netlist format.   But Verilog is being abused as a netlist format, handy because everyone has a parser for it.  Some people call the SPICE format a netlist, but it has model 'cards' which are attributes on the transistors, and it starts to get dicey: it's a netlist with models.

The format is Yosys JSON, which is loaded by a tool called NetlistSVG as well as d3-hwschematic.
In Yosys examples, they literally say ```write_json my_netlist.json```.

So I think the most accurate term is netlist.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: Let me think about this:  I need a concept of stopping level for netlist.  Primitives are part of that.  But there is a challenge with `Flop` as it has `Sequential` inside and clearly I can stop at `Flop`.

test/netlist_test.dart
desmonddak: Yes.  The new filter_bank example has almost every ROHD construct in order to do just that.  In upcoming PRs there are miter tests between synthesizers to ensure name matching across all of them.

This branch is `netlist_pre`.  The final `netlist` branch has these tests.  I had to split somewhere because there is a `Services` architecture coming and the tests are all written using the new API.

lib/src/examples/filter_bank_modules.dart
desmonddak: The dilemma is this:  ROHD proper doesn't have great examples to use in tests.  So I started putting the actual body of example modules in `lib/src/examples` and started pointing to them from `examples`.  You cannot use the `examples` directory in tests (or at least I struggled to do it cleanly).  The filter_bank example was carefully constructed to use almost all ROHD constructs, especially `LogicArray` and `LogicStructure` as well as an `ioput`.

lib/src/synthesizers/netlist/leaf_cell_mapper.dart
desmonddak: I'll take the AR to generate an example for this to test.  It could be as simple as generating SV for a leaf block (like a HalfAdder) and then treating it as custom and we should have this example.
Then I can fix this netlist blindspot.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: I migrated the concept of stopping level into the base synth classes through which netlist can program additional stopping points (like FlipFlop).

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: I fixed this -- this had to do with header generation as the JSON is single file.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: Completely refactored, pushing some new capabilities into the synth layer, specifically dealing with components that are automatically elided by SystemVerilog, but exist in netlist (concat, swizzle, field access).  Namer modified to also deal with consistent naming of these invisible-to-Verilog components.

lib/src/synthesizers/netlist/netlist.dart
desmonddak: This is changed because I rearchitected the way we specify stopping level or leaf with a generic capability in synth and then programmed lists in the netlist.

lib/src/module.dart
desmonddak: This is that 'bug' in dart format that introduces tail-commas. I removed.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: Done.  A pretty substantial refactor complete.

lib/src/synthesizers/netlist/netlist_utils.dart
desmonddak: I resolved this without using a catch.

lib/src/synthesizers/netlist/leaf_cell_mapper.dart
desmonddak: Added the test: 'filter bank can stop traversal at an opaque custom SV module'

lib/src/examples/filter_bank_modules.dart
mkorbel1: we do have tests that run examples (see example_test.dart) that just need to be relative paths.  but i see the point.  given the capability do you still think its better to keep it in lib/src?  i dont know best practices across dart ecosystem

lib/src/module.dart
```diff
@@ -1155,6 +1155,17 @@ abstract class Module {
             .getSynthFileContents()
             .join('\n\n////////////////////\n\n');
   }
+
+  /// Returns a synthesized netlist JSON representation of this [Module].
+  String generateNetlist(
```
mkorbel1, line 1168: i've been sort of regretting having `generateSynth` at all, maybe long term can deprecate it, but it keeps being convenient.  this expansion of the pattern is maybe not great in that regard.  what if we instead upgraded the `generateSynth` to accept a `Synthesizer` and default to `SystemVerilogSynthesizer` to maintain backwards compatibility?  thoughts?

lib/src/examples/filter_bank_modules.dart
desmonddak: It looks like best practice is to either put them in ./examples or in test/fixtures.  So I will move them back and use relative paths.

lib/src/synthesizers/netlist/netlist_module_translation.dart
```diff
@@ -0,0 +1,547 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+
```
mkorbel1: nit: blank line?  i noticed a handful of files have this.

lib/src/synthesizers/netlist/netlist_module_translation.dart
```diff
@@ -0,0 +1,547 @@
...
+  /// The synthesis definition for this module level, when one can be built.
+  final NetlistSynthModuleDefinition? synthDef;
+
+  final NetlistCellMapper _netlistCellMapper;
+  final bool Function(Module module) _generatesDefinition;
+  final String Function(Module module) _getInstanceTypeOfModule;
+
+  /// The next available integer wire identifier.
+  int nextId = 2;
```
mkorbel1, line 36: i remember seeing a note why this was 2, but it's gone now? -- we should keep it :)

lib/src/synthesizers/netlist/netlist_synth_module_definition.dart
```diff
@@ -0,0 +1,138 @@
...
+      }
+      final synthLogic = logicToSynthMap[signal];
+      if (synthLogic == null) {
+        return false;
+      }
+      return !portArraySynthLogics.contains(synthLogic.resolved);
+    }).forEach(_concatAssembleArray);
+  }
+
+  void _subsetReceiveArrayPort(LogicArray port) {
```
mkorbel1, line 84: nit: suggest adding doc comments even on private helpers where possible to make understanding easier for humans/AI in the future

lib/src/synthesizers/utilities/utilities.dart
```diff
@@ -1,7 +1,12 @@
 // Copyright (C) 2024-2025 Intel Corporation
```
mkorbel1, line 1: copyright year (and check in general)

lib/src/utilities/namer.dart
```diff
@@ -24,6 +24,32 @@ import 'package:rohd/src/utilities/uniquifier.dart';
...
+  /// Canonical base name for synthesis-created array concat operations.
+  static const String synthArrayConcatOperationName = 'array_concat';
+
+  /// Canonical base name for synthesis-created structure slice operations.
+  static const String synthStructureSliceOperationName = 'struct_slice';
+
+  /// Canonical base name for synthesis-created structure concat operations.
+  static const String synthStructureConcatOperationName = 'struct_concat';
+
+  /// Returns the canonical base instance name for a synthesis-created
```
mkorbel1: I don't understand what these changes in Namer are for.  Is this mimicking the work that `Uniquifier` could already do?  Why are there these special constant names?

lib/src/module.dart
desmonddak: Please look at the module_services branch.  I move away from generateSynth -- not deprecate (yet), but I use a 'Service' that has its own options and output instead of in the Module api

desmonddak: This was ONLY done to make this branch reviewable at this point in time (this 'era' of having generateSynth).

lib/src/synthesizers/netlist/netlist_module_translation.dart
desmonddak: Let me go look -- I believe that there is a problem in the netlist parser with 0 and 1 being interpreted as binary values but I need to see if I resolved that.

lib/src/utilities/namer.dart
desmonddak: The dilemma I have is that the netlister needs to create new elements to make things more easily visualized.  Things that disappear when you think about them from an SV perspective.  The problem is that means some leaking from netlister down into `Namer` to ask it for these names.  No other synthesizer will see these, I don't believe unless these become constructions in the synth layer (that SV ignores or elides):

```dart
  /// Canonical base name for synthesis-created array slice operations.
  ///
  /// Netlist synthesis uses these when a [LogicArray] port or submodule output
  /// must be decomposed into element wires, such as a child reading
  /// `values.elements[0]` from an aggregate array input port.
  static const String synthArraySliceOperationName = 'array_slice';

  /// Canonical base name for synthesis-created array concat operations.
  ///
  /// Netlist synthesis uses these when independently driven [LogicArray]
  /// elements must be reassembled for an aggregate consumer, such as
  /// `values.elements[0] <= a` and `values.elements[1] <= b` feeding a child
  /// array input port.
  static const String synthArrayConcatOperationName = 'array_concat';

  /// Canonical base name for synthesis-created structure slice operations.
  ///
  /// Netlist synthesis uses these internally when a [LogicStructure] aggregate
  /// must expose field wires. The emitted netlist projection can replace them
  /// with `$struct_unpack` cells so field names are preserved.
  static const String synthStructureSliceOperationName = 'struct_slice';

  /// Canonical base name for synthesis-created structure concat operations.
  ///
  /// Netlist synthesis uses these when independently driven [LogicStructure]
  /// fields must be packed back into an aggregate struct output or submodule
  /// input. The emitted netlist projection represents this as `$struct_pack`.
  static const String synthStructureConcatOperationName = 'struct_concat';
  ```

Comment (desmonddak) [reply lib/src/utilities/namer.dart]: I could move these back up into netlist as they currently are synthesized there.

File: lib/src/module.dart (hunk 10 above)
Comment (mkorbel1) [reply lib/src/module.dart:1168]: Should we omit it then from this PR if it's going to be removed/deprecated soon anyways?

File: lib/src/utilities/namer.dart (hunk 15 above)
Comment (mkorbel1) [reply lib/src/utilities/namer.dart]: Yeah, I think it would be cleaner to just let those names be selected (late if necessary) along with all the other things needing selection, and be uniquified as needed.  I had to do something similar with `net_connect`

File: lib/src/module.dart (hunk 10 above)
Comment (desmonddak) [reply lib/src/module.dart:1168]: Tests need it.  We will deprecate in module_services_api.

File: lib/src/synthesizers/utilities/utilities.dart (hunk 14 above)
Comment (desmonddak) [reply lib/src/synthesizers/utilities/utilities.dart:1]: I scanned for these.  I think this file was indeed changed in 2025 as I started netlist then.  But I didn't check, might as well make it 2026 to match the significant amount of changes made.

File: lib/src/synthesizers/netlist/netlist_synth_module_definition.dart (hunk 13 above)
Comment (desmonddak) [reply lib/src/synthesizers/netlist/netlist_synth_module_definition.dart:84]: I added these.

File: lib/src/synthesizers/netlist/netlist_module_translation.dart (hunk 12 above)
Comment (desmonddak) [reply lib/src/synthesizers/netlist/netlist_module_translation.dart:36]: Indeed there is a problem with d3-hwschematic interpreting 0 as Logic 0 and 1 as Logic 1.

File: lib/src/synthesizers/netlist/netlist_module_translation.dart (hunk 11 above)
Comment (desmonddak) [reply lib/src/synthesizers/netlist/netlist_module_translation.dart]: I scanned for this particular problem along with fixing Copyright dates.

File: lib/src/utilities/namer.dart (hunk 15 above)
Comment (desmonddak) [reply lib/src/utilities/namer.dart]: I moved these down into the synth utilities for now.

File: lib/src/module.dart (hunk 10 above)
Comment (mkorbel1) [reply lib/src/module.dart:1168]: im concerned that once we merge this if we do a release it will become an API that needs deprecation

File: lib/src/utilities/namer.dart (hunk 15 above)
Comment (mkorbel1) [reply lib/src/utilities/namer.dart]: i still dont understand why these are relevant constants to define in a central place.  what is this SynthOperationNamer for?

File: lib/src/synthesizers/netlist/netlist_passes.dart [hunk 16]
```diff
@@ -0,0 +1,786 @@
...
+
+        cellsToRemove.forEach(cells.remove);
+        changed = true;
+      }
+    }
+  }
+
+  /// Replaces a `$concat` of adjacent `$slice` outputs from the same source
+  /// with one wider `$slice`.
+  static void collapseConcatOfAdjacentSlices(
```
mkorbel1, line 490: question: is this already all merged with main and benefits from the common collapsing in the synth stack?

lib/src/synthesizers/netlist/netlist_utils.dart
```diff
@@ -0,0 +1,446 @@
...
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd/src/synthesizers/utilities/utilities.dart';
+
+/// Shared utility functions for netlist synthesis and post-processing passes.
+///
+/// All methods are static — no instances are created.
+@internal
+class NetlistUtils {
```
mkorbel1: only static functions, why not abstract and no constructor?

lib/src/synthesizers/netlist/netlist_validation.dart
```diff
@@ -0,0 +1,218 @@
...
+    final multipleDriverWarnings = <String>[];
+
+    void report(String message, {bool multipleDriver = false}) {
+      warnings.add(message);
+      if (multipleDriver) {
+        multipleDriverWarnings.add(message);
+      }
+      if (printWarnings) {
+        // ignore: avoid_print
+        print(message);
```
mkorbel1, line 74: where does this print go? like in what case does this get turned on and why?

lib/src/synthesizers/utilities/synth_array_concat.dart
```diff
@@ -0,0 +1,39 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+
```
mkorbel1: blank lines remain :(

lib/src/synthesizers/utilities/synth_logic.dart
```diff
@@ -125,6 +131,15 @@ class SynthLogic {
   /// may be additional conditions that prevent clearing.
   bool get isClearable => mergeable;
 
+  /// Like [isClearable], but additionally permits a *renameable* signal (whose
+  /// chosen name carries no semantic meaning and would simply be dropped).
+  ///
+  /// This is `false` only for [isReserved] (e.g. ports) and constant signals,
+  /// whose identities must be preserved.  It is used by the aggregate-collapse
+  /// optimizations to eliminate intermediate (possibly named) nets.
+  bool get isClearableOrRenameable =>
```
mkorbel1: has this merged main? this existed in main i thought

lib/src/synthesizers/netlist/netlist_passes.dart
desmonddak: I believe netlist passes contains operations that are supersets of the common collapsing that was just added to the synth stack.


- Adjacent $slice outputs feeding a $concat can collapse into one wider $slice; example module: [netlist_test.dart:79]
- Transparent $slice/$buf alias chains can collapse by tracing bits back to the external source; example module: [netlist_test.dart:93
- The netlist pass file also has cleanup passes for trivial concat aliases and unconsumed transparent cells in [netlist_passes.dart:322

Similarly, there is code for structure field extraction and concatenation that also need collapse.

I've factored out the common collapsing code to leverage what is in main, a first layer of optimization.

I think what we can do is generate examples that trigger these passes and see if we want to push these optimizations down into the synth layer.  I have a feeling they would subsume the current collapsing code since they are more general.

lib/src/utilities/namer.dart
desmonddak: Netlist optimizations can create NEW slices and concats.  These need to be named.  There are four flavors:  array and struct times slice and concat.

lib/src/module.dart
desmonddak: This PR is designed to come before module_services_api which we are still not completely sure of.
meanwhile, netlist needs tests.
So we could simply mark this as testing only for now.

desmonddak: This is why this PR is from branch `netlist_pre`.  The `_pre` means before module_services.  I tried to peel apart the changes into digestible PRs, but this means things like this need to happen: APIs need to be there for a period and then disappear.  I cannot design a perfect merge order, especially when we can't quite agree on the services api.

lib/src/utilities/namer.dart
desmonddak: These are simply the base names of these special new slices and concats (which are then numbered for uniqueness).

lib/src/synthesizers/utilities/synth_logic.dart
desmonddak: This is unused and this had the potential to regress functionality in main, so I removed this getter.

lib/src/synthesizers/netlist/netlist_validation.dart
desmonddak: This is a validation check while we are debugging netlists.  The issue is some blocks have unconnected outputs and it is expected versus not.  So we cannot throw.  I now store this in a warnings class.  Cleaning up netlist connectivity for visibility can result in disconnection, so until we have completed migration of common versus netlist specific optimization, it would be good to have the validation routines.

lib/src/synthesizers/netlist/netlist_utils.dart
desmonddak: Yes, I converted.

lib/src/utilities/namer.dart
desmonddak: Another possibliity is that when we join concat operations, say, we name them by concatenating the names.   But most of the time it would be concat_1 merging with concat_2, so concat_1_2 would not really be that meaningful. So these become array_concat (and the next one is array_concat_0.

lib/src/module.dart
desmonddak: I marked it `@visibleForTesting`.

mkorbel1: or we can make it an extension only visible in that test that uses it?

lib/src/synthesizers/netlist/netlist_options.dart
```diff
@@ -0,0 +1,118 @@
...
+/// for incremental synthesis.
+///
+/// Example usage:
+/// ```dart
+/// const options = NetlistOptions(
+///   collapseTransparentClusters: true,
+/// );
+/// final synth = NetlistSynthesizer(options: options);
+/// ```
+class NetlistOptions {
```
mkorbel1: the systemverilog one uses the word "configuration" instead of "options" (I know I think you came first, sorry, but mine merged first unfortunately), should we align?

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
```diff
@@ -0,0 +1,100 @@
...
+  final bool slimMode;
+
+  /// When `true`, contiguous ascending runs of ≥3 integer bit IDs in
+  /// `bits` arrays and cell `connections` arrays are replaced with
+  /// `"start:end"` range strings (e.g. `[52, 53, 54, 55]` → `["52:55"]`).
+  ///
+  /// This is backward-compatible: Yosys-format arrays already mix
+  /// integers with constant strings `"0"` and `"1"`.  Parsers can
+  /// detect range strings by the presence of `:`.
+  final bool compressBitRanges;
```
mkorbel1: why do you want this to be exposed as a configurability option?  any config you have is something you have to support or else deprecate in the future

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
```diff
@@ -0,0 +1,118 @@
...
+  /// results retain complete connectivity.
+  final bool slimMode;
+
+  /// When `true`, contiguous ascending runs of ≥3 integer bit IDs in
+  /// `bits` arrays and cell `connections` arrays are replaced with
+  /// `"start:end"` range strings (e.g. `[52, 53, 54, 55]` → `["52:55"]`).
+  ///
+  /// This is backward-compatible: Yosys-format arrays already mix
+  /// integers with constant strings `"0"` and `"1"`.  Parsers can
+  /// detect range strings by the presence of `:`.
```
mkorbel1, line 92: why do you want this to be exposed as a configurability option?  any config you have is something you have to support or else deprecate in the future.  this applies to the other ones in here too (e.g. enableDCE)

lib/src/synthesizers/netlist/netlist_options.dart
```diff
@@ -0,0 +1,118 @@
...
+  /// `$struct_unpack`, `$struct_pack`), traces each cluster's output
+  /// bits back to their ultimate source bits, and replaces every
+  /// multi-cell cluster with a direct `$buf`.  This subsumes all of
+  /// the individual collapse passes above.
+  final bool collapseTransparentClusters;
+
+  /// When `true`, dead-cell elimination is performed after aliasing to
+  /// remove cells whose inputs are entirely undriven or whose outputs
+  /// are entirely unconsumed.
+  final bool enableDCE;
```
mkorbel1: nit: why not `enableDeadCellElimination`?

lib/src/synthesizers/utilities/synth_operation_namer.dart
```diff
@@ -0,0 +1,143 @@
...
+
+  static int _identityIndex(Iterable<Logic> logics, Logic target) {
+    var index = 0;
+    for (final logic in logics) {
+      if (identical(logic, target)) {
+        return index;
+      }
+      index++;
+    }
+    return -1;
```
mkorbel1: why not return null isntead of -1 (`int?`)?

lib/src/synthesizers/utilities/synth_operation_namer.dart
```diff
@@ -0,0 +1,143 @@
...
+    final module = root.parentModule;
+    if (module == null) {
+      return [0, ...elementPath];
+    }
+
+    final location = _logicLocationInModule(module, root);
+    return [...location, ...elementPath];
+  }
+
+  static List<int> _logicLocationInModule(Module module, Logic root) {
```
mkorbel1: this feels like a weird style for pairs of things, why not a dart record (like a tuple)?

lib/src/synthesizers/utilities/synth_operation_namer.dart
```diff
@@ -0,0 +1,143 @@
...
+    if (inOutIndex >= 0) {
+      return [2, inOutIndex];
+    }
+
+    final internalIndex = _identityIndex(module.internalSignals, root);
+    if (internalIndex >= 0) {
+      return [3, internalIndex];
+    }
+
+    return const [4, 0];
```
mkorbel1: what do these numbers mean? 0-4?

lib/src/synthesizers/utilities/synth_operation_namer.dart
```diff
@@ -0,0 +1,143 @@
...
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:meta/meta.dart';
+import 'package:rohd/rohd.dart';
+import 'package:rohd/src/utilities/sanitizer.dart';
+
+/// Stable names for synthesis-created helper operations.
+@internal
+class SynthOperationNamer {
+  /// Canonical base name for synthesis-created array slice operations.
```
mkorbel1: why dont these constants live in the class they relate to instead of in this central spot?

lib/src/synthesizers/utilities/synth_operation_namer.dart
```diff
@@ -0,0 +1,143 @@
...
+  /// Canonical base name for synthesis-created structure concat operations.
+  ///
+  /// Netlist synthesis uses these when independently driven [LogicStructure]
+  /// fields must be packed back into an aggregate struct output or submodule
+  /// input. The emitted netlist projection represents this as `$struct_pack`.
+  static const String structureConcatOperationName = 'struct_concat';
+
+  SynthOperationNamer._();
+
+  /// Returns the canonical instance name for a synthesis-created operation
```
mkorbel1: why do you need a "canonical" instance name?

lib/src/synthesizers/netlist/netlist_passes.dart
mkorbel1: In these examples it looks like you're trying to removed "named" signals that could be collapsed?  In the SV one we try to respect `Naming`, so if it's named we dont eliminate it.  thoughts on what you're doing with netlist?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,1116 @@
...
+/// Usage:
+/// ```dart
+/// const options = NetlistOptions(collapseTransparentClusters: true);
+/// final synth = NetlistSynthesizer(options: options);
+/// final builder = SynthBuilder(topModule, synth);
+/// final json = synth.synthesizeToJson(topModule);
+/// ```
+class NetlistSynthesizer extends Synthesizer {
+  /// The current format version for netlist JSON produced by this synthesizer.
+  static const String formatVersion = '0.0.5';
```
mkorbel1: what defines and cares about this version?

lib/src/synthesizers/utilities/utilities.dart
mkorbel1: this still has 2025

lib/src/synthesizers/utilities/synth_structure_layout.dart
```diff
@@ -0,0 +1,114 @@
...
+//
+// synth_structure_layout.dart
+// Shared packed LogicStructure layout utility for synthesis backends.
+//
+// 2026 July 10
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:rohd/rohd.dart';
+
+/// Provides bit ranges and field names for a packed [LogicStructure].
```
mkorbel1, line 24: question: this applies to the leaves of a structure if there is hierarchy?

lib/src/synthesizers/synthesis_result.dart
```diff
@@ -23,9 +23,16 @@ abstract class SynthesisResult {
   /// The name of the definition type for this module instance.
   String get instanceTypeName => getInstanceTypeOfModule(module);
 
+  /// Non-fatal warnings reported while producing this synthesis result.
+  final List<String> warnings;
```
mkorbel1: In general, been avoiding adding "warnings" to ROHD -- either things are bad and fatal or good and allowed.  Trying to avoid warning soup that happens in a lot of EDA flows, where they grow forever to the point where it's just slowing down and bloating logs.  Do we really need warnings?

lib/src/synthesizers/synthesizers.dart
```diff
@@ -1,8 +1,10 @@
 // Copyright (C) 2021-2025 Intel Corporation
```
mkorbel1: year

lib/src/synthesizers/synthesizers.dart
```diff
@@ -1,8 +1,10 @@
...
 // SPDX-License-Identifier: BSD-3-Clause
 
+export 'netlist/netlist.dart';
 export 'synth_builder.dart';
 export 'synth_file_contents.dart';
 export 'synthesis_result.dart';
 export 'synthesizer.dart';
 export 'systemverilog/systemverilog.dart';
+export 'utilities/synth_module_stop_policy.dart';
```
mkorbel1: why export this?

mkorbel1 (commented):
can you pull main also again just so it's up to date?

lib/src/synthesizers/utilities/synth_structure_layout.dart
desmonddak: It recurses: _addStructure  line 46
```dart
if (element is LogicStructure && element is! LogicArray) {
  _addStructure(element, offset, path);
}
```

mkorbel1: but i mean if you ask for `a.b.c` vs `a.b` it can give you either one?

lib/src/synthesizers/utilities/synth_operation_namer.dart
desmonddak: This was paralleling the concept of indexed paths introduced by rohd_hierarchy.  By making a pubspec dependency, we can use the `CellOccurrence` and `SignalOccurrence` classes that represent these paths as classes.

lib/src/synthesizers/synthesizers.dart
desmonddak: I scanned for all files that were touched in this PR and extended the copyright.

lib/src/synthesizers/synthesizers.dart
desmonddak: Removed -- it was used only in the netlist synthesizer.

lib/src/synthesizers/synthesis_result.dart
desmonddak: I removed the warnings, and use a throw instead.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: We are adding fields to the netlist format that give enhanced capabilities for our visualizers.  So we need a base format version and check it when we add more features.  This should be 0.0.1 and that should be our baseline version.  Yosys outputs no version at all.  So we can differentiate.  But if we add more fields then when our widget loads it should know the fields are there or not.

lib/src/synthesizers/utilities/synth_operation_namer.dart
desmonddak: This can be resolved by the Namer, so I removed.

lib/src/synthesizers/utilities/synth_operation_namer.dart
desmonddak: I migrated them into the netlist area.  These are for synthesized constructs that occur as we simplify the connectivity.

lib/src/synthesizers/utilities/synth_operation_namer.dart
desmonddak: This had to do with encoding input/output/ioput/unknown.  This is no longer indexed in this way, but there is a dependency on encoding in this particular order (see netlist_synthesizer.dart:1193).

lib/src/synthesizers/utilities/synth_operation_namer.dart
desmonddak: Done.  But it is now in netlist_synthesizer.dart.

lib/src/synthesizers/netlist/netlist_options.dart
desmonddak: Done

lib/src/module.dart
desmonddak: I migrated to a test helper for now.

lib/src/synthesizers/utilities/utilities.dart
desmonddak: I think I finally caught the remaining.

lib/src/synthesizers/netlist/netlist_options.dart
desmonddak: I migrated to use Configuration

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
desmonddak: I made these internal only for now in tests.  The reason is that we really should study migrating these optimizations into the core synthesis traversal.

mkorbel1: I still see `compressBitRanges` as non-internal, is that intentional?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,1232 @@
...
+      // To preserve the structural decomposition in the schematic, we
+      // allocate fresh wire IDs for each array_slice Y output, then
+      // redirect all other cells that consume those IDs as inputs to
+      // read from the fresh IDs instead.  The slice input A keeps the
+      // original parent-array IDs, so the data flow becomes:
+      //   parent (original IDs) → slice A → slice Y (fresh IDs) → consumer
+
+      for (final cellEntry in cells.entries) {
+        if (!cellEntry.key.startsWith(
+          SynthArraySlice.operationName,
```
mkorbel1: is a string "starts with" really a robust method for identifying the type of a module?

lib/src/synthesizers/synthesis_result.dart
mkorbel1: I still see warnings APIs added to synthesis result etc.

lib/src/synthesizers/netlist/netlist_module_translation.dart
```diff
@@ -0,0 +1,777 @@
...
+  final String Function(Module module) _getInstanceTypeOfModule;
+
+  /// The next available integer wire identifier.
+  ///
+  /// Starts at 2 so consumers never confuse wire IDs 0 or 1 with the
+  /// Yosys-JSON constant bit strings `"0"` and `"1"`.
+  int nextId = 2;
+
+  /// Wire identifiers allocated for each synthesis logic.
+  final Map<SynthLogic, List<int>> synthLogicIds = {};
```
mkorbel1: i notice this `synthLogicIds`, `blockedConstSynthLogics`, are not referenced outside of this class.  even though it's an internal class, might be worth considering which pieces should be private to prevent leaky abstractions?

lib/src/synthesizers/netlist/netlist_module_translation.dart
```diff
@@ -0,0 +1,777 @@
...
+    }
+
+    if (lowerIndex != outputIds.length) {
+      return false;
+    }
+
+    concatConnections['Y'] = outputIds.cast<Object>();
+    concatDirections['Y'] = 'output';
+
+    cells['array_concat_output_$concatName'] = {
```
mkorbel1: is this guaranteed unique? does it have to be?

lib/src/synthesizers/netlist/netlist_module_translation.dart
```diff
@@ -0,0 +1,777 @@
...
+    final definition = synthDef;
+    if (definition == null) {
+      return false;
+    }
+
+    for (final instance in definition.subModuleInstantiations) {
+      if (instance.module is! SynthArrayConcat) {
+        continue;
+      }
+      if (instance.outputMapping.values.any(
```
mkorbel1: how often does this scan happen? performance issue with O(n^2) or worse? and similar helper below?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,1232 @@
...
+    SynthesisResult? Function(Module module)? lookupExistingResult,
+    Map<Module, SynthesisResult>? existingResults,
+  }) {
+    final attr = <String, Object?>{'src': 'generated'};
+
+    final translation = NetlistModuleTranslation(module,
+        netlistCellMapper: netlistCellMapper,
+        generatesDefinition: generatesDefinition,
+        getInstanceTypeOfModule: getInstanceTypeOfModule)
+      ..processPorts()
```
mkorbel1: why are these functions not called within the constructor instead of needing to remember to call them later?

lib/src/synthesizers/netlist/netlist_module_translation.dart
```diff
@@ -0,0 +1,777 @@
...
+
+      final submodule = instance.module;
+      final isLeaf = !_generatesDefinition(submodule);
+      final defaultCellType = isLeaf
+          ? submodule.definitionName
+          : _getInstanceTypeOfModule(submodule);
+      final rawPortDirs = <String, String>{};
+      final rawConnections = <String, List<Object>>{};
+
+      for (final (direction, mapping) in [
```
mkorbel1, line 304: the string version of direction is used in multiple places in this file.  would using an enum help make it safer?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,1232 @@
...
+        for (var i = 0; i < nonTrivialFields.length; i++) {
+          final f = nonTrivialFields[i];
+          params['FIELD_${i}_NAME'] = structLayout?.fieldNameAt(f.dstLowerIndex,
+                  fallbackName: f.srcSynthLogic.resolved.name) ??
+              f.srcSynthLogic.resolved.name;
+          params['FIELD_${i}_OFFSET'] = f.dstLowerIndex;
+          params['FIELD_${i}_WIDTH'] = f.dstUpperIndex - f.dstLowerIndex + 1;
+        }
+
+        cells['${cellName}_$structName'] = {
```
mkorbel1: what do you think of a data class that serializes to this structure to help make it more robust that we don't forget some fields (e.g. if fields change/grow in the future)?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,1232 @@
...
+  }
+
+  /// Convenience: synthesize [top] into a combined netlist JSON string.
+  ///
+  /// Builds a [SynthBuilder] internally and returns the full JSON.
+  ///
+  /// The [packageRoot] parameter is accepted for API compatibility with
+  /// downstream trace-enabled branches. [slimMode] overrides the configured
+  /// output mode for this call, allowing expansion after a slim request.
+  String synthesizeToJson(Module top, {String? packageRoot, bool? slimMode}) {
```
mkorbel1, line 1132: this seems only used by tests, add `visibleForTesting`?

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
```diff
@@ -0,0 +1,111 @@
...
+  ///
+  /// When this is provided, it owns the complete stopping policy and
+  /// [leafModuleTypes] is ignored.
+  final SynthModuleStopPolicy? moduleStopPolicy;
+
+  /// Exact [Module.runtimeType]s that should stop netlist hierarchy traversal
+  /// and be emitted as cells in their parent. Defaults to [FlipFlop], which
+  /// contains internal sequential submodules but should be emitted as a `$dff`
+  /// netlist cell.
+  final List<Type> leafModuleTypes;
```
mkorbel1: should this be private with an `UnmodifiableListView` accessor (if needed)?

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
```diff
@@ -0,0 +1,111 @@
...
+class NetlistSynthesizerConfiguration {
+  /// The policy used to decide which modules stop hierarchy traversal and are
+  /// emitted as cells in their parent instead of as separate module
+  /// definitions. When `null`, [SynthModuleStopPolicy.netlist] is used.
+  ///
+  /// When this is provided, it owns the complete stopping policy and
+  /// [leafModuleTypes] is ignored.
+  final SynthModuleStopPolicy? moduleStopPolicy;
+
+  /// Exact [Module.runtimeType]s that should stop netlist hierarchy traversal
```
mkorbel1: `runtimeType` is maybe a little too delicate?  First, it's sort of weird when you compile to javascript with optimization.  Second, it means it doesn't handle inheritance gracefully (e.g. if someone makes a custom version that  `extends FlipFlop` they need to remember to update this policy too).  Would a `Function` be better with a default that checks `is FlipFlop` be better?

lib/src/module.dart
```diff
@@ -927,6 +927,11 @@ abstract class Module {
     return outPort;
   }
 
+  bool _hasConsts(LogicStructure structure) => structure.elements.any(
```
mkorbel1: why did you change this from the private extension on LogicStructure to a method on Module?

test/synth_name_parity_test.dart
```diff
@@ -0,0 +1,378 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// synth_name_parity_test.dart
+// Tests that verify canonicalNameOf works consistently across
```
mkorbel1: there's no more `canonicalNameOf`?

pubspec.yaml
```diff
@@ -5,6 +5,7 @@ homepage: https://intel.github.io/rohd-website
 repository: https://github.com/intel/rohd
 issue_tracker: https://github.com/intel/rohd/issues
 documentation: https://intel.github.io/rohd-website/docs/sample-example/
+publish_to: none
```
mkorbel1: no :) we do want to publish rohd

is it complaining that the path dependency is not allowed if we publish?  what's the right way to handle this then?  do we need to release rohd_hierarchy also?

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
desmonddak: Made internal

pubspec.yaml
desmonddak: We want to be able to use the local version when we are in this repo.  I will try to figure this out, but it is the monorepo problem with multiple packages.  You don't want to be in a loop where you have to publish to test a change that crosses between the package and the rest of the repo.  But it complained about using a path for an internal package.

WE do need to release rohd_hierarchy in pub.dev.  Perhaps this is a dependency override situation.

desmonddak: https://dart.dev/tools/pub/workspaces

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: compressBitRanges?  
image

lib/src/synthesizers/utilities/synth_structure_layout.dart
desmonddak: Yes.  It gives you the range of bits for whatever field you choose regardless of what level of hierarchy.
I added a method that will export this range.
```dart
 test('returns bit ranges for nested field paths', () {
      final nested = LogicStructure([
        Logic(name: 'b', width: 2),
        LogicStructure([
          Logic(name: 'd', width: 3),
        ], name: 'c'),
      ], name: 'a');
      final structure = LogicStructure([
        Logic(name: 'prefix'),
        nested,
      ]);
      final layout = SynthStructureLayout(structure);

      expect(layout.bitRangeForPath('a'), (start: 1, end: 6));
      expect(layout.bitRangeForPath('a.b'), (start: 1, end: 3));
      expect(layout.bitRangeForPath('a.c'), (start: 3, end: 6));
      expect(layout.bitRangeForPath('a.c.d'), (start: 3, end: 6));
      expect(layout.bitRangeForPath('a.missing'), isNull);
    });
```

lib/src/synthesizers/synthesis_result.dart
desmonddak: I removed the dead plumbing for collecting ...

test/synth_name_parity_test.dart
```diff
@@ -0,0 +1,378 @@
...
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:rohd/rohd.dart';
+import 'package:rohd/src/synthesizers/utilities/utilities.dart';
+import 'package:test/test.dart';
+
+import '../example/filter_bank.dart';
+
+extension _NetlistTestModule on Module {
+  String generateNetlist(
```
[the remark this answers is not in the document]
desmonddak: This is correct.  We will not be adding to the `Module` API for new services, we will use the service model.  Only for SV and Waves do we have these generate routines.  This is as agreed, right -- the helper routines are for simplifying the most common two cases.

lib/src/synthesizers/netlist/netlist_cell_mapper.dart
```diff
@@ -0,0 +1,476 @@
...
+          cn['_resetValue'] = ctx.rawConns[rstVal] ?? [];
+        }
+        return (
+          cellType: r'$dff',
+          portDirs: pd,
+          connections: cn,
+          parameters: <String, Object?>{
+            'WIDTH': ctx.width(d),
+            'CLK_POLARITY': 1,
+          },
```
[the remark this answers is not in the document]
desmonddak: I expanded the output to properly map all our variants to these names.

lib/src/synthesizers/netlist/netlist_cell_mapper.dart
```diff
@@ -0,0 +1,476 @@
...
+    // ── TriStateBuffer → $tribuf ──────────────────────────────────────
+    register((ctx) {
+      if (ctx.module is! TriStateBuffer) {
+        return null;
+      }
+      final tsb = ctx.module as TriStateBuffer;
+      final inName = tsb.inputs.keys.first; // data input
+      final enName = tsb.inputs.keys.last; // enable
+      final outName = tsb.inOuts.keys.first; // inout output
+      final r = ctx.remap({inName: 'A', enName: 'EN', outName: 'Y'});
```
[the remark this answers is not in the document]
desmonddak: Fixed.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
```diff
@@ -0,0 +1,1232 @@
...
+      'the destination has no parent module.',
+    );
+  }
+
+  final rootLocation = _logicLocationInModule(module, destination);
+  final elementPath = _logicElementPathIndices(destination).path;
+  final parts = [
+    ..._modulePathIndices(module).path,
+    ...rootLocation.path,
+    ...elementPath,
```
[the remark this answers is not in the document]
desmonddak: Fixed.  path is now local to module

lib/src/synthesizers/netlist/netlist_module_translation.dart
desmonddak: We now reuse the mechanisms for other netlist synthesized structures like struct pack/unpack (this is the array pack equivalent)

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
desmonddak: We removed this -- it had a class based check so this was unneeded.

lib/src/module.dart
desmonddak: I finally migrated this to `LogicStructure` itself.

test/synth_name_parity_test.dart
desmonddak: Removed.  This is no longer the unique namer.

lib/src/synthesizers/netlist/netlist_cell_mapper.dart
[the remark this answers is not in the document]
desmonddak: We have a routine that outputs all supported leaf types into a netlist.

lib/src/synthesizers/netlist/netlist_cell_mapper.dart
```diff
@@ -0,0 +1,476 @@
...
+          GreaterThanOrEqual: r'$ge',
+        },
+        (ctx, type) =>
+            binaryABY(ctx, type, inAPrefix: '_in0', inBPrefix: '_in1'),
+      ),
+      (
+        const <Type, String>{
+          LShift: r'$shl',
+          RShift: r'$shr',
+          ARShift: r'$shiftx',
```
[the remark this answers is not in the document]
desmonddak: Fixed.

example/filter_bank/filter_bank.dart
```diff
@@ -0,0 +1,206 @@
...
+    this.numChannels = 2,
+    LogicNet? dataBus,
+    Logic? writeEnable,
+    super.name = 'FilterBank',
+    String? definitionName,
+  }) : super(definitionName: definitionName ?? 'FilterBank') {
+    if (coefficients.length != numChannels) {
+      throw Exception(
+          'coefficients must have $numChannels entries (one per channel).');
+    }
```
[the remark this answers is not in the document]
desmonddak: Validation now covers the full space.

example/filter_bank/filter_sample.dart
```diff
@@ -0,0 +1,51 @@
...
+// LogicStructure sample word for the polyphase FIR filter bank example.
+//
+// 2025 March 26
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:rohd/rohd.dart';
+
+/// A structured signal bundling a data sample with metadata.
+///
+/// Packs three fields — [data], and [valid] — into a single
```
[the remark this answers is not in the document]
desmonddak: Fixed.

lib/src/synthesizers/netlist/netlist_cell_mapper.dart
```diff
@@ -0,0 +1,476 @@
...
+        final pd = <String, String>{'A': 'input', 'B': 'input', 'Y': 'output'};
+        final cn = <String, List<Object>>{
+          'A': ctx.rawConns[in0] ?? [],
+          'B': ctx.rawConns[in1] ?? [],
+          'Y': ctx.rawConns[sumName] ?? [],
+        };
+        if (carryName.isNotEmpty) {
+          pd['CO'] = 'output';
+          cn['CO'] = ctx.rawConns[carryName] ?? [];
+        }
```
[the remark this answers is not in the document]
desmonddak: Fixed

lib/src/synthesizers/netlist/netlist_module_translation.dart
desmonddak: I scanned for other leakiness and fixed.

lib/src/synthesizers/netlist/netlist_synthesizer_configuration.dart
desmonddak: Fixed

pubspec.yaml
desmonddak: Turns out we don't yet depend on rohd_hierarchy internally (yet), so for this PR, we can just omit the dependency and not have to deal with the conflict of a published package needed an internal dependency.

example/filter_bank/mac_unit.dart
```diff
@@ -0,0 +1,86 @@
...
+    clk = addInput('clk', clk);
+    reset = addInput('reset', reset);
+    enable = addInput('enable', enable);
+    final result = addOutput('result', width: dataWidth);
+
+    // A 2-stage pipeline: multiply, then accumulate
+    final pipe = Pipeline(
+      clk,
+      reset: reset,
+      stages: [
```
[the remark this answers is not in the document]
desmonddak: We now stall on `~enable` and have a mac_unit_test.dart to demonstrate this pipeline example.

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: It's six fields, but it does make sense to really structure this so that it doesn't appear ad-hoc and there is the potential to simply add a field in one place and enforce, and not have entropy across these cell emitters.  This is stuff that may end up in the core synthesis rather than just making netlists pretty, so it is good to do it here and have a good reference.

lib/src/synthesizers/netlist/netlist_passes.dart
mkorbel1: I don't see any reference to `Naming` in the netlist passes.  What is the resolution here?  Do we not care to maintain the requested names or am I missing the mechanism?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
mkorbel1: huh?  i'm talking about checking if things start with the `operationName` string from these special synth objects.  that seems like it opens a door to surprise conflicts if someone happens to name something starting with `array_slice` which isn't too crazy, no?

lib/src/modules/conditionals/flop.dart
```diff
@@ -92,6 +92,10 @@ class FlipFlop extends Module with SystemVerilog {
   /// reset. If no `reset` is provided, this will have no effect.
   final bool asyncReset;
 
+  /// The constant reset value, or `null` when reset is absent or data-driven.
+  LogicValue? get constantResetValue =>
```
mkorbel1, line 96: this is still here, remove

tool/generate_gate_catalog.dart
mkorbel1, line 1: can you explain a bit why we need to check-in the generated JSON?

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: This was in response to:
```
I still see compressBitRanges as non-internal, is that intentional?
```

There was no button to respond to it.

I'll look at this `operationName` issue...

lib/src/modules/conditionals/flop.dart
desmonddak: Could you look at the comment at the end of the module_services_api PR #673 
I explain it there why we need `constantResetValue`.

tool/generate_gate_catalog.dart
desmonddak: We don't have to have it.  I'll delete.  I used it to validate that nothing changed at leaf level.

mkorbel1: did it add meaningful coverage that we should replace with something else?

lib/src/modules/conditionals/flop.dart
mkorbel1: got it, sounds good, thank you!

For future reference, it's this comment:
https://github.com/intel/rohd/pull/673#issuecomment-5399994990

lib/src/synthesizers/netlist/netlist_synthesizer.dart
desmonddak: I created a structure that tells us what kind of cell and if it is synthesized.

mkorbel1 (approved):
looks great!

lib/src/synthesizers/netlist/netlist_passes.dart
desmonddak: 'preserve Logic with reserved and renameable, but allow mergeable 
to be optimized' -- I think we will end up pushing netlist_passes into the synth area and be compatible with the optimizations already there.

mkorbel1 (approved):
naming change looks reasonable, but testing doesn't seem comprehensive.  since the impact is just potentially the netlist, seems good enough for now to me.  approved!

lib/src/synthesizers/netlist/netlist_passes.dart
mkorbel1: i think this looks good!

PR #676, opened 2026-07-08
Range/range assignment collapsing

mkorbel1:
## Description & Motivation

This PR improves generated SystemVerilog for contiguous partial array and range assignments by collapsing them into packed slice assignments when it is safe to do so.

It adds an internal range-to-range synth assignment representation and teaches the SystemVerilog backend to render both destination and source slices. The base synth stack now recognizes several related range-collapse patterns, including contiguous array element runs, bus subset sources feeding arrays, chained range assignments, generated `assignSubset` swizzle paths, and wide temporary slice helpers.

The optimization is intentionally conservative around naming, clearability, extra consumers, nets, constants, nested arrays, submodule uses, and aggregate connection inlining so that helper signals are only removed when all replaced uses are accounted for.

### Details

- Adds `RangeSynthAssignment` for assignments where both source and destination are ranges.
- Collapses contiguous partial array/range assignments into compact packed slice assignments.
- Preserves source ranges during SystemVerilog rendering instead of expanding them into whole-source concatenations.
- Moves these optimizations into the base synth flow so they benefit generated outputs broadly.
- Adds regression coverage for partial array ranges, bus subset-to-array cases, generated subset paths, extra consumers, wide temporaries, chained ranges, and aggregate concat behavior.
- Tags benchmark smoke tests so non-benchmark validation can run without executing benchmarks.

## Related Issue(s)

N/A

## Testing

Added new tests, existing tests cover a lot

## Backwards-compatibility

No

## Documentation

No


lib/src/synthesizers/systemverilog/systemverilog_synthesis_result.dart
```diff
@@ -143,21 +143,37 @@ class SystemVerilogSynthesisResult extends SynthesisResult {
...
           ' bidirectional net connections.');
 
-      var sliceString = '';
-      if (assignment is PartialSynthAssignment && assignment.width > 1) {
-        sliceString = assignment.dstUpperIndex == assignment.dstLowerIndex
-            ? '[${assignment.dstUpperIndex}]'
-            : '[${assignment.dstUpperIndex}:${assignment.dstLowerIndex}]';
+      var dstSliceString = '';
+      var srcSliceString = '';
+      if (assignment is RangeSynthAssignment) {
```
desmonddak, line 175: We moving to trailing commas now? :)

lib/src/synthesizers/utilities/synth_module_definition.dart
```diff
@@ -1173,6 +1233,1299 @@ class SynthModuleDefinition {
...
+              ),
+              sourceSubmodule: src.sourceSubmodule,
+              sourceSignal: src.sourceSignal,
+            );
+      final dst = _simpleAssignmentDestinationRange(assignment);
+      if (resolvedSrc == null ||
+          dst == null ||
+          resolvedSrc.range.width != 1 ||
+          dst.width != 1 ||
+          resolvedSrc.range.base == dst.base ||
```
desmonddak, line 1681: Trying to follow the logic here -- Why do the bases have to be the same?  As long as the ranges match, we can assign?  Would be helpful if a comment could be added like 'this is is assigning a[4:1] <= b[16:13]'

desmonddak (approved):
Could we add a controlling bool or set of bools to allow us to back out this optimization if we encounter a problem or want to see the difference with and without?

I could use a top level algorithm description as I got a bit lost in what was a group and a covering set, and  how initial assignments got accumulated into wider assignents  versus the collapsing of assignments in propagation direction (width versus depth).

lib/src/synthesizers/systemverilog/systemverilog_synthesis_result.dart
mkorbel1: either way is acceptable pre-new-formatter :)  part of why I like the new formatter removing that option

mkorbel1:
> Could we add a controlling bool or set of bools to allow us to back out this optimization if we encounter a problem or want to see the difference with and without?

I don't really like the idea of configurable optimizations, personally.  It increases the surface area for bugs by every switch and makes it harder for users to decide what they want.  I'd rather make this thing rock-solid and enabled always.


I agree on the points about improving documentation in the code explaining the logic, will do.

lib/src/synthesizers/utilities/synth_module_definition.dart
mkorbel1: Added more comments in a few places explaining the logic more

PR #678, opened 2026-07-10
Configurable SV generation, port object and data types

mkorbel1:
## Description & Motivation

Adds configurable object and data type declarations for generated SystemVerilog ports.

Generated ports now default to fully explicit declarations:

- Inputs: `input wire logic`
- Outputs: `output var logic`
- Inouts: `inout wire logic`

A new `SystemVerilogSynthesizerConfiguration` allows object and data types to be independently explicit or implicit using `SystemVerilogPortType`.

The configuration can be supplied through either `Module.generateSynth` or `SystemVerilogSynthesizer`:

```dart
final sv = module.generateSynth(
  configuration: const SystemVerilogSynthesizerConfiguration(
    portObjectType: SystemVerilogPortType.implicit,
    portDataType: SystemVerilogPortType.explicit,
  ),
);
```

The configuration passed to SystemVerilogSynthesisResult is an optional named argument with a default value, preserving compatibility with existing constructor calls.

This also updates the generation documentation, changelog, and existing SystemVerilog expectations.

## Related Issue(s)

N/A

## Testing

Added new tests and updated tests

## Backwards-compatibility

This adds to the API for SV synth stack as a new feature, should be backwards compatible.

## Documentation

No, API docs updated

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -2,7 +2,7 @@
 // README at: https://github.com/devcontainers/templates/tree/main/src/ubuntu
 
 {
-  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
 
   "updateContentCommand": "tool/gh_codespaces/run_setup.sh",
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,6 @@
 ## Next release
 
-- Added configurable explicit or implicit object and data types for generated SystemVerilog ports, defaulting to `input wire logic`, `output var logic`, and `inout wire logic`.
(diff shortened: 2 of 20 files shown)


.devcontainer/devcontainer.json
```diff
@@ -2,7 +2,7 @@
 // README at: https://github.com/devcontainers/templates/tree/main/src/ubuntu
 
 {
-  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-22.04",
+  "image": "mcr.microsoft.com/devcontainers/base:ubuntu-24.04",
```
desmonddak, line 5: I'm a bit worried about this -- I have a lot of external ingredients (e.g. Rust) that may get broken before they get migrated in and further delay devtools integration.
Do we have to migrate to 24.04 right now?

mkorbel1: I moved in this PR because 22.04 had an older iverilog version that didn't support as much SV syntax and didn't work with the new port declaration behavior.  do you have any specific concerns/issues with this?  alternative would be build our own iverilog in the container, but that's slower for every CI and also seems like we should move up to 24.04 sooner rather than later anyways

PR #679, opened 2026-07-14
Add support for preferred radix on `Const`s

### LINKED ISSUE #601: Ability to control displayed radix of constants
### Motivation

When creating a `Const` (or even `LogicValue`?), it is sometimes nice to be able to view the constant in a particular radix.  For `LogicValue`, there's already methods to convert, but in generated outputs (e.g. SystemVerilog), it's automatically selected.  It would be nice if we could control on a per-`Const` basis the "preferred" radix to display.

### Desired solution

Some sort of option to control radix printing for generated outputs.

### Alternatives considered

_No response_

### Additional details

_No response_

mkorbel1:
## Description & Motivation

Adds support for a `preferredRadix` on `Const`

## Related Issue(s)

Fix #601 

## Testing

Added new tests

## Backwards-compatibility

No

## Documentation

Yes, included

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 ## Next release
 
+- Added per-`Const` preferred radix control for generated literals, normalized `Const` names to reflect their displayed radix, and enhanced `LogicValue.toRadixString` to omit its width and radix decorator or digit separators (<https://github.com/intel/rohd/issues/601>).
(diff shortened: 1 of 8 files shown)


desmonddak (approved):
Looks good.
One question is how we propagate to other displays like wave and schematic viewer which are remote.  We likely will need a public form for querying the preferred radix.

mkorbel1:
> One question is how we propagate to other displays like wave and schematic viewer which are remote. We likely will need a public form for querying the preferred radix.

`preferredRadix` is visible on `Const` so that should be all you need?

PR #680, opened 2026-07-15
Prevent stack overflow when synthesizing combinational loops

mkorbel1:
## Description & Motivation

- Detect dependency cycles among inlineable SystemVerilog submodules.
- Keep cyclic modules materialized instead of recursively expanding them.
- Preserve combinational loops in generated SystemVerilog without changing circuit behavior.

## Related Issue(s)

Fixes #641 

## Testing

- Add a non-constant regression test that reproduces the original stack overflow.

## Backwards-compatibility

No

## Documentation

N/A


desmonddak (approved):
You defused a bomb here.

PR #681, opened 2026-07-15
Add rohd waveform package

desmonddak:
## Description & Motivation

This is a package used to communicate from ROHD debugging to the ROHD waveform viewer.  It can supply waveform services to other tools that query for incremental waveform updates as the simulator is running.

## Related Issue(s)

None.

## Testing

Local standalone tests are included.  It has also been tested in a full path through a remote copy of the ROHD waveform viewer (upcoming release).

## Backwards-compatibility

No.  This is a new capability.

## Documentation

None.  This is an internal protocol which we will document when stable.


packages/rohd_waveform/README.md
```diff
@@ -0,0 +1,29 @@
...
+- `ModuleStructure` — top-level waveform structure (metadata + hierarchy roots).
+- `SignalWaveform` — time-series waveform data with a backpointer to signal
+  metadata in `rohd_hierarchy`.
+- `WaveformData` — transfer object for incremental waveform updates.
+- `Data`, `WaveFormat`, and `MetaData` — waveform data primitives.
+
+For hierarchy types such as `HierarchyOccurrence` and `SignalOccurrence`,
+import `package:rohd_hierarchy/rohd_hierarchy.dart` directly.
+
+## Status
```
mkorbel1: why is this "status"?

packages/rohd_waveform/pubspec.yaml
```diff
@@ -0,0 +1,20 @@
...
+issue_tracker: https://github.com/intel/rohd/issues
+
+publish_to: none
+
+environment:
+  sdk: '>=3.0.0 <4.0.0'
+
+dependencies:
+  equatable: ^2.0.5
+  rohd_hierarchy:
```
mkorbel1: should probably have the path dep as a dependency override rather than the actual dependency?

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/regex_utils.dart
mkorbel1, line 1: why do we need these?

test/benchmark_test.dart
```diff
@@ -20,7 +20,7 @@ import '../benchmark/ssa_driver_search_benchmark.dart';
 import '../benchmark/wave_dump_benchmark.dart';
 
 void main() {
-  group('benchmark', tags: 'benchmark', () {
+  group('benchmark', () {
```
mkorbel1: why remove the tag? it's useful for excluding to run faster (`dart test -x benchmark`)

packages/rohd_waveform/lib/src/models/metadata.dart
```diff
@@ -0,0 +1,107 @@
...
+  /// The timescale of the metadata (e.g., "1ns", "100ps").
+  final String timescale;
+
+  /// The timescale factor (e.g., 1, 10, 100).
+  ///
+  /// This is optional and populated when loading from waveform files.
+  final int? timescaleFactor;
+
+  /// The date of the metadata.
+  final String date;
```
mkorbel1, line 29: should this be a date object instead of string?

packages/rohd_waveform/lib/src/models/data.dart
```diff
@@ -0,0 +1,34 @@
...
+
+/// A class that represents the data of a signal.
+///
+/// It contains a time and a value.
+class Data {
+  /// The time of the data.
+  int time;
+
+  /// The value of the data.
+  String value;
```
mkorbel1, line 18: why string not `LogicValue` or something?

packages/rohd_waveform/lib/src/models/wave_format.dart
```diff
@@ -0,0 +1,100 @@
...
+  ///
+  /// GHW is the native format for GHDL simulations. Reading is supported,
+  /// but writing is not.
+  ghw,
+
+  /// Unknown or unsupported format.
+  unknown;
+
+  /// Returns the file extension for this format.
+  String get extension {
```
mkorbel1: i think you could make a field of the num called `extension` or something, then you can allow default to be `name.toString()` and then all of these functions can become automatically updated by however many enums are added?

packages/rohd_waveform/lib/src/models/wave_format.dart
```diff
@@ -0,0 +1,100 @@
...
+        return WaveFormat.fst;
+      case 'ghw':
+        return WaveFormat.ghw;
+      default:
+        return WaveFormat.unknown;
+    }
+  }
+
+  /// Whether this format supports writing.
+  bool get supportsWriting {
```
mkorbel1: who supports writing?  rohd wave dumping?  why is that info necessary here?

packages/rohd_waveform/lib/src/models/signal_data_service.dart
```diff
@@ -0,0 +1,102 @@
...
+  final Map<String, dynamic>? metadata;
+
+  /// Creates a combined signal definition and waveform payload.
+  WaveData({required this.port, required this.data, this.metadata});
+
+  /// Get signal name from the shared signal model.
+  String get signalName => port.name;
+
+  /// Get signal direction from the shared signal model.
+  String get signalDirection => port.direction ?? 'inout';
```
mkorbel1, line 89: why is default inout?

packages/rohd_waveform/lib/src/models/signal_waveform.dart
```diff
@@ -0,0 +1,325 @@
...
+  static SignalLookup? signalLookup;
+
+  /// Clears the signal lookup function.
+  static void clearSignalLookup() {
+    signalLookup = null;
+  }
+
+  /// Whether this waveform was computed/synthesized (e.g. gate evaluation)
+  /// rather than directly fetched from the VM service.
+  bool isComputed;
```
mkorbel1: should this be private with a getter (and some others in here) to be safer?

packages/rohd_waveform/lib/src/models/signal_waveform.dart
```diff
@@ -0,0 +1,325 @@
...
+
+  /// Override width for computed sub-field waveforms whose signal metadata
```
mkorbel1: why wouldn't it be available?

packages/rohd_waveform/lib/src/models/signal_waveform.dart
```diff
@@ -0,0 +1,325 @@
...
+  /// Clears all waveform data.
+  void clearData() {
+    data.clear();
+  }
+
+  /// Gets the value at a specific time using binary search.
+  ///
+  /// Returns the value of the last data point at or before the given time.
+  /// If no data point exists at or before the time, returns the first value.
+  String getValueByTime(int time) {
```
mkorbel1, line 291: could you use this?
https://api.flutter.dev/flutter/package-collection_collection/binarySearch.html

packages/rohd_waveform/lib/src/waveform_api.dart
```diff
@@ -0,0 +1,102 @@
...
+//
+// rohd_waveform.dart
+// An abstract class that defines the API for module structure.
+//
+// 2024 January 29
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:rohd_waveform/rohd_waveform.dart';
+
+/// An abstract class that defines the API for module structure.
```
mkorbel1: why are there two comments here  saying "an abstract class"?  typo?

packages/rohd_waveform/test/metadata_test.dart
```diff
@@ -0,0 +1,73 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// metadata_test.dart
+// Unit tests for the MetaData and Data models.
+//
+// 2026
```
mkorbel1: you can put yourself as author?

packages/rohd_waveform/test/signal_waveform_test.dart
mkorbel1, line 1: tests seem thin/light compared to how much stuff is in here.  how's code coverage?

packages/rohd_waveform/lib/src/models/data.dart
desmonddak: THis package does nto depend on rohd, just rohd_hierarchy.  We could move these packages out if we wanted.  They need to be central for the widgets to communicate.

packages/rohd_waveform/lib/src/models/metadata.dart
desmonddak: It's coming out of VCD/FST in a non-standard format for both. manipulating it into a Date and back would likely not be worth it.  These formats don't enforce a format.

packages/rohd_waveform/lib/src/models/signal_waveform.dart
desmonddak: Not every signal needs to be computed by the `WaveformService`.  We can optimize for minimizing the amount of store and send by knowing that the client can compute some waveforms on the fly.  This can dramatically reduce the store.

packages/rohd_waveform/lib/src/models/signal_waveform.dart
desmonddak: Unfortunately that package does exact match and they really should enhance to give closest higher or lower.

packages/rohd_waveform/lib/src/waveform_api.dart
desmonddak: Removed

packages/rohd_waveform/lib/src/models/wave_format.dart
desmonddak: That helped a lot in simplifying this code.

packages/rohd_waveform/lib/src/models/wave_format.dart
desmonddak: These are not called and I will remove.

packages/rohd_waveform/lib/src/models/signal_data_service.dart
desmonddak: If you load a netlist and have no direction on a signal, it is the best assumption.

packages/rohd_waveform/lib/src/models/signal_waveform.dart
desmonddak: Yes, I restructured this class.

packages/rohd_waveform/test/metadata_test.dart
desmonddak: Done and I scanned the remaining files for header issues.

packages/rohd_waveform/test/signal_waveform_test.dart
desmonddak: I added coverage tests.

test/benchmark_test.dart
desmonddak: I added back.  I was getting a warning during coverage that I was worried about blocking CI.

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/regex_utils.dart
desmonddak: We don't.  This was a redirection layer when I was experimenting with other regex (in waveform viewer).  It can call its own for now.  The concept was to share regex in widgets to have the same functionality.

Removed.

packages/rohd_waveform/lib/src/models/signal_waveform.dart
desmonddak: I did make it use some of the collections capabilities.

desmonddak:
I added flutter test coverage and a set of durable tests that should survive the downstream merges as a core set of tests for the rohd_devtools_extension as well.

test/benchmark_test.dart
mkorbel1: i still see it missing?

packages/rohd_waveform/lib/src/models/signal_waveform.dart
mkorbel1: i still see override* as non-private and non-final?

packages/rohd_waveform/README.md
mkorbel1, line 1: good to throw a copyright footer on these package-level README's (see main README)

rohd_devtools_extension/tool/generate_coverage.sh
mkorbel1, line 1: is this intended to generate a badge for display somewhere like the main rohd coverage badge?

test/benchmark_test.dart
desmonddak: If fixed in another branch, not this one.

packages/rohd_waveform/lib/src/models/signal_waveform.dart
desmonddak: Refactored to be an argument of `SignalWaveform`

rohd_devtools_extension/tool/test_devtools_install.dart
```diff
@@ -5,7 +5,6 @@
 // Smoke-test that Flutter DevTools can discover the installed ROHD DevTools
 // extension from a package root or extension/devtools install directory.
```
mkorbel1, line 6: list yourself as author?

rohd_devtools_extension/tool/view_coverage.sh
mkorbel1, line 1: why is this necessary?  can't you just open the html file?

tool/view_coverage.sh
mkorbel1, line 1: duplicate file -- do we need to?

rohd-multipackage.code-workspace
mkorbel1, line 1: what's this?

desmonddak: Ack!  So I keep having this problem that VSCode cannot understand having packages in the repo with their own pubspec.yaml.  This is a way to be in the same VSCode session but open a different workspace.  I have not verified that this is lighter than just opening a new window.  But it quieted the lengthy Problems I was seeing when I opened the toplevel.  I'll see if I still need it.

rohd_devtools_extension/tool/view_coverage.sh
desmonddak: Need a server.  It doesn't just display.

rohd_devtools_extension/tool/generate_coverage.sh
desmonddak: I hadn't planned. It was just to generate coverage locally.  But I thought I fixed the top-level generate_coverage.sh script to allow you to be in a subdir and just do coverage there (say in a package).

rohd-multipackage.code-workspace
mkorbel1: if it's useful in general to keep then great!  ill leave it to your testing

rohd_devtools_extension/tool/view_coverage.sh
mkorbel1: from copilot:

> No—an lcov/genhtml coverage report is static, so coverage/html/index.html can normally be opened directly. The existing generator even instructs users to do that at [generate_coverage.sh line 57](https://github.com/intel/rohd/blob/6b5716ea47780a9f04f4925fc9d703dbbf24f39e/rohd_devtools_extension/tool/generate_coverage.sh#L55-L57).
> 
> A server is useful only when:
> 
> - running in a remote container/VM where the report must be reached through port forwarding;
> - the pages use browser features blocked under file://—unlikely for standard genhtml output;
> another tool expects an HTTP URL.
> - If HTTP is needed, binding to 127.0.0.1 is safer than exposing it on every interface with 0.0.0.0

if you double click the html file in windows or open it in chrome it doesnt work for you?

rohd_devtools_extension/tool/generate_coverage.sh
mkorbel1: ah that would be better!

rohd_devtools_extension/tool/view_coverage.sh
desmonddak: I run in a container which has a limited ability to install a quality browser.
I made these scripts run in the current directory, to localize coverage stats.
I added a --force flag for view to regenerate stats

tool/view_coverage.sh
desmonddak: I removed the lower level duplicates.

rohd-multipackage.code-workspace
desmonddak: File -> Open Workspace from File definitely works to open packages as distinct repo roots and then the analyzer can understand the boundaries.  So, I'll add a comment at the top of the file.

packages/rohd_hierarchy/lib/src/hierarchy_occurrence_traversal.dart
mkorbel1, line 1: question: why an extension?

mkorbel1 (approved):
great!

packages/rohd_hierarchy/lib/src/hierarchy_occurrence_traversal.dart
desmonddak: By extension, I think you mean the additional changes added:  the hierarchy package will get more capability that operates only with Occurrence handles.  So find children, find leaf children, fanin, fanout, search. And that means we can centralize all this for devtools and for other applications needing design traversal, even at a distance like devtools.  So this is baseline behavior entered here with branches adding more later.

mkorbel1: i meant `extension HierarchyOccurrenceTraversal on HierarchyOccurrence`

PR #682, opened 2026-07-15
Optimize constant operations and enforce Const immutability

### LINKED ISSUE #429: Avoid module creation for simple constant scenarios in gates

### LINKED ISSUE #486: Protect `Const` from having its value modified
### Motivation

Currently, `Const` is just a `Logic` that initializes itself with a specific value.  Technically, the value could be changed later via a `put` directly on the `Const` or on another `Logic` which shares an underlying `_Wire`.  This can be confusing for debug, and worse, could cause synthesized outputs (e.g. SystemVerilog) to have constant values which do not match the requested value at the time of `Const` creation.

### Desired solution

Some ideas:
- Have a way to flag/throw an exception when the value of a `Const` is attempted to be changed
- Have a way to ensure synthesized outputs have the intended constant value

### Alternatives considered

_No response_

### Additional details

Somewhat related to #429

mkorbel1:
## Description & Motivation

This PR supersedes #481 and continues the work started there. The original commits from #481 are included in this branch so that @mjayasim9 remains credited for her contributions.

This completes and expands the constant optimizations described in #429:

- Moves constant-specific operation handling into `Const`.
- Avoids creating `Module`s for constant bitwise operations, reductions, comparisons, and shifts when the result can be computed immediately.
- Returns the original signal for shifts by a known constant zero.
- Avoids creating a `Mux` when its control is a valid `Const`.
- Preserves four-state semantics by retaining gates where returning an operand directly would incorrectly preserve `z` instead of producing `x`.
- Preserves existing width validation and exception behavior.

This also addresses #486 by preventing a `Const` value from being changed through `put` or `inject`, including attempts made through another `Logic` driven by that `Const`. These attempts now throw an `UnassignableException` that identifies the signal through which the update was attempted and the immutable signal driving it.

Together, these changes improve simulation performance, reduce unnecessary generated SystemVerilog, and ensure that simulated and synthesized constant values remain consistent.

## Related Issue(s)

Closes #429.

Closes #486.

Supersedes #481.

## Testing

Added and expanded tests covering:

- Constant folding for bitwise NOT, AND, OR, and XOR.
- Constant unary reductions and equality comparisons.
- Constant shifts and shifts by zero.
- Muxes with valid and invalid constant controls.
- Width mismatch behavior.
- Four-state `x` and `z` behavior, including cases that must not be optimized into direct operand reuse.
- Direct attempts to mutate a `Const`.
- Attempts to mutate a signal driven by a `Const`.
- Preservation of the original constant value in generated SystemVerilog after rejected mutation attempts.

## Backwards-compatibility

Narrowly, yes. Code that directly modifies a `Const`, or modifies another `Logic` driven by a `Const`, through `put` or `inject` will now throw an `UnassignableException`.

This is an intentional behavior correction for #486. Mutating a `Const` contradicted its contract and could cause simulation results to disagree with synthesized output. Normal uses of `Const` and existing operation APIs are otherwise unchanged.

## Documentation

Yes. API documentation for `Const` and constant-controlled mux behavior has been updated.


desmonddak (approved):
Looks great.
I'm curious about simulation performance improvements, especially the shift operators.

PR #685, opened 2026-07-20
Systemc leaf full

desmonddak:
## Description & Motivation

This is an effort to rearchitect the way synthesizers generate leaf code.  It is highly disruptive because it removes the assumption that the gates produce SystemVerilog code and instead they all call a backend of some kind to produce the gate code as well as operators that produce inline code.  So the architecture is language agnostic and symmetric with respect to the two languages we currently have: SystemVerilog and SystemC, as a way of stressing the architecture.

## Related Issue(s)

None.  A discussion in PR #662 led to this exploration.

## Testing

SV and SC tests in ROHD all pass.

## Backwards-compatibility

Likely:  while not a user api, it is likley there are some public methods that have been changed or removed.

## Documentation

None included.  Not likely needed except to document how to handle new gates and possibly custom block handling (which may not be working yet).


desmonddak:
This refactoring is not quite complete.  The gates portion is mature.  The conditionals still have generateVerilog methods.
But generally, this is an architecture to generalize the backend and make the synthesizer architecture symmetric.

PR #688, opened 2026-07-21
Cleaning up dependencies

mkorbel1:
## Summary

- Remove unused dependencies from the ROHD hierarchy package, DevTools extension, and Dart source navigator.
- Remove the unused ESLint toolchain from the VS Code extension.
- Make VS Code extension builds reproducible by pinning `@vscode/vsce`, committing `package-lock.json`, and using `npm ci`.
- Require Node.js 22 or newer and provide an `.nvmrc` selecting Node 24.
- Package the production `ws` dependency in the VSIX instead of excluding dependencies.
- Add the extension license and explicit `.gitignore` and `.vscodeignore` files.
- Update extension build and installation documentation.

The DevTools-level ROHD override remains in place because it overrides the transitive ROHD dependency from `rohd_devtools_widgets`.

## Testing

- Source navigator: 15 tests passed.
- ROHD hierarchy: 339 tests passed.
- DevTools extension: 3 tests passed.
- DevTools widgets: analyzer clean and 3 tests passed.
- TypeScript compilation passed.
- Installer shell syntax validation passed.
- VSIX packaging passed and the resulting archive was verified to contain:
  - The compiled extension
  - The BSD license
  - The `ws` runtime package
- Verified development-only sources and tooling are excluded from the VSIX.
- 
## Backwards-compatibility

No

## Documentation

No


desmonddak (approved):
These changes look good.

PR #689, opened 2026-07-21
Return the original signal when replicating by 1

### LINKED ISSUE #522: When replicating by 1, just return the 1 signal
### Motivation

If you `signExtend` by 1 bit, you end up with a `1{...}`, which is unnecessarily verbose in generated SystemVerilog.

The `replicate` function has the same thing.

### Desired solution

Make it so that `Logic.replicate`, when `multiplier` is 1, just returns the same thing.

Make it so `Logic.signExtend` uses `replicate` (instead of directly `ReplicationOp`), so that it gets the same benefit.

### Alternatives considered

_No response_

### Additional details

_No response_

aishasalim:
`Logic.replicate(1)` built a `ReplicationOp`, which emits a redundant `{1{...}}` in the generated SystemVerilog:

    assign rep1 = {1{a}};

Replicating once is a no-op, so return the signal directly instead. The generated output becomes:

    assign rep1 = a;

`signExtend` already routes through `replicate`, so it inherits the same benefit: sign-extending a 1-bit signal to width 1, or any signal to its existing width, no longer emits a replication.

Validation of non-positive multipliers is unaffected; those still reach `ReplicationOp` and throw `InvalidMultiplierException`.

Adds tests for the returned signal identity and for the absence of `{1{` in generated SystemVerilog. The existing "multiply by 1" test only checked simulated values, which pass either way.

Fixes #522

## Backwards-compatibility

## Documentation

### DIFF
--- a/test/replication_test.dart
+++ b/test/replication_test.dart
@@ -1,4 +1,4 @@
-// Copyright (C) 2023 Intel Corporation
+// Copyright (C) 2023-2026 Intel Corporation
 // SPDX-License-Identifier: BSD-3-Clause
 //
 // replication_test.dart
@@ -18,6 +18,13 @@ class ReplicationOpModule extends Module {
     final b = addOutput('b', width: newWidth < 1 ? a.width : newWidth);
 
     b <= a.replicate(multiplier);
+  }
+}
+
+class SignExtendModule extends Module {
+  SignExtendModule(Logic a, int newWidth) {
+    a = addInput('a', a, width: a.width);
+    addOutput('b', width: newWidth) <= a.signExtend(newWidth);
   }
 }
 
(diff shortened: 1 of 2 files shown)


test/replication_test.dart
mkorbel1, line 1: nit: you could update the copyright year at the top of the file to `-2026`

test/replication_test.dart
```diff
@@ -52,6 +59,27 @@ void main() {
         ], 1, originalWidth: 4);
       });
 
+      test('multiply by 1 returns the original signal', () {
+        final a = Logic(width: 4);
+        expect(identical(a.replicate(1), a), isTrue);
+      });
+
+      test('multiply by 1 generates no replication in SystemVerilog', () async {
+        final mod = ReplicationOpModule(Logic(width: 4), 1);
```
mkorbel1, line 68: I'm a little confused with this line -- the implementation of `ReplicationOpModule` uses `signExtend` with a `newWidth`; here the new width is 1 but original is 4, so I would expect this to throw an exception since `1 < 4` when you try to sign extend it to something smaller than the original.

mkorbel1 (changes requested):
Thank you for your contribution!  A couple notes/questions

test/replication_test.dart
mkorbel1: oh whoops i was confused this is not the `SignExtendModule`, nevermind!

test/replication_test.dart
aishasalim: Done — bumped to `2023-2026`. Thanks!

mkorbel1 (approved):

PR #694, opened 2026-08-06
Throw an exception for duplicate interface port names

### LINKED ISSUE #514: `Interface._setPort` assertion should be an exception
### Describe the bug

This assertion exists in `interface.dart` in `Interface._setPort`:
```dart
    assert(!_ports.containsKey(portName),
        'Port named $portName already exists on this interface.');
```

This should be an `Exception`, not an `assert`.

### To Reproduce

_No response_

### Expected behavior

_No response_

### Actual behavior

_No response_

### Additional: Dart SDK info

_No response_

### Additional: pubspec.yaml

_No response_

### Additional: Context

_No response_

aishasalim:
Fixes #514.

`Interface._setPort` caught a port name that already exists on the interface with an `assert`:

```dart
assert(!_ports.containsKey(portName),
    'Port named $portName already exists on this interface.');
```

Asserts are stripped in release mode, so outside of tests the duplicate silently replaced the earlier entry in `_ports` and the interface ended up carrying a port the caller never intended.

This replaces it with a `DuplicatePortNameException`, a new `RohdException` alongside the existing `InterfaceNameException` and `InterfaceTypeException` in `lib/src/exceptions/interface/`, so the collision is reported in every build mode.

### Testing

New test in `interface_test.dart` builds an interface that declares the same port name twice and expects the exception. Confirmed it fails against unmodified `interface.dart` first — it threw `_AssertionError` rather than a `RohdException` — so the test genuinely covers the change.

Full gate locally: `dart test` 1320 passed, `dart format` 0 changed, `dart analyze --fatal-infos` clean, `dart doc` 0 warnings and 0 errors, `pana` 160/160.

### Note

This is a behavior change for release-mode builds that were unknowingly relying on the silent overwrite — that is the point of the issue, but flagging it since it could surface in downstream code. Happy to rename the exception or reword the message if you would prefer something different.

### DIFF
--- a/lib/src/exceptions/interface/duplicate_port_name_exception.dart
+++ b/lib/src/exceptions/interface/duplicate_port_name_exception.dart
@@ -1,0 +1,20 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// duplicate_port_name_exception.dart
+// Definition for an exception thrown when a port name is already in use on an
+// interface.
+//
+// 2026 August 5
+// Author: Aisha Salimgereyeva <aishasalimg@gmail.com>
+
+import 'package:rohd/src/exceptions/rohd_exception.dart';
+
+/// An [Exception] thrown when a port is added to an interface with a name that
+/// is already in use by another port on that same interface.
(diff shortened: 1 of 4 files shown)


mkorbel1 (approved):

PR #697, opened 2026-08-18
Feature/signal value format registry

desmonddak:
## Description & Motivation

This is a dependency needed as central for widgets to communicate changes in the way a signal value is displayed.

## Related Issue(s)

None.

## Testing

This feature is used in future widgets where testing is performed, but not introduced yet.  But basic testing for store/recall of format is performed.

## Backwards-compatibility

No.

## Documentation

No, not yet.  This is a narrow dependency needed to introduce future widgets.


rohd_devtools_extension/packages/rohd_devtools_widgets/test/signal_value_format_registry_test.dart
```diff
@@ -0,0 +1,77 @@
...
+// Tests for shared signal display-format preferences and value formatting.
+//
+// 2026 August
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:flutter_test/flutter_test.dart';
+import 'package:rohd_devtools_widgets/rohd_devtools_widgets.dart';
+
+void main() {
+  test('formats bare binary and hexadecimal waveform values', () {
```
mkorbel1, line 17: some of these are surprising to me.  How does it know `11111111` is binary and not decimal or hex?  how does it know `ff` is hex?

mkorbel1, line 17: some of these are surprising to me.  How does it know `11111111` is binary and not decimal or hex?  how does it know `ff` is hex?

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/signal_value_format_registry.dart
```diff
@@ -0,0 +1,210 @@
...
+
+/// Shared display-format preferences keyed by signal hierarchy path.
+///
+/// Viewer packages publish format names from their local format enums; other
+/// embedded surfaces read the same names without depending on those enums.
+class SignalValueFormatRegistry {
+  SignalValueFormatRegistry._();
+
+  /// Current occurrence-format preferences.
+  static final formats = ValueNotifier<Map<String, String>>(
```
mkorbel1: why aren't we using enums for the different supported types?

rohd_devtools_extension/packages/rohd_devtools_widgets/test/signal_value_format_registry_test.dart
```diff
@@ -0,0 +1,77 @@
...
+    expect(
+      SignalValueFormatRegistry.formatForAny([
+        'unresolved-id',
+        'top/adder/result',
+      ]),
+      'unsignedDecimal',
+    );
+  });
+
+  test('looks up a uniquely matching hierarchy suffix', () {
```
mkorbel1: i'm confused why these non-matching things should match?  it always matches any suffix?

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/signal_value_format_registry.dart
desmonddak: Done.

rohd_devtools_extension/packages/rohd_devtools_widgets/test/signal_value_format_registry_test.dart
desmonddak: I'm now using `LogicValue` to help interpret values and convert them.  In the case above, the length was the helper to disambiguate.

desmonddak: I'm now using `LogicValue` to help interpret values and convert them.  In the case above, the length was the helper to disambiguate.

rohd_devtools_extension/packages/rohd_devtools_widgets/test/signal_value_format_registry_test.dart
desmonddak: It now doesn't match this way as it is using the `SignalOccurrence` class and utilities.

desmonddak:
Replaced by #701

PR #699, opened 2026-08-19
fix for coverage failure

desmonddak:
## Description & Motivation

Current CI is failing as gen_coverage.sh stops when it cannot find any branch coverage (we currently do not enable).

## Related Issue(s)

None filed. We need CI running.

## Testing

Ran gen_coverage.sh to completion.  Ran dart test.

## Backwards-compatibility

No

## Documentation

No


mkorbel1 (approved):

mkorbel1:
Thank you for the fix!

PR #701, opened 2026-08-25
Snapshot/signal value format registry

desmonddak:
## Description & Motivation

This is a dependency needed as central for widgets to communicate changes in the way a signal value is displayed.
It replaces PR #697 which is now a draft.

## Testing

This feature is used in future widgets where testing is performed, but not introduced yet. But basic testing for store/recall of format is performed.

## Backwards-compatibility

No

## Documentation

Not yet.

### DIFF
--- a/packages/rohd_hierarchy/lib/src/occurrence_trie.dart
+++ b/packages/rohd_hierarchy/lib/src/occurrence_trie.dart
@@ -1,0 +1,115 @@
+// Copyright (C) 2026 Intel Corporation
+// SPDX-License-Identifier: BSD-3-Clause
+//
+// occurrence_trie.dart
+// Compact storage for values keyed by hierarchy occurrence addresses.
+//
+// 2026 August
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:rohd_hierarchy/src/occurrence_address.dart';
+
+/// A prefix-sharing map from [OccurrenceAddress] values to values of type [T].
+///
+/// Common address prefixes are stored once, making this more compact than a
+/// conventional map when many values belong to the same hierarchy subtree.
+class OccurrenceTrie<T extends Object> {
+  /// The root node, which represents [OccurrenceAddress.root].
+  final _OccurrenceTrieNode<T> _root = _OccurrenceTrieNode<T>();
+
+  /// Whether this trie contains no values.
+  bool get isEmpty => _root.isEmpty;
+
+  /// The value stored at [address], if any.
+  T? operator [](OccurrenceAddress address) {
+    var node = _root;
+    for (final index in _validatedPath(address)) {
+      final child = node.children[index];
+      if (child == null) {
+        return null;
+      }
+      node = child;
+    }
+    return node.value;
+  }
+
+  /// Associates [value] with [address].
+  ///
+  /// This is equivalent to [set], without returning the previous value.
+  void operator []=(OccurrenceAddress address, T value) {
+    set(address, value);
+  }
+
+  /// Associates [value] with [address].
+  ///
+  /// Returns the value previously stored at [address], if any.
+  T? set(OccurrenceAddress address, T value) {
+    var node = _root;
+    for (final index in _validatedPath(address)) {
+      node = node.children.putIfAbsent(index, _OccurrenceTrieNode<T>.new);
+    }
+    final previous = node.value;
+    node.value = value;
+    return previous;
+  }
+
+  /// Removes and returns the value stored at [address], if any.
+  T? remove(OccurrenceAddress address) {
+    final path = _validatedPath(address);
+    final nodes = <_OccurrenceTrieNode<T>>[_root];
+    var node = _root;
+    for (final index in path) {
+      final child = node.children[index];
+      if (child == null) {
+        return null;
+      }
+      nodes.add(child);
+      node = child;
+    }
+
+    final previous = node.value;
+    if (previous == null) {
+      return null;
+    }
+    node.value = null;
+    for (var index = path.length - 1; index >= 0; index--) {
+      final child = nodes[index + 1];
+      if (!child.isEmpty) {
+        break;
+      }
+      nodes[index].children.remove(path[index]);
+    }
+    return previous;
+  }
+
+  /// Removes every value from this trie.
+  void clear() {
+    _root
+      ..value = null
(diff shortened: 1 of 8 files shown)


packages/rohd_hierarchy/lib/src/occurrence_trie.dart
```diff
@@ -0,0 +1,108 @@
...
+// 2026 August
+// Author: Desmond Kirkpatrick <desmond.a.kirkpatrick@intel.com>
+
+import 'package:rohd_hierarchy/src/occurrence_address.dart';
+
+/// A prefix-sharing map from [OccurrenceAddress] values to values of type [T].
+///
+/// Common address prefixes are stored once, making this more compact than a
+/// conventional map when many values belong to the same hierarchy subtree.
+class OccurrenceTrie<T extends Object> {
```
mkorbel1, line 16: question: do you really need to specify `extends Object`?

packages/rohd_hierarchy/lib/src/occurrence_trie.dart
```diff
@@ -0,0 +1,108 @@
...
+      }
+      node = child;
+    }
+    return node.value;
+  }
+
+  /// Associates [value] with [address].
+  ///
+  /// Returns the value previously stored at [address], if any.
+  T? set(OccurrenceAddress address, T value) {
```
mkorbel1: API question: why not make it so that you can assign using `operator []` also?

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/signal_value_format_registry.dart
```diff
@@ -0,0 +1,239 @@
...
+                      .toInt();
+              return code >= 0x20 && code <= 0x7e ? code : 0x2e;
+            },
+          ),
+        ),
+      SignalValueFormat.waveform => canonical,
+    };
+  }
+
+  static (LogicValue, String)? _waveformValue(String value, int width) {
```
mkorbel1, line 237: nit: add doc comments even for private functions and fields?

mkorbel1 (commented):
looks good to me!

packages/rohd_hierarchy/lib/src/occurrence_trie.dart
desmonddak: We need 'null'

mkorbel1: what does that mean?  sorry can you explain?

packages/rohd_hierarchy/lib/src/occurrence_trie.dart
```diff
@@ -0,0 +1,115 @@
...
+      node = child;
+    }
+    return node.value;
+  }
+
+  /// Associates [value] with [address].
+  ///
+  /// This is equivalent to [set], without returning the previous value.
+  void operator []=(OccurrenceAddress address, T value) {
+    set(address, value);
```
mkorbel1: good, but then do you need `set` to be a public API anymore?

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/signal_value_format_registry.dart
```diff
@@ -0,0 +1,271 @@
...
+      SignalValueFormat.waveform => canonical,
+    };
+  }
+
+  /// Parses [value] as a waveform literal and returns it with its canonical
+  /// waveform representation.
+  static (LogicValue, String)? _waveformValue(String value, int width) {
+    final trimmed = value.trim().replaceAll('\u0000', '');
+    if (trimmed.isEmpty || _containsUnknownDigits(trimmed)) {
+      return null;
```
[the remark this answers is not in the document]
desmonddak: We are now using our `LogicValue` methods to deal with waveform values correctly.

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/signal_value_format_registry.dart
```diff
@@ -0,0 +1,271 @@
...
+    final radixLiteral = isRadixLiteral ? lower : "$displayWidth'$radix$digits";
+    try {
+      final logicValue = LogicValue.ofRadixString(radixLiteral);
+      return (
+        logicValue,
+        isRadixLiteral ? trimmed : logicValue.toString(),
+      );
+    } on LogicValueConstructionException {
+      return null;
+    }
```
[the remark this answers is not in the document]
desmonddak: Done

packages/rohd_hierarchy/lib/src/occurrence_trie.dart
desmonddak: Removed and replaced calls with the operator.

packages/rohd_hierarchy/lib/src/occurrence_trie.dart
desmonddak: We use `null` to recognize that no value is stored at this address.  If we allow an arbitrary type, like `int`, then we need to have a different sentinal to tell that there is no value stored.

rohd_devtools_extension/packages/rohd_devtools_widgets/lib/src/signal_value_format_registry.dart
```diff
@@ -0,0 +1,293 @@
...
+    if (trimmed.isEmpty) {
+      return null;
+    }
+    final lower = trimmed.toLowerCase();
+    final displayWidth = width > 0 ? width : 1;
+    final isRadixLiteral = RegExp(r"^\d+'[bqodh]").hasMatch(lower);
+    final digits = lower.startsWith('0x') || lower.startsWith('0b')
+        ? lower.substring(2)
+        : lower;
+    final radix = lower.startsWith('0x')
```
mkorbel1, line 248: do we really need this routine to discover the radix? can't we just be more explicit?  can we at least use a regex or something instead of this opaque `codeUnit` approach?

desmonddak: I replaced with `RegEx`.  I want to share as much as I can with `LogicValue` where we have the radix string parsing and production available.

mkorbel1 (approved):
Looks good to me!

PR #702, opened 2026-08-28
Fix/devtools popup menu material UI

desmonddak:
## Description & Motivation

We are migrating devtools to use the latest analysis options and discovered some deprecated symbols.
Also, we found a Javascript file was needed in the shared widget support area for future widgets (module_info.js)

Importantly, we amended this PR to deploy the web app so that it is invokeable from our git page.

The assembled deployment tree becomes:

```text
doc/api/
├── index.html
├── libraries.html
├── ... Dart API documentation ...
└── rohd_devtools_extension/
    ├── index.html
    ├── main.dart.js
    ├── flutter_bootstrap.js
    └── assets/
```

The existing deployment action publishes all of `doc/api/` to the `docs` branch:

```yaml
- name: Deploy generated documentation
  uses: JamesIves/github-pages-deploy-action@v4
  with:
    folder: doc/api
    branch: docs
```

Consequently, the nested build directory becomes a top-level URL beneath the ROHD Pages site:

```text
https://intel.github.io/rohd/rohd_devtools_extension/
```

The key point is that documentation and the compiled web app are deployed together in one operation. This prevents one deployment from erasing the other.

## Related Issue(s)

None

## Testing

Ran existing tests.

## Backwards-compatibility

No

## Documentation

None needed.  This is lint cleanup, moving away from deprecated Flutter classes, and centralizing a resource file.


rohd_devtools_extension/tool/gh_actions/build_app.sh
```diff
@@ -0,0 +1,68 @@
...
+      base_href="/"
+    else
+      base_href="/$repository/"
+    fi
+  else
+    base_href="/rohd_devtools_extension/"
+  fi
+fi
+
+if [[ ! "$base_href" =~ ^/[A-Za-z0-9._/-]*/$ ]]; then
```
mkorbel1: Nit, non-blocking: this regex rejects `/`, although the owner-level Pages branch above explicitly sets `base_href="/"`. Could we accept `/` as well as the existing subdirectory paths? ROHD's configured `/rohd/rohd_devtools_extension/` path is unaffected.

mkorbel1 (approved):
Looks good to me. The UI migration, source-navigation support, and combined documentation/app deployment look reasonable. Relying on your functional testing and the passing Run Checks and Build Devtools checks; I have not rerun the UI tests or builds independently. One non-blocking nit on root-hosted base-path validation, which does not affect the configured ROHD deployment.

README.md
```diff
@@ -36,6 +36,33 @@ You can also open this repository in a GitHub Codespace to run the example in yo
 - [Attend the ROHD Forum meetings](https://intel.github.io/rohd-website/forum/rohd-forum/)
 - [GitHub Discussions](https://github.com/intel/rohd/discussions)
 
+## Generating a netlist
```
mkorbel1: Please remove the new "Generating a netlist" section from the main ROHD README. We do not need a specialized netlist-generation walkthrough in the project overview; this belongs in the generation user guide, where this PR already adds it. Keep the detailed example there rather than duplicating it in the README.

desmonddak: You are right!  I didn't see this duplication.

mkorbel1 (approved):
looks good to me!

PR #712, opened 2026-09-08
Allow same-name reserved and renameable signals to merge during synthesis

mkorbel1:
## Description & Motivation

Allow connected, equivalent `Naming.reserved` and `Naming.renameable` signals to share a generated declaration when all preserved names match exactly in the module's namespace before uniquification. This applies to internal signals and port aliases, including chains with mixed naming modes.

Previously, these connections could generate redundant uniquified aliases or throw reserved-name collision errors despite representing the same signal. The updated contract preserves a generated name without requiring a separate declaration for every original signal.

Merged groups retain every original `Logic` and its mappings. Different preserved names remain separate, unrelated reserved-name collisions still fail, and unrelated renameable signals are still uniquified. Module interfaces and structural merge restrictions remain protected.

## Related Issue(s)

None linked.

## Testing

- Added coverage for all reserved/renameable combinations in three-signal chains at internal, input, and output locations.
- Covered fanout, signal mappings, repeated synthesis, constants, qualified names, unrelated collisions, packed/unpacked arrays, bidirectional nets, and structural merge restrictions.
- Checked four-state behavior in ROHD and Icarus Verilog, using packed monitors for unpacked-array outputs.
- Existing tests cover a lot

## Backwards-compatibility

Somewhat, but it shouldn't be in a meaningful way.  This changes the documented naming behavior and may change generated identifiers and declarations. Equivalent same-name preserved signals may now share a declaration, and connected same-name reserved signals no longer trigger collision errors. Downstream scripts or tests that depend on separate generated aliases or the previous collision errors may need updates. No Dart API signatures change.

## Documentation

Yes. Updated the `Naming` API doc comments, synthesis merge doc comments, and the generation user guide to explain the name-preservation contract and its restrictions. Included a changelog entry under Next Release.


desmonddak:
Looks like it was a naming bug before ! :)

I don't quite see why the three-signal chain is enough to test this completely.

Are there any chaining combinations that could block (I don't see that many permutations in the tests).

PR #725, opened 2026-09-23
Fix aggregate netlist validation for disjoint field drivers, Issue #724

### LINKED ISSUE #724: Netlist validation rejects disjoint drivers within structured aggregate nets
### Describe the bug

`NetlistValidation.validate` incorrectly reports a multiple-driver error when different cells drive disjoint bits, fields, or elements of a structured aggregate net.

The validation logic gathers every driver associated with the aggregate net and requires the aggregate to have at most one distinct driver. This treats independent assignments to separate fields/elements as overlapping electrical drivers, even though no individual wire bit has more than one driver.

For example, synthesizing a structured array output with independent muxes for different lanes fails with an error like:

```text
aggregate net "headWindow_orderedRobTags" is reached from multiple drivers:
cell mux_394.Y ($mux), cell mux_402.Y ($mux)
```
The generated SV drivers separate aggregate elements:
```verilog
assign headWindow_orderedRobTags[0] = ...;
assign headWindow_orderedRobTags[1] = ...;
```

### To Reproduce

```markdown
1. Create or synthesize a module whose output is a packed structured aggregate (for example, a `LogicArray` or `LogicStructure`).
2. Drive independent fields/elements of that aggregate from separate cells, such as distinct muxes or buffers.
3. Export the design through the JSON netlist synthesizer.
4. Observe that validation throws an aggregate multiple-driver error, despite the drivers targeting non-overlapping bits.

A minimal validator-level equivalent is:

- one `$buf` drives bits `[200, 201]`;
- another `$buf` drives bits `[400, 401]`;
- both bit ranges are fields of the same structured aggregate netname.

This should succeed because the two output bit ranges are disjoint.

### Expected behavior

Netlist validation should allow different drivers for disjoint bits within the same structured aggregate.

It should continue to reject a true conflict when two cell outputs drive the same wire bit.

### Actual behavior

Netlist validation rejects the aggregate merely because it has more than one distinct driver, even when every driven bit has exactly one driver.

### Additional: Dart SDK info

Dart SDK version: 3.13.2 (stable) on linux_x64

### Additional: pubspec.yaml

```yaml
name: rohd
version: 0.6.10

environment:
  sdk: '>=3.0.0 <4.0.0'
```

### Additional: Context

_No response_

desmonddak:
## Description & Motivation

Remove aggregate-wide driver validation that incorrectly reports multiple drivers when independent fields or elements of a structured aggregate are driven by different cells.

Netlist validation already checks drivers per wire bit, which correctly rejects electrically overlapping drivers. The aggregate-wide union check was redundant and rejected valid synthesized designs, including structured array outputs whose independent lanes are assigned by distinct muxes.

Update the regression coverage to accept disjoint field drivers and to verify that a shared driven bit is still rejected.

## Related Issue(s)

Fixes #724

## Testing

- `dart test test/netlist_synthesizer_test.dart`
  - 87 tests passed.
- Focused analyzer checks for the modified synthesizer and validation sources passed.
- `git diff --check`

## Backwards-compatibility

No. This relaxes an incorrect validation failure for legal disjoint aggregate-field drivers while retaining per-bit multiple-driver detection.

## Documentation

No. This is an internal netlist-validation correction with updated regression coverage.


test/netlist_synthesizer_test.dart
```diff
@@ -2145,30 +2145,42 @@ void main() {
...
-            ...List<Object>.generate(8, (index) => 400 + index),
-          ],
-          'logic_type': {
-            'typeName': 'PairStructure',
-            'fields': [
-              {'name': 'first', 'width': 8},
-              {'name': 'second', 'width': 8},
-            ],
+      expect(
+        () => NetlistValidation.validate(ports, cells, 'struct_module'),
```
mkorbel1: Non-blocking test suggestion: could we add a synthesis-level regression reduced from the failing design in #724? This test now omits `netnames`, so it also passes with the old validator, where the aggregate check was conditional on that argument. I ran the entire revised test file against the pre-fix production sources and all 87 tests passed. A case that retains an aggregate netname spanning independent mux drivers and fails before this fix would protect the behavior being corrected; the shared-bit rejection test can remain as complementary coverage. The implementation itself looks correct; this is a coverage improvement, not a functional defect or merge blocker.

mkorbel1 (commented):
The implementation looks correct: removing the aggregate-wide restriction allows independent field drivers while retaining the per-bit overlap check. No functional defect found. The inline regression-test suggestion is non-blocking.

At this revision, the focused netlist suite passed all 87 tests in my September 23 review; tests were not rerun for this submission.

